PR #2, opened 2026-06-04
refactor: unify package namespace to io.github.eschizoid

domdorn:
Migrate all source from com.github.eschizoid to io.github.eschizoid, matching the Maven group id already used in build files. Moves the java source roots across all modules/source sets and rewrites package declarations, imports, module-info, SPI service files, and docs.


eschizoid (approved):

eschizoid (approved):

eschizoid:
thanks for this 😸

PR #8, opened 2026-06-06
feat(core): typed container subclasses + lattice-routed assembly + reflection deletion + Beans reshape

eschizoid:
## Summary

Pushes the **lattice-first** principle further: deletes the runtime container-dispatch escape hatch (and with it all `java.lang.reflect.Array` usage), introduces compile-checked typed container subclasses, reshapes `DeepMap.assembleIso` to compose three Isos via pure `.then(...)`, and reshapes the read side of `Beans` into a lattice primitive. Ergonomics are preserved for every existing call site that doesn't use the deleted escape hatches.

## DSL additions

- **Sealed `Telescope<S, A>`** with four nested typed subclasses: `ListPath<S, X>`, `SetPath<S, X>`, `MapPath<S, K, V>`, `OptionalPath<S, X>`. Each exposes its compile-checked typed terminal (`.each()` / `.values()` / `.present()`) that descends via a pure lattice composition (no runtime container dispatch).
- **Typed instance accessors** on `Telescope`: `.list(Accessor)`, `.set(Accessor)`, `.map(Accessor)`, `.optional(Accessor)` — return the narrower subclass so the terminal is available.
- **Static promotion factories**: `Telescope.asList(path)`, `.asSet(path)`, `.asMap(path)`, `.asOptional(path)` for pre-built `Telescope<S, Container<X>>` fragments (codegen + power users).

## Reflection deletion

| Surface | Before | After |
|---|---|---|
| `Telescope.each()` no-arg | runtime-dispatched escape hatch | **deleted** |
| `Traversals.eachContainer` + `containerStream` + `containerUpdate` | ~80 LOC \`instanceof\` dispatch chain | **deleted** |
| `Traversals.arrayStream` / `arrayUpdate` | \`java.lang.reflect.Array.get/set/newInstance\` | **deleted** |
| \`java.lang.reflect.Array\` import anywhere in `:core` | yes | **zero** |
| Array containers (`int[]`, `String[]`, …) | supported via runtime reflection | **dropped** — wrap as `List` (see PLAN.md item #6) |

`Traversals.eachIterable()` is the kept polymorphic primitive used by the typed `.each(Accessor<A, ? extends Iterable<E>>)` form — dispatches once on the runtime class to pick the concrete rebuild shape (List → ArrayList, Set → LinkedHashSet). No array branch.

## Codegen (`:codegen`)

- `AbstractTelescopeProcessor` deduplication — pulled the 90-LOC mirror code out of `FocusProcessor.emitComponentMethod` / `emitStep` into shared `emitNavigatorMethod` / `emitContainerStep` / `emitPathClassHeader`. ~50 LOC delete net.
- Container-step emission rewritten to use the typed static factory at the body site: `Telescope.<R, X>asList(path).each()`, `Telescope.<R, X>asSet(path).each()`, `Telescope.asMap(path).values()`, `Telescope.<R, X>asOptional(path).present()`. No runtime dispatch in the generated code either.
- `TraversalShape` carries `containerKind` (`list`/`set`/`map`/`optional`/`iterable`) so the right factory is chosen.
- `navigableType(...)` / `isAnnotatedClass(...)` widened to accept both `CLASS` (`@BeanFocus` / Lombok targets) and `RECORD` (`@Focus` targets) so the consolidated emit path works for either.

Minor cleanups: `staticNoArgMethod(type, \"builder\")` → `staticBuilderMethod(type)`; `emitTelescopeForwarders(out, \"path\", focusType)` → `emitTelescopeForwarders(out, focusType)`.

## Beans reshape (Phases 5 + 6)

- **`Beans.getter(Class<P>, String name)` → `Getter<P, Object>`** — the lattice-primitive read for one property, backed by the cached `Method` from the `GETTERS` `ClassValue`. `readProperty(p, name)` is now a thin convenience over it. Routes the read side through the lattice rather than bare `Method.invoke`.
- **`Reflective.structuralIso(Class<T>)`** — new default method that turns a `Reflective` impl into `Iso<Map<String, Object>, T>` (forward = `construct` from the map; backward = read every component / property into a fresh `LinkedHashMap`).
- **`DeepMap.assembleIso` rewritten** as a pure lattice composition:

  \`\`\`java
  final Iso<S, Map<String, Object>> srcReader = srcRefl.structuralIso(source).reverse();
  final Iso<Map<String, Object>, Map<String, Object>> remap = remapIso(byTargetName, bySourceName);
  final Iso<Map<String, Object>, T> tgtBuilder = tgtRefl.structuralIso(target);
  return nullable(srcReader.then(remap).then(tgtBuilder));
  \`\`\`

  Bidirectional behavior falls out of `.then(...)` automatically. No inline backward lambda, no manual `step.iso.from(...)` plumbing.

## Javadoc + error-message sweep

- `Beans.java` class-level rewritten — points at the actual API (`Telescope.ofBean` / `Telescope.map` / `WriteHint.writeBean`), names `Beans.getter` as the lattice read primitive.
- All four `*Writer` inner messages rewritten: stale `fromBean(...).viaFields() / viaConstructor() / viaBuilder() / viaSetters()` → `writeBean(class, FIELDS|CONSTRUCTOR|BUILDER|SETTERS)`.
- `Beans.computeAutoWriter` regex shim **deleted** — no longer needed; inner messages are clean.
- `DeepMap.writerFor` try/catch wrap **deleted** — same reason.
- `DeepMap` class-level rewritten to describe the three-step lattice composition via `structuralIso` + `remap`.

## README sweep

Reviewed every section. Replaced demolished APIs in:

- **DSL surface table**: dropped `.each()` (no-arg) row; added typed `.list/.set/.map/.optional` + `Telescope.asList/.asSet/.asMap/.asOptional` rows; mentioned `WriteHint` on the `Telescope.map` row.
- **Working with POJOs**: rewrote the whole section. The `Telescope.fromBean(...).viaX()` and `Telescope.mapBean(...).build()` chains are gone — examples now use `Telescope.map(A.class, B.class, ...)` with `Mapping.to/via` rows and `WriteHint.writeBean(target, strategy)`.
- **Constraints**: collapsed \"two documented runtime-check points\" → one (just `.fieldByName`). Updated the breakage-history entry to record the demolitions.
- **Architecture table**: `Traversals.eachContainer` → `eachIterable`, added a row for the typed `.list(...)` form.
- **Cookbook**: new \"Typed container leaves (pre-built fragments)\" entry.

## Net diff

- `+1457 / -757` across 14 files
- **266 core tests pass** (was 249) — 17 new (13 `TypedContainerPathTest` + 4 `ReflectiveStructuralIsoTest`)
- Full multi-module build green
- `Telescope.java` instruction coverage: **97%**

## Breaking changes (pre-1.0; recorded in README + PLAN.md)

- `Telescope.each()` no-arg deleted. Migration: typed `.list/.set/.map/.optional(accessor)` + their terminals, or `Telescope.asList(path).each()` for pre-built fragments.
- Array containers (`int[]`, `String[]`) no longer supported. Wrap as `List`/`Set`.
- `Traversals.eachContainer` deleted; replaced by typed primitives + `Traversals.eachIterable` for the polymorphic case.

## Test plan

- [x] `./gradlew :core:test` — 266 tests pass
- [x] `./gradlew build` — full multi-module build green (`:core`, `:codegen`, `:lombok`, `:benchmarks`)
- [x] `./gradlew :core:jacocoTestReport` — `Telescope.java` 97% instruction coverage
- [x] Spotless + javadoc clean
- [ ] Smoke-test the typed container subclasses in a downstream consumer once one exists
- [ ] Verify CI green on push


eschizoid:
@copilot please review

eschizoid:
@copilot round 1 addressed — Iterable codegen now uses explicit wrap type witnesses; Beans.readProperty inlines Method.invoke; Telescope private-field suggestion left as package-private since subclass inheritance can't read private parent fields (Java rule, verified). Please re-review.

eschizoid:
Post-merge follow-up addressed all 5 round-2 comments in #10: Beans.getter javadoc clarified, FieldsWriter exception message tightened, eachIterable rejects non-List/non-Set Iterables explicitly, codegen comment accurate, regression test added.

PR #9, opened 2026-06-06
feat(core): Phase 1 — LambdaMetafactory record-component readers (no Method.invoke on the record read path)

eschizoid:
## Summary

Phase 1 of the 5-phase plan to move every `Method.invoke` / `Constructor.newInstance` / `Field.set` off the hot path. This PR handles **record component reads** — the most-traveled reflective call in the codebase.

`RecordComponent.getAccessor().invoke(source)` is replaced with a cached `Function<Object, Object>` per component, built once via `LambdaMetafactory` at class-cache fill time. `java.lang.reflect.Method` no longer appears anywhere in the steady-state record read path.

## What changed

- **`Records.RecordInfo`** gains a `Function<Object, Object>[] readers` field. `RecordInfo.of(cls)` builds the readers via `LambdaMetafactory.metafactory(...)` — the metafactory synthesizes a class implementing `Function` whose `apply(Object)` calls the component accessor directly. Primitive returns are auto-boxed via the `instantiatedMethodType` bridge.
- **`readField` / `updateField`** dispatch through `info.readers()[idx].apply(source)` instead of `info.components()[idx].getAccessor().invoke(source)`.
- **`CACHE`** migrated from `ConcurrentHashMap<Class<?>, RecordInfo>` to `ClassValue<RecordInfo>` for the same classloader-unload safety already in `Beans`.

## Hot-path reflection inventory in `Records.java`

| Surface | Before | After |
|---|---|---|
| `Method.invoke` per component read | 2 sites called per read/update | **0 sites** |
| `Constructor.newInstance` (rebuild) | 1 site per rebuild | unchanged — Phase 5 |
| Bootstrap reflection (`getRecordComponents`, `unreflect`, etc.) | once per class | unchanged; fires inside `RecordInfo.of` only |

## What's NOT in this PR

- `Records.construct(...)` still calls `Constructor.newInstance(args)` — that's Phase 5 (closes the rebuild half).
- `Beans.java` reflection is unchanged — Phases 2-4 handle that ladder.
- `LambdaIntrospection.java` (SerializedLambda) has no alternative; it stays as-is.

## Constraint

`MethodHandles.privateLookupIn(cls, lookup())` needs the target's package to be `opens` to the telescope module — exactly the same JPMS constraint as the previous `setAccessible(true)` had. No new requirement, no worse. The error message documents the `opens` directive if access fails.

## Test plan

- [x] `./gradlew :core:test` — 266 tests pass (unchanged from main)
- [x] `./gradlew build` — full multi-module build green
- [x] Spotless / javadoc clean
- [ ] Benchmark sweep (deferred to a follow-up PR — runs across the 5-phase stack to compare cleanly at the end)

## Stack

This is Phase 1 of 5. Each phase ships independently.

- **Phase 1 (this PR)**: `Records.read` via LambdaMetafactory
- Phase 2: `Beans.getter` / `Beans.readProperty` same treatment
- Phase 3: `Beans.SettersWriter` via `BiConsumer`
- Phase 4: `Beans.BuilderWriter` setters + `build()`
- Phase 5: `Beans.ConstructorWriter` + `FieldsWriter` + `Records.construct` (closes the rebuild half)

After phase 5, `Method.invoke` / `Constructor.newInstance` / `Field.set` are gone from every code path the user's data flows through. They survive only inside the per-class `computeValue(...)` lambda of the `ClassValue` caches — fired once per class for the lifetime of the JVM.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/internal/Records.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/internal/Records.java
@@ -1,22 +1,26 @@
 package io.github.eschizoid.telescope.internal;
 
 import io.github.eschizoid.telescope.internal.optics.Lens;
+import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.RecordComponent;
 import java.util.Arrays;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
 /**
- * Reflection-based machinery for building {@link Lens}es over record components by name. Backs the
+ * Lattice-routed machinery for building {@link Lens}es over record components by name. Backs the
  * field navigation of {@link io.github.eschizoid.telescope.Telescope}.
  *
- * <p>State is a single per-record-class {@code RecordInfo} cache: the record's {@link
- * RecordComponent} array (in canonical-constructor order) and the canonical {@link Constructor},
- * resolved once via reflection and stored in a {@link ConcurrentHashMap}. On the first lookup for a
- * class, accessors and the constructor are {@code setAccessible(true)} so later reads and rebuilds
- * skip access checks; every subsequent operation is constant-time.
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot ping — please review

PR #10, opened 2026-06-06
chore: post-merge follow-up on PR #8 Copilot review

eschizoid:
## Summary

Copilot fired on PR #8 **two minutes after merge** with 5 actionable comments that landed too late to be addressed in the original PR. This PR addresses all 5.

## Fixes

1. **`Beans.getter(...)` javadoc** — dropped the misleading \"built once per (beanClass, name) pair\" claim. The method allocates a capturing lambda per call by design (it's the lattice-shape entry point for composing a `Getter` into other optics). Reworded to point hot-path callers at `readProperty` (no lambda alloc — invokes the cached `Method` directly). Also removed the stale \"Reflective.BEANS#read delegates here\" line; round-1 of PR #8 inlined that path for exactly the alloc reason.

2. **`Beans.FieldsWriter` access() exception message** — toned down the \"switch to CONSTRUCTOR / BUILDER (which use public members only)\" guidance. CONSTRUCTOR / BUILDER / SETTERS all still invoke `setAccessible(true)` on their resolved members, so a closed package keeps failing for them too. Updated to make this explicit.

3. **`Traversals.eachIterable` cast safety** — refuse non-List / non-Set Iterable sources upfront with a clear `IllegalArgumentException` instead of fabricating an unsafe `(C)` cast. The previous \"rebuild as unmodifiable List, cast to C\" fallback was a latent ClassCastException hazard: if the caller's `C` was `Queue<E>` or any other non-`List` Iterable, the cast succeeded here but threw downstream. Class-level + method-level javadoc updated to record the supported-shapes constraint.

4. **`AbstractTelescopeProcessor` codegen comment** — removed the \"any custom user iterable\" claim from the Iterable-case branch. Replaced with an accurate description that notes the runtime rejection for `Queue`/`Deque`/custom iterables and tells users to re-declare such leaves as `List<E>` or `Set<E>` at the source so the codegen lands on the typed `list`/`set` branches instead.

5. **`LatticeCoverageTest.EachIterable`** — new test pinning the non-List/non-Set rejection behavior with an `ArrayDeque` source.

## Test plan

- [x] `./gradlew :core:test` — 267 tests pass (was 266; +1 for the new rejection test)
- [x] `./gradlew build` — full multi-module build green
- [x] Spotless / javadoc clean

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
@@ -744,14 +744,16 @@ public abstract class AbstractTelescopeProcessor extends AbstractProcessor {
       case "set" -> "Telescope.<R, " + elementType + ">asSet(path).each()";
       case "optional" -> "Telescope.<R, " + elementType + ">asOptional(path).present()";
       case "map" -> "Telescope.asMap(path).values()";
-      // Iterable case: declared leaf is some `? extends Iterable<E>` shape (e.g. Iterable<E>,
-      // Collection<E>, a custom user iterable). Java's type inference can't always work
-      // backward
-      // from path's leaf type to `eachIterable()`'s bounded type parameter, so we pin both type
-      // arguments to `Telescope.wrap(...)` explicitly — that way the produced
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #11, opened 2026-06-07
feat(core): Phase 3 — LambdaMetafactory bean setters

eschizoid:
## Summary

Phase 3 of the telescope reflection-removal stack. Swap `Method.invoke` inside `Beans.SettersWriter#construct` for cached `BiConsumer<Object, Object>` invokers built once via `LambdaMetafactory`, mirroring the canonical Phase 1 pattern (`Records.java` record-component readers, #9).

### Substrate swap — not a redesign

- `BeanWriter<P>` sealed interface stays byte-identical at the type level.
- `SettersWriter<P> implements BeanWriter<P>` declaration unchanged.
- `construct(String[] names, Function<String, Object> valueByName)` signature unchanged.
- Only the body of `construct(...)` and the `setterFor(...)` helper change: instead of resolving a `Method` and calling `setterFor(name).invoke(pojo, value)`, we now resolve the `Method` once per setter, build a `BiConsumer<Object, Object>` via `LambdaMetafactory.metafactory(...)` over an `unreflect`'d `MethodHandle`, cache it in a `ConcurrentHashMap`, and call `consumer.accept(pojo, value)` on the hot path.
- The `ctor.newInstance()` invocation stays as `Constructor.invoke` — Phase 5 owns that swap.

### Lattice preservation

`SettersWriter` is one rung of the sealed `BeanWriter` write-side ladder; the swap is pure plumbing inside the existing rung. No new optic types, no new public API, no lattice extension. Sibling parallel PRs own `FieldsWriter` / `BuilderWriter` / `ConstructorWriter` and the read-side `getter` / `readProperty` swap; this PR does not touch them.

### JPMS access behavior

Building the `BiConsumer` now first adds a read edge from `io.github.eschizoid.telescope` to the target module and tries a plain `MethodHandles.lookup().unreflect(setter)` path. That preserves compatibility for **public setters on exported types in named modules without requiring `opens`**. If extra access is still needed, the implementation falls back to `MethodHandles.privateLookupIn(...)`; only that fallback requires the target package to be opened. The thrown `IllegalStateException` still points users at the precise `opens` fix when the fallback is blocked.

### Primitive auto-unboxing

The SAM signature is `void accept(Object, Object)`. For a primitive setter (`setAge(int)`), the `instantiatedMethodType` parameter must be the **wrapper** class (`Integer`) — the validator rejects raw `int` against the SAM's `Object`. The metafactory then synthesizes the `Object → Integer → int` unbox bridge automatically. New test `settersAutoUnboxesPrimitiveArg` pins this end-to-end via a `Map<String, Object>` source feeding boxed `Integer` into `setScore(int)`.

## Test plan

- [x] `./gradlew :core:test --tests 'io.github.eschizoid.telescope.internal.BeansTest' --rerun-tasks` — targeted setter / lens / auto-writer coverage green
- [x] `./gradlew :core:test --rerun-tasks` — full core test suite green
- [x] New focused test `BeansTest.Setters.settersAutoUnboxesPrimitiveArg` covers the primitive auto-unboxing bridge
- [x] New focused test `BeansTest.Setters.settersKeepWorkingForExportedModulesWithoutOpens` covers the named-module exported-package JPMS regression
- [x] Existing tests `settersRoundTrip` (covers `setName(String)` + `setScore(int)`) and `LensFactory.lensReadWriteRebuild` / `lensModify` exercise the full SettersWriter path through `Beans.lens` and continue to pass

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
@@ -14,6 +14,7 @@ import java.lang.reflect.Modifier;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiConsumer;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

eschizoid:
Both inline review threads addressed by [PR #16](https://github.com/eschizoid/telescope/pull/16) (merged). `Beans.SettersWriter.buildSetterInvoker` and `Beans.buildGetterInvokers` now both use `*.getDeclaringClass()` for `privateLookupIn`, the `instantiatedMethodType` receiver, and the JPMS-opens error message — via a new shared `Beans.privateLookupOrThrow(declaringClass, ownerClass, accessorKind)` helper. Two new tests pin the inherited-getter and inherited-setter contracts (`ParentBean` / `ChildBean` fixture). Marking these resolved.

PR #12, opened 2026-06-07
feat(core): Phase 2 — LambdaMetafactory bean getters

eschizoid:
## Summary

Phase 2 of the reflection-removal stack — mirrors Phase 1 (PR #9) for the bean side. Substrate swap: `Beans.getter` and `Beans.readProperty` now dispatch through a cached `Function<Object, Object>` built once per property via `LambdaMetafactory`, instead of calling `Method.invoke` on every read. The lattice-primitive `Getter<P, Object>` shape, every public signature on `Beans`, and the "one capturing lambda per call" contract on `Beans.getter` are all preserved.

## Lattice-preservation framing

Same principle as Phase 1: this is a substrate swap, not a redesign.

- **Unchanged:** `Getter<P, Object>` return type, `Beans.getter` public signature, `Beans.readProperty` public signature, `Beans.hasProperty`, `Beans.propertyType`, `Beans.propertyNames`. No public surface changes.
- **Changed:** what runs inside the returned `Getter` lambda (`reader.apply(source)` instead of `method.invoke(source)`) and inside `readProperty` (cached function apply instead of `method.invoke`).
- **Lattice-shape preserved:** `Beans.getter` still allocates one capturing lambda per call — that's the lattice contract for a `Getter` (composable into other optics). `Beans.readProperty` is the hot-path shortcut that skips the per-call allocation and calls the cached `Function` directly.

A new private `ClassValue<Map<String, Function<Object, Object>>> GETTER_INVOKERS` sits alongside the existing `GETTERS` `Method` cache. On first miss it derives one `Function<Object, Object>` per property by `LambdaMetafactory`-binding each `Method`. The `instantiatedMethodType` is `(declaringClass) -> returnType`, so the synthesized bridge auto-boxes primitive returns (`int getAge()` → `Integer`, `boolean isFlagged()` → `Boolean`, `long getCount()` → `Long`) without a per-call reflective boxing dance.

## JPMS `opens` requirement

`MethodHandles.privateLookupIn(beanClass, lookup)` has identical access rules to `setAccessible(true)` — if the bean's package isn't `opens`-exposed to `io.github.eschizoid.telescope`, the cache build throws an `IllegalStateException` with the same message shape Phase 1 uses, pointing the user at the `opens` directive. No additional `opens` are required for the in-tree tests (all fixtures are in the same module).

## Tests

- `BeansTest` gains a `WithPrimitives` fixture (`int getAge()` / `long getCount()` / `boolean isFlagged()`) and two new tests pinning the LMF auto-boxing path through both `readProperty` (cached `Function` direct call) and the lattice-primitive `Getter` (capturing-lambda body delegates to the cached `Function`).
- All pre-existing `BeansTest` cases pass unchanged.
- Full `:core`, `:codegen`, `:lombok` build green via `./gradlew spotlessApply build`.

## Test plan

- [x] `./gradlew :core:spotlessApply :core:test` — green
- [x] `./gradlew spotlessApply build` — green (`:core`, `:codegen`, `:lombok` all pass)
- [x] New primitive-return tests pin the LMF auto-boxing path (`int`/`long`/`boolean`)
- [x] All pre-existing `BeansTest` cases still pass
- [x] Public surface unchanged — no module-info changes, no signature changes

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
@@ -2,6 +2,9 @@ package io.github.eschizoid.telescope.internal;
 
 import io.github.eschizoid.telescope.internal.optics.Getter;
 import io.github.eschizoid.telescope.internal.optics.Lens;
+import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.lang.reflect.AccessibleObject;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

Issue #1, opened 2026-06-04
do not use `org.telescope` namespace


domdorn:
This library is amazing and finally brings optics to Java. 
However, the usage of `org.telescope` is problematic, as "telescope.org" is controlled by someone else.
Using someone elses namespace opens up this project to potential supply-chain attacks. 
Suggestion: Either register a real domain or use `com.github.eschizoid.telescope` instead.

eschizoid:
@domdorn thanks for the heads up, fixed on version 0.2.0

domdorn:
@eschizoid i noticed we now have `com.github` and `io.github` mixed. ideally, we'd settle on one for both. as you're already publishing with `io.github`, that one might be the better one.

eschizoid:
@domdorn I am AFK, if you don't mind opening a PR I’ll review ASAP and then release to maven central.

domdorn:
@eschizoid opened a PR, review/merge at your convenience.

eschizoid:
Thank you so much. Merged and released.

PR #13, opened 2026-06-07
feat(core): Phase 4 — LambdaMetafactory builder writers

eschizoid:
## Summary

Phase 4 of the reflection-removal stack: swap the three `Method#invoke` calls inside `Beans.BuilderWriter` for cached functional invokers built once via `LambdaMetafactory`. Mirrors Phase 1's pattern in `Records.java` (PR #9).

This is a **pure substrate swap**:
- `BeanWriter<P>` sealed interface — unchanged.
- `BuilderWriter<P> implements BeanWriter<P>` declaration — unchanged.
- `construct(String[] names, Function<String, Object> valueByName)` signature — unchanged.
- The bodies of the constructor, `setterFor(...)`, and `construct(...)` swap reflection for direct-MethodHandle-backed functional interfaces.

## Lattice framing

The lattice principle is preserved: nothing in the type system surface changed. The `Beans.BeanWriter` writer family stays the public lattice-facing contract; the swap happens entirely inside `BuilderWriter`'s body. No public API change, no module-info change, no test-only fixture change. `:lombok` integration tests — which exercise `BuilderWriter` end-to-end through real Lombok-generated `@Builder` classes — pass without modification.

## Three sub-swaps

1. **Static `builder()` factory → `Supplier<Object>`** captured once in the constructor via LMF over the zero-arg static factory's `MethodHandle`. Hot-path becomes a single virtual `Supplier.get()`.

2. **Setter (lazily per name) → `BiFunction<Object, Object, Object>`** captured via LMF and cached in a `ConcurrentHashMap<String, BiFunction<…>>`. Two setter shapes are supported:
   - **Fluent setters** (return the builder type, e.g. `Builder setName(String)` returns `this`): bound directly as a `BiFunction`. The returned builder is ignored at the call site. LMF accepts the direct `MethodHandle` as-is.
   - **Void setters** (return `void`, e.g. `void setName(String)`): bound via a `BiConsumer` LMF call site (whose SAM return is `void` — a direct match), then wrapped as a `BiFunction` returning the builder passed in. This restores behavioral parity with the pre-LMF `Method#invoke` path, which silently discarded the return value for both shapes.
   - **Primitive auto-unboxing.** For setters like `setScore(int)`, the LMF `instantiatedMethodType` parameter is pinned to `Integer.class` (the wrapper). LMF inserts a single auto-unbox adapter in the synthesized bridge — the call site still passes a boxed `Integer`, the bridge unboxes once before invoking the primitive setter.

3. **`build()` → `Function<Object, Object>`** captured once in the constructor via LMF over the builder type's `build()` `MethodHandle`. Hot-path becomes a single virtual `Function.apply(builder)`.

After the first call per setter, dispatch is a single virtual call the JIT inlines — no `Method#invoke`, no per-call argument array, no access-check.

## JPMS opens requirement

Each LMF binding requires `MethodHandles.privateLookupIn(target, lookup())` on both the POJO class and its builder type. This is the **same JPMS constraint** as the previous `setAccessible(true)` path — fully-public same-module targets work as-is; closed-package targets need an `opens` directive on the consumer's `module-info.java`. The thrown `IllegalStateException` spells out the exact directive.

## Test plan

- [x] `./gradlew :core:spotlessApply :core:test` — passes (existing `BeansTest$Builders` cases + new pins for primitive auto-unboxing, cache reuse, void-setter support, and dispatch failure propagation)
- [x] `./gradlew :lombok:test` — passes end-to-end `BuilderUserPath.start().email().update(...)` through real Lombok-generated `@Builder`
- [x] `./gradlew spotlessCheck build --rerun-tasks` — passes
- [x] No public API changes; no module-info changes

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
@@ -15,7 +15,9 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * Reflection-based machinery for navigating and constructing JavaBeans-style POJOs. Powers {@code
@@ -244,6 +246,27 @@ public final class Beans {
         e
       );
     }
+  }
+
+  /**
+   * Box a primitive type to its wrapper; non-primitives pass through unchanged. Required for the
+   * LambdaMetafactory {@code instantiatedMethodType}: a SAM parameter typed {@code Object} can't
+   * directly match an instantiated primitive {@code int} (validator rejects "int is not a subtype
+   * of class java.lang.Object"); the wrapper class is the bridge type, and the metafactory
+   * generates the corresponding unbox.
+   */
+  private static Class<?> wrap(final Class<?> c) {
+    if (!c.isPrimitive()) return c;
+    if (c == int.class) return Integer.class;
+    if (c == long.class) return Long.class;
+    if (c == double.class) return Double.class;
+    if (c == float.class) return Float.class;
+    if (c == boolean.class) return Boolean.class;
+    if (c == byte.class) return Byte.class;
+    if (c == short.class) return Short.class;
+    if (c == char.class) return Character.class;
+    if (c == void.class) return Void.class;
+    throw new IllegalStateException("Unknown primitive type: " + c);
   }
 
   private static Map<String, Method> scanGetters(final Class<?> cls) {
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

eschizoid:
Rebased onto current main (now includes Phases 1+2+3 and the PR #16 inherited-accessor fix). This force-push folds in the following follow-ups addressing the Copilot inline review:

**Copilot inline #1 — \`construct\` error wrap regression (line 606)**: \`BuilderWriter.construct\` now pre-resolves all setter \`BiFunction\`s outside the execution try/catch (so resolution-time \`IllegalArgumentException\` for an unknown name still leaks through unwrapped, preserving the existing test contract), then wraps any \`RuntimeException\` from \`builderSupplier.get()\` / setter \`apply\` / \`buildFn.apply\` in \`new RuntimeException("Failed to build " + cls.getName() + " via its builder", re)\`. Matches the pre-LMF shape.

**Copilot inline #2 — void-returning setter fail-fast (line 657)**: \`buildSetterInvoker\` now rejects \`setter.getReturnType() == void.class\` with a descriptive \`IllegalStateException\` ("matched builder method '…' on … returns void; the builder strategy requires a fluent setter …"). Pinned by new test \`builderRejectsVoidSetter\` against the \`BuilderWithVoidSetter\` fixture.

**Forward-ported PR #16 inherited-accessor fix**: \`buildSetterInvoker\`, \`buildBuilderSupplier\`, and \`buildBuildFn\` all now use \`*.getDeclaringClass()\` for \`privateLookupIn\`, the \`instantiatedMethodType\` receiver, and the JPMS-opens error message — same correctness guarantee Phases 2/3 got. Removed the local \`privateLookup\` helper in favor of the shared \`Beans.privateLookupOrThrow(declaringClass, ownerClass, accessorKind)\` from PR #16.

**Refactor — wrap dedup**: Promoted the primitive-to-wrapper \`wrap\` helper to a Beans-level private static (\`SettersWriter\` and \`BuilderWriter\` had duplicates). One source of truth for the LMF instantiatedMethodType primitive bridge.

**New tests**: \`builderRejectsVoidSetter\` (pins the void-setter fail-fast), \`builderConstructWrapsFailure\` (pins the new RuntimeException wrap shape).

Full build green: \`./gradlew :core:test :lombok:test\`.

eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #14, opened 2026-06-07
feat(core): Phase 5 — LambdaMetafactory rebuild path (ctor + fields + Records.construct)

eschizoid:
## Summary

Phase 5 of the reflection-removal stack — close the rebuild half. Swaps `Constructor.newInstance` and `Field.set` for cached invokers across the three rebuild sites the spec covers. **Lattice contracts unchanged**: `BeanWriter<P>` stays sealed, `FieldsWriter` / `ConstructorWriter` declarations stay byte-identical at the type level, public `Records.construct(Class<R>, Function<String, Object>)` signature unchanged. Only the substrate changes — reflective per-call dispatch → cached invokers built once per class at cache-warm time.

## The three swaps

1. **`Records.RecordInfo`** — extended with a `Function<Object, Object> ctorFn` over an `asSpreader`-wrapped canonical-constructor `MethodHandle`. Both `Records.construct(...)` (the spec-named target) and the per-component `updateField` rebuild now route through it instead of `ctor.newInstance(args)`. The `Constructor<?>` field is retained for its metadata.

2. **`Beans.ConstructorWriter`** — replaces `ctor.newInstance(args)` with a cached `Function<Object, Object>` over an `asSpreader`-wrapped `MethodHandle`. Same `-parameters`-vs-positional name resolution preserved.

3. **`Beans.FieldsWriter`** — swaps `ctor.newInstance()` for a `LambdaMetafactory`-built `Supplier<Object>` (no-arg ctor handle is direct, LMF accepts it) and `Field.set` for a cached `BiConsumer<Object, Object>` per field over an `unreflectSetter` `MethodHandle`.

## The asSpreader trick (and why it doesn't use LMF for the ctor)

`LambdaMetafactory` cannot bridge a multi-arg constructor directly to `Function<Object, Object>`. The standard workaround is `handle.asSpreader(Object[].class, arity)` which converts `(T1,...,Tn) → R` into `(Object[]) → R`. However LMF rejects the resulting **non-direct adapter** with `"MethodHandle(Object[])R is not direct or cannot be cracked"`. The constructor invoker therefore wraps the spread `MethodHandle` in a tiny capturing lambda that calls `invokeExact` on it. Same pattern is used for the field setter — LMF also rejects `putField` handles (`"Unsupported MethodHandle kind: putField"`). `MethodHandle` invocation through a `final` field still skips the per-call access check and the varargs allocation that `Constructor.newInstance` / `Field.set` pay, with the JIT inlining `invokeExact` through the captured reference.

Spec language (`LambdaMetafactory.metafactory(...) ... spread, ...`) doesn't compile under the JDK; the cached-MH path is the standard JDK alternative and the javadoc on every swap site records why LMF was bypassed there. The no-arg constructor in `FieldsWriter` still goes through LMF (direct handle, no spreader needed).

## JPMS opens guidance

Preserved at cache-warm time on the `privateLookupIn` / `setAccessible` call sites. Mirrors the Phase 1 reader pattern. Same JPMS constraint as the previous reflection path — no worse, identical messaging.

## After this PR lands

Hot path on the rebuild side now has **zero `Constructor.newInstance` and zero `Field.set`** for both records and beans. `setAccessible` / `privateLookupIn` run once per class at cache-warm time only. The `SettersWriter` / `BuilderWriter` / getter / `readProperty` paths remain untouched — they're owned by sibling parallel PRs.

## Test plan

- [x] `./gradlew :core:spotlessApply :core:test` — 266 tests pass.
- [x] `./gradlew spotlessApply build` — full multi-module build green (`:core`, `:codegen`, `:lombok`, `:benchmarks`).
- [x] New tests in `BeansTest`: `Constructors#constructorAutoUnboxesPrimitives` and `Fields#fieldsAutoUnboxesPrimitives` pin the multi-primitive (`int`/`long`/`double`/`boolean`) auto-unbox path through both the spread-MH ctor invoker and the cached-MH field setter.
- [x] All existing record-based tests (`TelescopeTest`, `DeepMappingTest`, `OpticLawsTest`, etc.) and all bean-based tests still pass — they transitively exercise the new rebuild path on every `.update(...)` / mapping run.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
@@ -3,6 +3,7 @@ package io.github.eschizoid.telescope.internal;
 import io.github.eschizoid.telescope.internal.optics.Getter;
 import io.github.eschizoid.telescope.internal.optics.Lens;
 import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
 import java.lang.reflect.AccessibleObject;
@@ -16,6 +17,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * Reflection-based machinery for navigating and constructing JavaBeans-style POJOs. Powers {@code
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

eschizoid:
Rebased onto current main (now includes Phases 1+2+3 and PR #16's inherited-accessor fix). This force-push folds in the following follow-ups addressing every Copilot inline review on this PR:

**FieldsWriter.construct error wrap** — \`construct(...)\` now wraps the LMF \`Supplier\`-built no-arg ctor invocation with the same \`RuntimeException("Failed to instantiate " + cls.getName(), …)\` shape the pre-LMF \`Constructor#newInstance\` path used. NPE / ISE from inside the bean's no-arg ctor body keep the class-context prefix instead of escaping raw.

**FieldsWriter "cached spread MethodHandle" wording (javadoc + inline)** — \`buildFieldSetter\` uses \`MethodHandle.asType\` to adapt a \`(receiver, value) -> void\` handle to \`(Object, Object) -> void\` — not \`asSpreader\`. Class-level javadoc and the inline comment both now say so.

**ConstructorWriter raw \`setAccessible(true)\`** — dropped. Access flows through \`privateLookupOrThrow(declaringClass, cls, "CONSTRUCTOR strategy")\` so JPMS failures surface through the tailored opens-pointing \`IllegalStateException\` instead of the lower-context \`InaccessibleObjectException\` from raw \`setAccessible\`.

**Records.updateField context loss** — wraps the shared \`ctorFn.apply(args)\` call with the rebuild-specific \`"Failed to rebuild " + cls.getSimpleName()\` message, restoring the context lost when routing through the generic \`buildCtorFn\` error path. Useful for debugging \`Records.with(...)\` / lens-driven single-field updates.

**Records.RecordInfo.of raw \`setAccessible\`** — dropped (same reasoning as ConstructorWriter).

**Forward-ported PR #16 inherited-accessor fix** — \`ConstructorWriter.buildCtorFn\` pins the lookup to \`ctor.getDeclaringClass()\` for synthetic same-arity ctors declared on a parent type whose package may be opened differently from \`cls\`'s.

**MethodHandle import cleanup** — all \`java.lang.invoke.MethodHandle\` FQN references in code / javadoc / comments now use the short name; one canonical \`import java.lang.invoke.MethodHandle\` at the top.

Full build green: \`./gradlew :core:spotlessApply :core:test\` (266 tests).

PR #15, opened 2026-06-07
chore(core): centralize null-passthrough in Iso lift helpers

eschizoid:
## Summary

- Adds private static `Iso.nullablyMap(source, body)` to centralize the `null in → null out` pass-through used by every `lift*` method.
- Refactors `liftList` / `liftOptional` / `liftSet` / `liftMapValues` through the helper. Each direction was repeating the same `c == null ? null : ...` branch — 8 occurrences across the four container shapes collapse to one helper.
- Suppresses the `OptionalAssignedToNull` / `OptionalUsedAsFieldOrParameterType` IDE inspections on `liftOptional`: the null pass-through is intentional per the existing javadoc (records and beans may legally hold null container references; deep mapping treats them as pass-through), but the generic IDE rule flags it as an anti-pattern.

## Why

Pure cleanup driven by IntelliJ's "Duplicated code fragment (4 lines long)" warnings at `liftList` / `liftSet` and the matching pair of Optional inspections on `liftOptional`. No behavior change — the existing optic-laws tests still cover the round-trip behavior of every lift.

## Test plan

- [x] `./gradlew :core:spotlessApply :core:test` — green
- [x] Iso laws (`OpticLawsTest#isoLiftListLaws`, `isoLiftOptionalLaws`, `isoLiftSetLaws`, `isoLiftMapValuesLaws`) — green
- [x] No public surface change; no impact on `:codegen` / `:lombok` modules

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/internal/optics/Iso.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/internal/optics/Iso.java
@@ -107,6 +107,12 @@ public interface Iso<A, B> extends Lens<A, B>, Prism<A, B> {
     return of(x -> x, x -> x);
   }
 
+  // Centralizes the "null in → null out" pass-through every lift uses — records and beans may
+  // legally hold null container references, and deep mapping treats them as pass-through.
+  private static <C, R> R nullablyMap(final C source, final Function<? super C, ? extends R> body) {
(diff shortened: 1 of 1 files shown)


eschizoid:
Reverted locally — not pursuing this refactor; keeping the squiggle per project mantra.

PR #16, opened 2026-06-07
fix(core): LMF inherited-accessor lookup uses declaring class (PR #11 follow-up)

eschizoid:
## Summary

Post-merge follow-up addressing two Copilot inline comments on [PR #11](https://github.com/eschizoid/telescope/pull/11) about inherited-accessor correctness in the LambdaMetafactory bean dispatch.

Both Phase 2's \`Beans.buildGetterInvokers\` and Phase 3's \`SettersWriter.buildSetterInvoker\` were using \`cls\` (the inheritor's class) for \`MethodHandles.privateLookupIn(...)\` and the JPMS-opens error message. For an accessor inherited from a parent class / interface in a different package or module, the package that needs the \`opens\` directive is the **declaring** class's package, not the inheritor's. Using the inheritor's class for \`privateLookupIn\` would also fail to bind a handle whose declaring package is open to telescope but whose inheritor's package is closed (or vice-versa).

## Changes

1. New private static helper \`Beans.privateLookupOrThrow(declaringClass, ownerClass, accessorKind)\` that throws an \`IllegalStateException\` with a JPMS-opens hint pointing at the **declaring** class's package. When the declaring class differs from the inheritor, the message calls out the inheritance.
2. \`buildGetterInvokers\` now caches one \`Lookup\` per declaring class in a \`HashMap<Class<?>, MethodHandles.Lookup>\` so a single bean with getters spread across multiple declaring classes resolves each correctly. \`instantiatedMethodType\` already used \`getDeclaringClass()\`.
3. \`SettersWriter.buildSetterInvoker\` pins \`privateLookupIn\`, the \`instantiatedMethodType\` receiver, and the error message all to \`setter.getDeclaringClass()\`.

## Tests

New \`ParentBean\` / \`ChildBean extends ParentBean\` fixture pair, plus:

- \`getterOnInheritedAccessor\` — reads inherited \`getId()\` from a \`ChildBean\` through both \`readProperty\` and the lattice-primitive \`Getter\`, plus the child-declared \`getName()\`.
- \`settersOnInheritedSetter\` — round-trip via inherited \`setId(String)\` and child-declared \`setName(String)\`.

## Phases 4 and 5

Their open PRs (#13, #14) carry the same \`cls\`-based pattern in their own LMF code (\`BuilderWriter\`, \`ConstructorWriter\`, \`FieldsWriter\`). I'll forward-port this fix as part of rebasing each onto main.

## Test plan

- [x] \`./gradlew :core:spotlessApply :core:test\` — green (272 tests)
- [x] Two new tests pin inherited-getter and inherited-setter behavior end-to-end
- [x] No public API change

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
@@ -172,41 +172,30 @@ public final class Beans {
     final Map<String, Method> getters
   ) {
     if (getters.isEmpty()) return Map.of();
-    final MethodHandles.Lookup lookup;
-    try {
-      // `privateLookupIn` is needed when the bean's module/package isn't open to the telescope
-      // module; for fully-public types in the same module this is equivalent to a plain
-      // `MethodHandles.lookup()`. Same JPMS constraint as `setAccessible(true)` — no worse than
-      // the previous reflection path.
-      lookup = MethodHandles.privateLookupIn(cls, MethodHandles.lookup());
-    } catch (final IllegalAccessException e) {
-      throw new IllegalStateException(
-        "Cannot access " +
-          cls.getName() +
-          " to build LambdaMetafactory getter invokers. Add 'opens " +
-          cls.getPackageName() +
-          " to io.github.eschizoid.telescope;' to that module's module-info.java.",
-        e
-      );
-    }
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #17, opened 2026-06-07
docs(adr): ADR-0005 — LambdaMetafactory over MethodHandle.invoke

eschizoid:
## Summary

- Adds `docs/adr/0005-lambdametafactory-over-method-handle-invoke.md` to re-litigate ADR-0003 in light of the LambdaMetafactory substrate swap landed across PRs #9, #12, #11, #13, #14.
- Clarifies the distinction ADR-0003 didn't make: raw `MethodHandle.invoke` dispatch (rejected, then and now) vs. LMF-built `Function` / `BiConsumer` / `Supplier` instances (the JIT inlines through these like ordinary method-reference literals).
- Records the perf, JPMS, cache-warmup, and class-load-footprint trade-offs; cross-links to ADR-0004 for why we still keep runtime and codegen as separate strategies.

## Test plan

- [x] `./gradlew spotlessApply` — markdown reflow applied
- [x] `./gradlew spotlessCheck` — passes
- [x] Heading hierarchy and fenced blocks match the existing 0001-0004 ADRs
- [x] No code changes; no build verification needed

### DIFF
--- a/docs/adr/0005-lambdametafactory-over-method-handle-invoke.md
+++ b/docs/adr/0005-lambdametafactory-over-method-handle-invoke.md
@@ -1,0 +1,68 @@
+# ADR-0005: LambdaMetafactory-built functional interfaces, not raw MethodHandle dispatch
+
+**Status:** Accepted (refines [ADR-0003](0003-reflection-over-method-handles.md)) · **Date:** 2026-06-07
+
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #20, opened 2026-06-07
docs(annotations): correct @Focus javadoc — emits <X>Path<R>, not <X>Focus

eschizoid:
## Summary

\`@Focus\`'s class-level javadoc described an old generated shape: a static class named \`<Record>Focus\` with \`public static final Telescope<Record, FieldType>\` constants per component. The processor doesn't emit that shape — it emits a fluent navigator class \`<X>Path<R>\` with one method per record component, plus the full Telescope op-surface forwarded so any hop reads / updates / traverses without reaching for \`.get()\`.

The javadoc-vs-implementation contradiction was the only outright one the v1.0 API stability sweep found. README + CLAUDE.md already document the \`<X>Path<R>\` shape correctly; this PR brings the annotation javadoc in line.

## Changes

- Rewrote the class-level javadoc on \`@Focus\` to describe the actual \`<X>Path<R>\` shape, including the container sub-navigator (\`<X><Cap>Step<R>\`) emission for \`List\` / \`Set\` / \`Map\` / \`Optional\` / \`Iterable\` components and the recursive descent into nested \`@Focus\` types.
- Updated the inline example to show \`UserPath.start().name().update(...)\` and \`UserPath.start().address().city().set(...)\` — both a primitive component read/update and a nested \`@Focus\` drill-down.

## Why now

Part of the v1.0.0 API stability readiness sweep. Docs-vs-implementation drift is the most user-hostile bug class; fixing this one before 1.0 freezes is cheap and the right time.

## Test plan

- [x] \`./gradlew :core:spotlessApply\` — green
- [x] No behavior change — javadoc only
- [x] Inline example compiles in spirit (matches the README walkthrough at the codegen section)

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/annotations/Focus.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/annotations/Focus.java
@@ -7,18 +7,21 @@ import java.lang.annotation.Target;
 
 /**
  * Marks a {@link Record} for codegen by the {@code telescope-codegen} annotation processor. For
- * each annotated record, the processor emits a sibling class {@code <RecordName>Focus} with {@code
- * public static final Telescope<Record, FieldType>} constants per record component, each built via
- * {@link io.github.eschizoid.telescope.Telescope#lens(java.util.function.Function,
- * java.util.function.BiFunction)}.
+ * each annotated record, the processor emits a sibling class {@code <RecordName>Path<R>} — a fluent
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

eschizoid:
Folding this into a single consolidated v1.0 readiness PR (Focus javadoc + .map/.set rename + internal-leak cleanup + Reflective record refactor). Closing here; the Focus javadoc commit will be cherry-picked into the bundle.

PR #22, opened 2026-06-07
chore(core): internal-leak cleanup for v1.0.0 API stability

eschizoid:
## Summary

Tightens the public surface of `telescope-core` before v1.0 freezes the API. None of the targets here are intended for external use; they were left public during 0.x for cross-package convenience inside the module. Tightening now prevents accidental external dependencies on internal shapes. **All changes are compile-time access / javadoc edits — zero runtime behavior change.**

### 1. `Telescope#wrap(Traversal)` and `Telescope#optic()`
Files: `core/src/main/java/io/github/eschizoid/telescope/Telescope.java` (~L135-150, ~L190-204)
- Kept `public`. Expanded the javadoc to mark both methods as **module-internal seams — NOT public API**.
- Why kept public: the `<X>Path<R>` navigators emitted by the codegen processors are compiled in downstream consumer modules, and they call `Telescope.wrap(...)` with the unexported `Traversals.eachIterable()` (`AbstractTelescopeProcessor.java:761`). Making `wrap` package-private would break that generated code.
- Why this is safe: `Traversal` lives in the unexported `internal.optics` package, so external code cannot supply a real argument regardless. `@SuppressWarnings(\"exports\")` retained.

### 2. `From` and `To` no-arg constructors
Files:
- `core/src/main/java/io/github/eschizoid/telescope/conversion/From.java` (~L11)
- `core/src/main/java/io/github/eschizoid/telescope/conversion/To.java` (~L13)

- `To()` → package-private. Caller (`From#to`) lives in the same `conversion` package.
- `From()` → kept `public`. Caller (`Telescope#from`) lives in a different package; expanded javadoc to mark as module-internal.

### 3. `Mapper` public constructor + `PatchEntry` record
`core/src/main/java/io/github/eschizoid/telescope/conversion/Mapper.java` (~L34, ~L42-56)
- Both kept `public` with module-internal javadoc markers (\"**Module-internal seam — NOT public API.**\").
- Why kept public: the only caller (`DeepMap` in `mapping` package) constructs both. Co-locating `DeepMap` into `conversion` would require also moving / exposing the three package-private `Mapping` impl records (`SameTypedTo`, `TypedTransformTo`, `Via`) plus their package-private `fieldIso()` accessors — a far larger leak than the one being closed.
- The `Function<Object, Object>` leak on `PatchEntry.backward` is now documented as part of the internal contract.

### 4. `Mapping` interface — remove `sourceClass` / `targetClass` / `sourceField` / `targetField`
Files:
- `core/src/main/java/io/github/eschizoid/telescope/mapping/Mapping.java` (removed four methods, lines ~74-89 in original)
- `core/src/main/java/io/github/eschizoid/telescope/mapping/MappingInternals.java` (new package-private sealed sibling interface)
- `core/src/main/java/io/github/eschizoid/telescope/mapping/SameTypedTo.java` (added `MappingInternals<A, B>` to permits implementation)
- `core/src/main/java/io/github/eschizoid/telescope/mapping/TypedTransformTo.java` (same)
- `core/src/main/java/io/github/eschizoid/telescope/mapping/Via.java` (same)
- `core/src/main/java/io/github/eschizoid/telescope/mapping/DeepMap.java` (added `internalsOf(Mapping<?, ?>)` helper, switched two call sites to use it)

The four `SerializedLambda`-recovery methods were on the public sealed `Mapping<A, B>` interface, leaking the existence of the recovery machinery to anyone naming `Mapping` directly. Moved to a new **package-private** `MappingInternals<A, B>` sealed sibling interface with the same permit list (`SameTypedTo`, `TypedTransformTo`, `Via`); `DeepMap` casts to `MappingInternals` at the two call sites that need the keying info. Cast is safe because every permitted `Mapping` impl implements `MappingInternals`.

## Module exports
`core/src/main/java/module-info.java` is unchanged; no package shifts.

## Test plan
- [x] `./gradlew :core:spotlessApply :core:test :codegen:test :lombok:test` — green locally on Java 25 / Apple Silicon.
- [x] No `module-info.java` change required (verified).
- [x] Runtime behavior unchanged — every edit is a javadoc, an access modifier, or a refactor of an internal call site.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -133,14 +133,16 @@ public sealed class Telescope<
   }
 
   /**
-   * Wrap an internal optic ({@link Traversal} or any of its subtypes — {@link Iso}, {@link Lens},
-   * {@link Prism}) as a {@code Telescope<S, A>}. Used by the conversion-builder sub-package and the
-   * mapping sub-package to produce telescopes from internally-composed optics without depending on
-   * a package-private constructor.
+   * <b>Module-internal seam — NOT public API.</b> Wrap an internal optic ({@link Traversal} or any
+   * of its subtypes — {@link Iso}, {@link Lens}, {@link Prism}) as a {@code Telescope<S, A>}.
    *
-   * <p>The {@link Traversal} type itself lives in the unexported {@code internal.optics} package,
-   * so consumers of the module cannot construct one to pass in — this factory is effectively
-   * module-internal even though declared public.
+   * <p>This method is declared {@code public} purely so the {@code conversion} and {@code mapping}
+   * sub-packages — and the {@code <X>Path<R>} navigators emitted by the codegen processors — can
+   * construct {@code Telescope} instances from internally-composed optics. The {@link Traversal}
+   * type lives in the unexported {@code internal.optics} package, so module consumers cannot supply
+   * a real argument. Treat this signature as part of the module's internal contract: it may change
+   * or disappear without a deprecation cycle. External code must use the documented entry points
+   * ({@link #of(Class)}, {@link #ofBean(Class)}, {@link #lens}, {@link #from(Class)}, etc.).
    */
   @SuppressWarnings("exports") // Intentional: Traversal is module-internal; users can't construct one.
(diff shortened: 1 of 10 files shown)


eschizoid:
@copilot please review

eschizoid:
Folded into #24 (v1.0 readiness bundle). Closing here.

PR #23, opened 2026-06-07
feat(core)!: rename .map(Accessor) → .mapField(Accessor) and .set(Accessor) → .setField(Accessor)

eschizoid:
## Summary

Two instance navigation methods on `Telescope<S, A>` are renamed for 1.0 API stability — their old names collided with sibling APIs that do conceptually different things:

| Old (1.0-pre)                                | New (1.0)                                          | Why                                                                                                                                                                                |
| -------------------------------------------- | -------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `Telescope#map(Accessor<A, Map<K,V>>)`       | `Telescope#mapField(Accessor<A, Map<K,V>>)`        | Collides with the static deep-conversion factory `Telescope.map(Class<A>, Class<B>, Mapping<A,B>...)`. They share a verb but do conceptually different things.                     |
| `Telescope#set(Accessor<A, Set<X>>)`         | `Telescope#setField(Accessor<A, Set<X>>)`          | Cognitive collision with the write terminal `Telescope#set(S, A)`. Different arg types, but enough shared-verb load that javac disambiguation is doing work the reader shouldn't. |

`Telescope#list(Accessor)` and `Telescope#optional(Accessor)` are left **unchanged** — they don't collide with anything.

### Before / after

```java
// Before
final SetPath<Cart, Item> items = Telescope.of(Cart.class).set(Cart::items);
final MapPath<TagMap, String, Tag> tags = Telescope.of(TagMap.class).map(TagMap::tagsByKey);

// After
final SetPath<Cart, Item> items = Telescope.of(Cart.class).setField(Cart::items);
final MapPath<TagMap, String, Tag> tags = Telescope.of(TagMap.class).mapField(TagMap::tagsByKey);
```

### Affected files

- `core/.../Telescope.java` — the two method declarations + javadocs (with a note explaining the rename rationale).
- `core/.../TypedContainerPathTest.java` — the two tests that exercised the renamed methods + their `@DisplayName`s.
- `README.md` — the DSL-surface table and the typed-container-leaves cookbook example.

### What this PR is *not*

- **No `@Deprecated` aliases.** We're at 1.0 freeze posture; the right move is a clean rename, not a long deprecation window. Migration is a mechanical find/replace.
[...]

PR #21, opened 2026-06-07
chore(internal): collapse Reflective to record with function-typed fields

eschizoid:
## Summary

- Collapse `internal/Reflective.java` from a public interface + two 30-line anonymous-class singletons (`RECORDS`, `BEANS`) + an anonymous `beansWithHints` impl into a single `record` whose five fields are per-side `Function` / `BiFunction` references (~186 LOC -> ~167 LOC, structural duplication gone).
- Each singleton becomes a one-liner: a `new Reflective(Records::componentNames, Records::componentType, Records::read, Reflective::constructRecord, name -> name)` and a `new Reflective(Beans::propertyNames, ...)` — instead of two parallel anonymous classes overriding five methods each.
- Forwarding instance methods (`names`, `genericType`, `read`, `construct`, `normalize`) preserve the prior interface-method calling convention so every `srcRefl.names(cls)` / `srcRefl.read(v, n)` / `srcRefl.construct(cls, fn)` site in `DeepMap` and `Mapper` compiles unchanged.
- `structuralIso` flips from a `default` method to a plain instance method with the same body and semantics.

## No behavior change

- Internal-only file (`io.github.eschizoid.telescope.internal`, not exported by `module-info.java`) — the `interface` -> `record` API-shape change is invisible to consumers, and a grep confirmed no external implementers of `Reflective` exist in this codebase.
- `Records.java` / `Beans.java` substrate untouched.
- `DeepMap.java` / `Mapper.java` call sites unchanged — the forwarding instance methods absorb the entire diff.
- The `Reflective.RECORDS` / `Reflective.BEANS` identity comparisons in `DeepMap#slot(Reflective)` still work — both are still static final singletons with stable identity.

## Verification

Full multi-module build is green:

```
./gradlew :core:spotlessApply :core:test :codegen:test :lombok:test build
BUILD SUCCESSFUL in 5m 57s
```

`:core:test` (278 tests), `:codegen:test`, and `:lombok:test` all pass. The Lombok module exercises `Reflective.BEANS` end-to-end via the generated `Path` navigators on `@Data` / `@Builder` / `@Value` fixtures.

## Test plan

- [x] `./gradlew :core:test` passes
- [x] `./gradlew :codegen:test` passes
- [x] `./gradlew :lombok:test` passes
- [x] `./gradlew build` passes (full multi-module green)
- [x] No call-site changes outside `Reflective.java` for the refactor itself (one drive-by Spotless javadoc reflow in `Beans.java` picked up by `:core:spotlessApply`).

PR #24, opened 2026-06-07
feat(core)!: v1.0.0 API stability bundle — @Focus javadoc + Reflective refactor + internal-leak cleanup + .map/.set rename

eschizoid:
## Summary

Consolidates four API-stability changes for v1.0.0 readiness into a single PR. Supersedes individual PRs #21 (Reflective record refactor), #22 (internal-leak cleanup), #23 (.map/.set rename), and the closed #20 (@Focus javadoc fix).

### 1. \`docs(annotations): correct @Focus javadoc\` — \`<X>Path<R>\`, not \`<X>Focus\`

Class-level javadoc described an old generated shape that was never the actual output. Rewrote to describe the real fluent navigator \`<X>Path<R>\` shape, including the container sub-navigators (\`<X><Cap>Step<R>\`) and recursive descent into nested \`@Focus\` types. Updates the inline example to show \`UserPath.start().name().update(...)\` and \`UserPath.start().address().city().set(...)\` — both a primitive component update and a nested \`@Focus\` drill-down.

The only outright doc-vs-implementation contradiction the API stability sweep found.

### 2. \`chore(internal): collapse Reflective to a record with function-typed fields\`

Pure internal cleanup, **no behavior change**. \`Reflective\` was a public interface with two ~30-line anonymous-class singletons (\`RECORDS\`, \`BEANS\`) plus an anonymous \`beansWithHints\` impl, all heavy on delegation boilerplate.

Collapsed to one \`record\` with five function-typed fields plus forwarding instance methods that preserve every existing call-site shape. ~150 lines → ~30 lines. \`DeepMap.java\`, \`Mapper.java\`, and \`ReflectiveStructuralIsoTest.java\` required zero changes.

### 3. \`chore(core): internal-leak cleanup for v1.0.0 API stability\`

Tightened public-surface members that should never have been external:

- **\`Mapping#sourceClass\` / \`targetClass\` / \`sourceField\` / \`targetField\`** — fully removed from the public \`Mapping\` interface. New package-private sealed sibling \`MappingInternals<A, B>\` with same permit list (\`SameTypedTo\`, \`TypedTransformTo\`, \`Via\`); all three impls now \`implements Mapping<A, B>, MappingInternals<A, B>\`. \`DeepMap\` casts via new \`internalsOf(...)\` helper.
- **\`To()\` constructor** — made package-private (caller \`From#to\` is same-package).

The following stayed public but got stronger "module-internal seam — NOT public API" javadoc, after the agent found that closing them would cascade into larger leaks elsewhere:

- \`Telescope#wrap(Traversal)\` / \`Telescope#optic()\` — codegen forwarders in downstream consumer modules call \`Telescope.wrap(...)\` (\`AbstractTelescopeProcessor.java:761\`); closing would break downstream builds.
- \`From()\` ctor — \`Telescope#from\` lives in a different package and needs constructability.
- \`Mapper\` ctor and \`Mapper.PatchEntry\` — \`DeepMap\` constructs them cross-package; co-locating would cascade through three impl records plus their \`fieldIso()\` accessors.

### 4. \`feat(core)!: rename .map(Accessor) → .mapField, .set(Accessor) → .setField\` (BREAKING)

Two instance methods on \`Telescope<S, A>\` get renamed because their names collide with sibling static factories doing conceptually different things:

- \`.map(Accessor<A, Map<K,V>>)\` → \`.mapField(...)\` — disambiguates from the deep-conversion factory \`Telescope.map(Class<A>, Class<B>, Mapping<A,B>...)\`.
- \`.set(Accessor<A, Set<X>>)\` → \`.setField(...)\` — disambiguates from the write terminal \`.set(S, A)\`.

\`.list(Accessor)\` and \`.optional(Accessor)\` left unchanged (no collisions).

**Codegen unaffected:** \`AbstractTelescopeProcessor#emitTelescopeForwarders\` only forwards the write terminal \`set(R, value)\`, never the accessor-taking \`set(getter)\`/\`map(getter)\` overloads. Container subclasses (\`ListPath\`, \`SetPath\`, \`MapPath\`, \`OptionalPath\`) inherit the renamed methods from the parent; their own typed terminals (\`.each()\`, \`.values()\`, \`.present()\`) untouched.

**Updated:** README DSL surface table + Typed-container cookbook section, tests in \`TypedContainerPathTest\`.

## Verification

\`./gradlew :core:spotlessApply :core:test :codegen:test :lombok:test\` — green, 278+ tests pass.

## Closes

- #21 (folded in)
- #22 (folded in)
- #23 (folded in)
- The closed #20 (folded in)


eschizoid:
@copilot please review

PR #25, opened 2026-06-07
chore(codegen): drop redundant TraversalShape.containerKind() accessor

eschizoid:
## Summary

\`TraversalShape\` is a record with a \`containerKind\` component — Java auto-generates an identical accessor. The explicit \`public String containerKind() { return containerKind; }\` override in \`AbstractTelescopeProcessor\` was pure noise (IntelliJ's "Method overrides identical method" inspection flagged it).

Folded the override's docstring paragraph (which listed the \`"list"\`/\`"set"\`/\`"map"\`/\`"optional"\`/\`"iterable"\` values \`containerKind\` takes) into the record's class-level javadoc so no information is lost.

## Why

Static-analysis pass on \`AbstractTelescopeProcessor\` ahead of v1.0; this was the one real warning worth acting on inline.

## Test plan

- [x] \`./gradlew :codegen:spotlessApply :codegen:test :lombok:test\` — green
- [x] No behavior change — same accessor signature, same return value, just no longer manually written

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
@@ -152,21 +152,14 @@ public abstract class AbstractTelescopeProcessor extends AbstractProcessor {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #26, opened 2026-06-07
docs(adr): ADR-0006 — codegen ↔ runtime 1:1 lookup via sibling metadata holder

eschizoid:
## Summary

Records the hybrid plan that closes the "reflection still in the runtime" question for \`@Focus\` / \`@BeanFocus\` users end-to-end. After [ADR-0005](https://github.com/eschizoid/telescope/blob/main/docs/adr/0005-lambdametafactory-over-method-handle-invoke.md) (LambdaMetafactory substrate), per-call dispatch is at codegen parity (~15-35 ns/op) but metadata-probing reflection remains. ADR-0006 documents how the codegen processors will emit a sibling \`<X>Telescope\` class with per-field \`Telescope<X, FieldType>\` constants the runtime probes for first, falling through to today's LMF path when absent.

## What's in the bundle

- **\`docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md\`** — the ADR itself. Resolves all ten open design questions from the scout report: holder location (top-level public class in user's package, \`<X>Telescope\` naming), what the holder exposes (typed \`Telescope<X, FieldType>\` constants — no \`Type[]\` arrays), cache type (\`ClassValue<Optional<HolderRef>>\` matching the existing pattern), lookup-miss policy (throw with diagnostic, not silent fallback), generics-with-bounds (codegen-time error), \`@Bridge\` interaction (leave it alone), Lombok handling (existing \`processingOver()\` deferral). Phased rollout: A emit → B runtime probe → C deep-mapping use.
- **\`benchmarks/README.md\`** — extends the LMF-tier descriptive table with three \`_methodInvoke\` baseline rows (\`recordComponentRead_methodInvoke\`, \`beanGetterRead_methodInvoke\`, \`beanSetterDispatch_methodInvoke\`). Results table has placeholder cells pointing at the in-flight 5+10×3 JMH run. The config is documented in-band so anyone can reproduce.

## Why this lands as a standalone PR

The implementation work (Phases A, B, C) is sizable and will land as separate PRs. The ADR pins the design decisions first so the implementation PRs can reference it without re-litigating choices in PR bodies.

## Test plan

- [x] No code change; ADR is markdown + README table extension
- [x] Spotless markdown formatter clean
- [x] In-flight: 5+10×3 JMH run to fill the \`_methodInvoke\` placeholder cells in \`benchmarks/README.md\`. Will land as a separate follow-up PR after the run completes.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -41,14 +41,17 @@ call — without composing through a multi-level optic. They measure the residua
 `Function<Object, Object>` / `BiConsumer<Object, Object>` synthesized once via `LambdaMetafactory` against a directly
 inlined Java call. The delta is the per-dispatch overhead the LMF wrapper adds on top of an inlined accessor / setter.
 
-| Benchmark                        | What it does                                                                                          |
-| -------------------------------- | ----------------------------------------------------------------------------------------------------- |
-| `recordComponentRead_lmf`        | `Records.read(record, "name")` — Phase 1 LMF reader hot path.                                         |
-| `recordComponentRead_handRolled` | Direct `record.name()` — record accessor baseline.                                                    |
-| `beanGetterRead_lmf`             | `Beans.readProperty(pojo, "name")` — Phase 2 LMF getter hot path.                                     |
-| `beanGetterRead_handRolled`      | Direct `pojo.getName()` — bean getter baseline.                                                       |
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #29, opened 2026-06-07
feat(examples): add :examples subproject — first downstream consumer

eschizoid:
## Summary

- Adds a new `:examples` Gradle subproject — a small Java 25 application that exercises every public entry point of the library in `main()` and prints a human-readable "before / after" trace for every operation.
- Implements PLAN.md §3 ("the unified factory in anger") by making the examples module the first downstream consumer of the library. The module compiles against the public API and runs as a smoke test; an API break in `:core` / `:codegen` / `:lombok` now trips a compile error here before it could reach a real user.
- Includes a small `chore(spotless)` commit that normalizes markdown table padding in `benchmarks/README.md` and `docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md` so `./gradlew check` stays green. Both files already had pending spotless violations on `main`; the fix is pure spotlessApply output, no semantic changes.

## Entry points covered

One demo file per surface, all driven by `Main.java`:

- **RuntimeNavigation** — `Telescope.of(Class)` / `ofBean(Class)` / `fieldByName(String)` / `fieldByName(String, Class<B>)`
- **ContainerNavigation** — typed `.list / .setField / .mapField / .optional` + `each / each / values / present` terminals
- **SealedAndFilter** — `.as(Class)` Prism narrowing (hit + miss-as-no-op), `.filter(Predicate)`, `each + as` over a `List<sealed>`
- **MultiEdit** — `Telescope.all(over(...), over(...), over(...))` with reuse across sources
- **Indexed** — `withIndex()`, `updateIndexed`, `toListIndexed`
- **EffectfulUpdate** — `updateAsync` (same-thread + Executor overload), `updateOptional` (short-circuit), `updateEither` (Right + Left), `updateValidated` (error accumulation)
- **Conversion** — `from(A).to(B).using(fwd, back)` and its composition with `.then(...)`
- **DeepMapping** — `Telescope.map / mapper` with same-name auto-recursion, `to(...)` renames, `to(...)` typed transforms, `via(...)` pre-built nested mappers, `Mapper.patch` sparse overlay, `WriteHint.writeBean` for an immutable all-args-only POJO
- **Codegen** — `@Focus` records, `@BeanFocus` POJO, `@Bridge` with the `as<Target>()` hop landing on a `@Focus`-annotated target Path
- **Lombok** — `@Data` and `@Builder` fixtures whose `<X>Path<R>` navigators are emitted by `telescope-lombok`'s `LombokFocusProcessor`. Loaded reflectively because the processor emits on `processingOver()` (last round) and the generated FQNs aren't importable by sibling sources in the same compile task — same constraint as the `:lombok` integration test.

## Gradle wiring

- `java` + `application` plugins; `JavaLanguageVersion 25`
- `annotationProcessor(project(":codegen"))` + `annotationProcessor(project(":lombok"))` + `annotationProcessor(libs.lombok)` + `compileOnly(libs.lombok)`
- `mainClass = io.github.eschizoid.telescope.examples.Main`
- No test sources — `./gradlew :examples:run` IS the smoke test

## Test plan

- [ ] `./gradlew :examples:build` succeeds — verifies the public API is compilable from a downstream consumer's point of view.
- [ ] `./gradlew :examples:run` prints the full demo trace ending with `All demos completed`.
- [ ] `./gradlew check` (CI command) is green across all subprojects.
- [ ] No reformat of existing modules beyond the spotless table-alignment fix called out above.
- [ ] Spot-check that a contrived API regression (e.g. rename `.field` to `.foo` in `:core`) would trip a compile error in `examples/src/main/java/...` — by inspection of the call sites in each demo file.


eschizoid:
@copilot please review

PR #30, opened 2026-06-07
chore(core): clear IntelliJ inspections across :core public surface

eschizoid:
## Summary

Sweep of every substantive \`:core\` Java file via \`mcp__idea__get_file_problems\`. Six real findings actioned; the rest are documented false-positives or by-design.

## What changed

1. **\`Telescope.java\` javadoc reference** — \`{@link #map(Class, Class, Mapping[])}\` broke after the internal-leak cleanup moved \`Mapping\` to the package-private \`mapping/\` subpackage. Updated to \`{@link #map(Class, Class, io.github.eschizoid.telescope.mapping.MapStep...)}\`.
2. **\`Telescope.updateValidated\` / \`updateEither\`** — redundant explicit type args \`ValidatedK.<E>forError()\` / \`EitherK.<E>forLeft()\` dropped. The bound \`<E>\` on the surrounding method pins them; inference is enough.
3. **\`Beans.ConstructorWriter.cls\`** — private field assigned in the ctor but never read elsewhere. Dropped both the field and the \`this.cls = cls;\` assignment. The local \`cls\` parameter stays — used in the constructor loop + error messages.
4. **\`Beans.BuilderWriter.buildBuildFn\`** — \`cls\` parameter unused (only mentioned in comments describing what we *don't* use it for). Dropped from the signature; updated the single call site.
5. **\`Iso.liftOptional\`** — \`@SuppressWarnings({\"OptionalAssignedToNull\", \"OptionalUsedAsFieldOrParameterType\"})\` added. The null pass-through behavior is documented in the existing javadoc (records/beans may legally hold null \`Optional\` references; deep mapping treats nulls as pass-through). The warnings flag a real \`Optional\` anti-pattern but one this lift deliberately allows.
6. **\`DeepMap.groupOverridesByPair\`** — \`computeIfAbsent\` unused key lambda parameter renamed \`k\` → \`_\` (Java 22+ unnamed parameter).

## What's deliberately left

- **\`Class<S>\` type-witness parameters** on \`of\`, \`ofBean\`, \`from\`, \`fieldByName\` — common Java idiom; the \`Class\` is used for type inference even when not stored.
- **\`Traversal\` / \`Iso\` \"not exported\"** warnings on documented module-internal seams — \`@SuppressWarnings(\"exports\")\` already in place where needed.
- **\`Validated.isInvalid\` / \`fold\` / \`combine\` \"never used\"** — public sealed-type API surface.
- **\`Records\` / \`Beans\` \`Method#invoke\` javadoc references** — IntelliJ false positive on signature-polymorphic methods; \`javac\` accepts.
- **\`Beans.ClassValue.computeValue\` \`@NotNull\` warnings** — the project doesn't use JetBrains nullness annotations anywhere else.
- **Container-Path inconvertible-types** in \`asList\` / \`asSet\` / \`asMap\` / \`asOptional\` — IntelliJ generic-projection bug; the compiler accepts.
- **\`Iso.liftList\` / \`liftSet\` 4-line duplicated patterns** — intentional; different concrete types, can't share.

## Test plan

- [x] \`./gradlew :core:spotlessApply :core:test :codegen:test :lombok:test\` — green
- [x] No behavior change — substrate-only cleanups
- [x] Pure dead-code removal (3 unused symbols) + 1 documented suppression + 1 javadoc fix + 1 inference cleanup + 1 unnamed-param refresh

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -475,8 +475,9 @@ public sealed class Telescope<
    * {@link MapPath} carrying the map type for later {@link MapPath#values()} navigation.
    *
    * <p>Named {@code mapField} (rather than {@code map}) to disambiguate from the sibling static
-   * deep-conversion factory {@link #map(Class, Class, Mapping[])} — those do conceptually different
-   * things and share the same verb otherwise.
+   * deep-conversion factory {@link #map(Class, Class,
+   * io.github.eschizoid.telescope.mapping.MapStep...)} — those do conceptually different things and
+   * share the same verb otherwise.
    */
   public <K, V> MapPath<S, K, V> mapField(final Accessor<A, Map<K, V>> getter) {
     final Lens<A, Map<K, V>> lens = fieldOptics.lensFor(getter);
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #31, opened 2026-06-07
chore(benchmarks): drop _methodInvoke ns/op + honest perf framing

eschizoid:
## Summary

Fills in the \`_methodInvoke\` placeholder cells in \`benchmarks/README.md\` with measured numbers from a proper 5 warmup + 10 measurement × 3 fork JMH run. Also rewrites the framing around the LMF perf claim to match what the numbers actually show.

## The surprising finding

**At the single-step dispatch primitive level, LMF and \`Method.invoke\` are roughly comparable**, not 5-10× apart as the older directional capture suggested. On the bean getter and setter, \`Method.invoke\` is actually a touch faster in this micro-benchmark (error bars overlap, but the means are within ~2-10 ns of each other). Modern HotSpot has been aggressively optimising \`Method.invoke\` for years; the historical \"100-260 ns per call\" reflection cost no longer holds at this scale for trivial accessors after warmup.

This **doesn't invalidate ADR-0005** — the case for the LMF substrate was always structural: removing per-call \`Object[]\` allocation, eliminating the access-check, and giving the JIT a normal SAM call site to inline through composed lens chains. Those wins don't show up in JMH benchmarks that time a single isolated read; they show up at the inlining boundaries of bigger workloads. The micro-benchmarks just confirm that at the primitive level, both substrates are JIT'd to similar costs after warmup.

The README \"Honest read of the numbers\" section now says this explicitly rather than implying a 5-10× per-call win.

## Numbers (5+10×3 fork, JDK 25, Apple Silicon)

| Row | ns/op | ±error |
|---|---:|---:|
| \`recordComponentRead_handRolled\`   |  1.334 |  ± 0.845 |
| \`recordComponentRead_lmf\`          |  9.394 |  ± 4.190 |
| \`recordComponentRead_methodInvoke\` | 18.260 |  ± 7.588 |
| \`beanGetterRead_handRolled\`        |  1.008 |  ± 0.354 |
| \`beanGetterRead_lmf\`               | 13.083 |  ± 5.309 |
| \`beanGetterRead_methodInvoke\`      | 10.660 |  ± 1.635 |
| \`beanSetterDispatch_handRolled\`    |  3.203 |  ± 0.954 |
| \`beanSetterDispatch_lmf\`           | 22.285 |  ± 9.487 |
| \`beanSetterDispatch_methodInvoke\`  | 12.001 |  ± 4.192 |

Error bars are wide (±20-70% of the mean) because the operations are 1-22 ns and JMH's \`Blackhole\` overhead is in the same order of magnitude — direction is reliable, absolute values aren't tight. The JMH output's own warning about Blackhole interactions is honoured in the README.

## Test plan

- [x] \`./gradlew spotlessApply\` — green
- [x] No code change — markdown only

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -74,26 +74,36 @@ publication-grade) gave:
 Both deep-field benchmarks walk three levels — divide by three for per-level cost: record reflection ≈87 ns/level, the
 `lens` path ≈15 ns/level, native `ofBean` ≈163 ns/level.
 
-A second-run capture of the LMF-tier benchmarks (single-step dispatch, no composition) gave:
+A tighter LMF-tier capture at **5 warmup + 10 measurement × 3 fork** (single-step dispatch, no composition) gave:
 
-| Benchmark                          |                                                             ns/op |                       vs hand-rolled |
-| ---------------------------------- | ----------------------------------------------------------------: | -----------------------------------: |
-| `recordComponentRead_handRolled`   |                                                              ~0.8 |                    record-read floor |
-| `recordComponentRead_lmf`          |                                                               ~15 |                                 ~18× |
-| `recordComponentRead_methodInvoke` | _to be captured at 5 warmup + 10 measurement × 3 fork before 1.0_ | apples-to-apples reflection baseline |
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #28, opened 2026-06-07
feat(codegen): Phase A — FocusProcessor emits <X>Telescope metadata holder (ADR-0006)

eschizoid:
## Summary

Phase A of [ADR-0006](docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md): `FocusProcessor` now emits a sibling `<X>Telescope` metadata holder alongside the existing `<X>Path<R>` navigator. The holder is a top-level `public final` utility class in the user's package, with one `public static final Telescope<X, FieldType>` constant per record component.

This PR is **purely additive**:
- No runtime consumer of the holder yet (Phase B wires the `ClassValue<Optional<HolderRef>>` probe in `Records.fieldLens(...)`).
- `DeepMap` is not touched (Phase C).
- `BeanFocusProcessor` / `LombokFocusProcessor` not touched (follow-up PR).
- No existing tests change. No existing API change. Existing emission unchanged.

## Design (per ADR-0006)

- Holder location: top-level `<X>Telescope` in the user's package, mirroring the `<X>Bridge` precedent.
- Each constant is `Telescope.lens(X::comp, (s, v) -> new X(...))` — same expression the Path emits, intentionally duplicated rather than abstracted.
- Container components surface as raw container lenses (`Telescope<X, List<E>>`) — the Path's container step lifts; the holder does not. Consumers compose with `.then(...)` when they need element-level navigation. Keeps the holder shape uniform.
- Sub-`@Focus` record components surface as `Telescope<X, SubRecord>` (terminal-to-record). No composition with the sub-holder; consumers compose via `.then(...)` themselves.
- Wildcard / type-var generics rejected at compile time with a precise diagnostic pointing the user to drop `@Focus` (ADR-0006 §7).

Shared helpers `writeMetadataHolderClass(...)` and `isEmittableAsTypedConstant(...)` live in `AbstractTelescopeProcessor` for reuse by the bean processors in the follow-up.

## Tests added

Four new tests in `FocusProcessorTest`:

- `generatesTelescopeHolder` — scalar + primitive boxing (`Telescope<Person, Integer>` for `int age`).
- `telescopeHolderForContainerComponent` — `Telescope<Bag, List<String>>` is the raw container lens (no lifted element step).
- `telescopeHolderForSubFocusComponent` — `Telescope<User, demo.Address>` for an `@Focus` sub-record (no composition with `AddressTelescope`).
[...]

PR #32, opened 2026-06-07
feat(codegen): Phase A — BeanFocus + Lombok emit <X>Telescope metadata holder (ADR-0006)

eschizoid:
## Summary

Sibling PR to #28 (record side). Implements the bean half of Phase A of [ADR-0006](docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md): alongside the existing `<X>Path<R>` navigator, `BeanFocusProcessor` and the out-of-tree `LombokFocusProcessor` (via the shared `AbstractTelescopeProcessor#emitBeanNavigator` entry point) now emit a sibling `<X>Telescope` utility class with one `public static final Telescope<X, PropertyType>` constant per bean property.

This PR is **purely additive**:

- No runtime consumer of the holder yet (Phase B wires the `ClassValue<Optional<HolderRef>>` probe in `Beans.lens(...)` / `Records.fieldLens(...)`).
- `DeepMap` is not touched (Phase C).
- No existing tests change. No existing API or emission change.
- Lombok's `processingOver()` deferral already covers this — the holder is emitted in the same round as the Path, transparently.

## Design (per ADR-0006)

- **Holder location.** Top-level `<X>Telescope` in the user's package, mirroring both `<X>Path` and `<X>Bridge` (and the record-side from #28).
- **Lens expression.** Each constant uses `Telescope.lens(X::getProp, <rebuild>)` where `<rebuild>` is the same builder-or-no-arg-ctor expression the Path navigator emits via `beanRebuild`. Write semantics across Path and holder are identical.
- **Container properties** surface as raw container lenses (`Telescope<X, List<E>>`, etc.) — the Path's container step lifts; the holder does not. Consumers compose via `.then(...)` to descend.
- **Sub-bean / sub-record properties** surface as `Telescope<X, SubType>` — terminal to the sub-value, not composed with the sub-holder.
- **Wildcard / type-var generics** rejected at compile time with a precise diagnostic pointing the user to drop the trigger annotation (ADR-0006 §7).

## Implementation notes

- The new `emitBeanMetadataHolder(...)` method lives in `AbstractTelescopeProcessor` and is called at the tail of `emitBeanNavigator(...)`. Both `BeanFocusProcessor` and `LombokFocusProcessor` benefit automatically — neither subclass needed a modification.
- `collectStdImports` moved from `FocusProcessor` (private static) up to `AbstractTelescopeProcessor` (protected static) so record and bean emitters share one rule. `FocusProcessor`'s now-duplicate copy and its now-unused `TypeMirror` import are dropped.

## Tests added

`BeanFocusProcessorTest` — five new tests in a `MetadataHolder` nested class:

- `generatesTelescopeHolderForBean` — scalar + primitive boxing (`Telescope<Person, Integer>` for `int age`), no-arg-ctor + setX rebuild expression.
- `telescopeHolderForBuilderBean` — builder-strategy rebuild.
- `telescopeHolderForBeanContainerProperty` — `Telescope<Bag, List<String>>` is the raw container lens.
- `telescopeHolderForBeanSubBeanProperty` — `Telescope<UserA, demo.AddressB>` for an `@BeanFocus` sub-bean (no composition with `AddressBTelescope`).
- `wildcardBeanGenericsRejected` — wildcard-bound generics trigger the precise diagnostic.

`LombokFocusProcessorTest` — four new tests in a `MetadataHolder` nested class:

- `dataHolder` — `DataUserTelescope` has `public static final Telescope` fields per property (verified by reflection).
- `builderHolder` — same for `@Builder`.
- `valueBuilderHolder` — same for `@Value + @Builder`.
- `holderConstantsAreUsable` — end-to-end roundtrip: get the `email` constant by reflection, call `.update(user, String::toLowerCase)`, assert the email is lower-cased and `id` round-trips untouched.

## Test plan

- [x] `./gradlew :codegen:test :lombok:test :core:test :examples:test :benchmarks:compileJmhJava` — green.
- [x] `./gradlew spotlessCheck --rerun-tasks` — clean.
- [x] IntelliJ `get_file_problems` on all four touched files — zero new warnings (existing weak warnings unchanged).
- [x] Pre-existing `FocusProcessorTest` / `BeanFocusProcessorTest` / `BridgeProcessorTest` / `LombokFocusProcessorTest` tests still pass unchanged.

## Related

- ADR: [`docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md`](docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md)
- Record-side sibling PR: #28


eschizoid:
@copilot please review

PR #33, opened 2026-06-07
feat(core): Phase B — runtime probe for <X>Telescope metadata holder (ADR-0006)

eschizoid:
## Summary

Phase B of [ADR-0006](docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md): the runtime probe that consumes the `<X>Telescope` metadata holders [PR #28](https://github.com/eschizoid/telescope/pull/28) (Phase A) emits.

A new `ClassValue<Optional<HolderRef>>`-cached `MetadataHolderProbe` short-circuits the reflective `Records.fieldLens` / `Beans.lens` dispatch sites in `Telescope`. When a sibling `<X>Telescope` holder is on the classpath, `.field(Accessor)` routes the `SerializedLambda`-recovered method name to the holder's pre-baked `Telescope<X, FieldType>` constant and pulls its `Lens` out directly. When no holder is on the classpath, today's reflective path runs **unchanged**.

Pure runtime change inside `:core/internal/*`:

- **No public API change.** `Telescope` keeps its current signatures; only the private `RecordFieldOptics` / `BeanFieldOptics` enums learn to consult the probe before falling through.
- **No codegen change.** Phase A emits the holder; Phase B reads it.
- **No `Reflective` change.** Phase B's probe sits BEFORE `Reflective.of(cls)`; the fallback path is preserved exactly as today.
- **No behavioural change for non-annotated users.** Probe-absent → fall through silently to `Records.fieldLens` / `Beans.lens`.

## Design (per ADR-0006)

- §1 Holder location — top-level `<X>Telescope` in the user's package, discovered via `Class.forName(cls.getName() + "Telescope", false, cls.getClassLoader())`.
- §5 `Reflective` unchanged — Phase B is a two-tier probe in front of `Reflective.of(cls)`.
- §6 Cache — `ClassValue<Optional<HolderRef>>`, matching the existing `Records.CACHE` / `Beans.GETTERS` pattern; the absent case is memoised too so a non-annotated class doesn't repeat the `forName` miss.
- §9 Lookup miss — holder PRESENT + name MISSING throws `IllegalStateException` with a precise diagnostic ("Component 'name' not found in `User`'s metadata holder. Re-run the @Focus / @BeanFocus processor."). Silent fallback would mask stale codegen / accessor mismatches.

The probe unwraps each holder constant via the existing module-internal `Telescope#optic()` seam (already `public` with `@SuppressWarnings("exports")` for the conversion sub-package). The unwrap casts `Traversal<?, ?>` to `Lens<?, ?>` — load-bearing because the holders are emitted via `Telescope.lens(getter, setter)`, which wraps a `Lens` exactly.

## Tests added

- **`MetadataHolderProbeTest`** (6 tests):
  - `probeFindsHolder` — fixture `ProbedRecord` (top-level, `@Focus`) → `FocusProcessor` emits `ProbedRecordTelescope`; the probe finds it and lists both constants.
  - `probeReturnsEmptyForUnannotatedClass` — fixture `UnannotatedRecord` (top-level, no annotation) → `Optional.empty()`.
  - `probeIsCached` — repeated calls return the same `Optional` instance (ClassValue memoisation); the empty case is cached too.
  - `lensFromHolderReturnsNullWhenAbsent` — probe-absent + lookup → `null` (dispatch site falls through).
  - `lensFromHolderReturnsConstantWhenPresent` — probe-present + valid name → pre-baked `Lens` that round-trips identically to the record accessor.
  - `holderPresentNameMissingThrows` — probe-present + unknown name → `IllegalStateException` with the ADR §9 diagnostic.

- **`HybridDispatchIntegrationTest`** (4 tests):
  - `annotatedDispatchReadsSameValue` — `Telescope.of(User.class).field(User::name).read(user)` (annotated record → holder lens path) yields the same value as the direct accessor.
  - `annotatedDispatchUpdate` — `.update(user, String::toUpperCase)` rebuilds correctly through the holder lens.
  - `annotatedDispatchPrimitiveBoxing` — `int age` component boxes cleanly via the typed `Telescope<User, Integer>` constant.
  - `unannotatedDispatchFallsThroughToReflectivePath` — a record without `@Focus` still works via the existing `Records.fieldLens` path; the probe transparently misses.

## Test plan

- [x] `./gradlew :core:spotlessApply :core:test :core:check` — green.
- [x] `./gradlew build -x :core:javadoc` — all 45 tasks pass (`:core:javadoc` excluded due to a pre-existing reference-not-found issue on `conversion/Mapper.java#L30`, unrelated to this PR).
- [x] No new `Telescope` signature change; the `@Focus`-emitted `<X>Path<R>` navigator API is untouched.
- [x] Behavioural parity for non-annotated records (the reflective `Records.fieldLens` path is exercised by every pre-existing `:core` test that doesn't carry `@Focus`).


eschizoid:
@copilot please review

PR #34, opened 2026-06-07
feat(core): Phase C — deep mapping uses <X>Telescope holder lenses (ADR-0006)

eschizoid:
## Summary

ADR-0006 Phase C — when a class carries a sibling `<X>Telescope` metadata holder, `Reflective#structuralIso(cls)` routes the per-component instance reads through the holder's pre-baked `Lens` constants instead of `Records.read` / `Beans.readProperty`. This is the deep-mapping (`Telescope.map(...)` / `Telescope.mapper(...)`) leg of the hybrid codegen ↔ runtime story.

- New helper `Reflective#resolveHolderReaders(cls, names)` pre-resolves the `name -> Lens` table at `structuralIso` construction time. Returns `null` when the holder is absent **or** doesn't cover every component, so the backward branch falls through to today's reflective `read` path with no per-component branching inside the hot loop.
- The forward branch (`construct`) is untouched — the holder doesn't expose a constructor primitive, so canonical-ctor / `BeanWriter` dispatch remains the only build path.
- Fallback semantics: holder-absent, partial-coverage, and mixed (one side annotated, one side plain) all degrade gracefully. Net effect for non-annotated users: zero behaviour change.

## Test plan

- [x] `DeepMapPhaseCHolderTest` — both sides annotated, neither side annotated, holders-present sanity check, and mixed annotated/plain round trips
- [x] `./gradlew :core:test` passes (existing `DeepMappingTest`, `HybridDispatchIntegrationTest`, `MetadataHolderProbeTest` all green)
- [x] Spotless clean

## ADR / CHANGELOG

- `docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md` — Phase C bullet updated with what actually landed (backward-branch reads, forward unchanged)
- `CHANGELOG.md` — Phase C moved from "planned" to landed under `[Unreleased]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,11 +17,13 @@ story.
   `public static final Telescope<X, FieldType>` constant per record component / bean property. Pure additive codegen
   output; no runtime consumer in Phase A. See
   [ADR-0006](docs/adr/0006-codegen-runtime-1-1-lookup-via-metadata-holder.md).
-- **Phase B (in flight).** `ClassValue<Optional<HolderRef>>` short-circuit in front of `Records.fieldLens(...)` /
-  `Beans.lens(...)`. When the holder is present on the classpath, the SerializedLambda-derived field name routes to the
-  pre-baked constant directly; when absent, the LMF substrate path runs unchanged.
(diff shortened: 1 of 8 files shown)


eschizoid:
@copilot please review

PR #35, opened 2026-06-07
docs(core): qualify Mapper constructor link in PatchEntry javadoc

eschizoid:
## Summary

`./gradlew :core:javadoc` was failing with `reference not found` on `Mapper.java:30`:

```
* {@link Mapper} via {@link #Mapper(Iso, Class, Class, Map)}. The raw {@link Function} field is
                               ^
```

The `{@link #Mapper(Iso, Class, Class, Map)}` reference sits inside the nested `PatchEntry` record, so `#Mapper(...)` resolves against `PatchEntry` (which has no such constructor) instead of the enclosing `Mapper` class. Qualifying the link as `Mapper#Mapper(Iso, Class, Class, Map)` points it at the actual constructor.

## Test plan

- [x] `./gradlew :core:javadoc` passes
- [x] `./gradlew javadoc` (all modules) passes
- [x] Spotless clean

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/conversion/Mapper.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/conversion/Mapper.java
@@ -27,9 +27,9 @@ public final class Mapper<A, B> {
    *
    * <p>Declared {@code public} solely so the cross-package {@link
    * io.github.eschizoid.telescope.mapping.DeepMap} engine can construct entries when building a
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

eschizoid:
Folded into #36 — same one-line fix to Mapper.java javadoc, no longer needed standalone.

PR #36, opened 2026-06-07
chore(core): drop instanceof short-circuit on asList/asSet/asMap/asOptional

eschizoid:
## Summary

The four typed-container promotion methods (`asList` / `asSet` / `asMap` / `asOptional` on `Telescope`) carried three `@SuppressWarnings` entries — `"unchecked"`, `"exports"`, `"CastConflictsWithInstanceof"` — because of an `instanceof` short-circuit that re-cast through a wildcard projection:

```java
@SuppressWarnings({ "unchecked", "exports", "CastConflictsWithInstanceof" })
public static <S, X> ListPath<S, X> asList(final Telescope<S, List<X>> path) {
    if (path instanceof ListPath<?, ?> lp) return (ListPath<S, X>) lp;
    return new ListPath<>(path.optic, path.fieldOptics, path.chain);
}
```

In practice the short-circuit almost never fires: `.list(getter)` returns `ListPath` directly, so the only callers of `asList(...)` are promoting a generic `Telescope<S, List<X>>` built from another path — the case that needs the allocation regardless. Dropping the short-circuit shrinks each method to a single `@SuppressWarnings("exports")`. The remaining suppression is genuine (`optic` / `fieldOptics` / `chain` are internal-package types crossing a public constructor boundary).

## Changes

- `Telescope.java` — four `as*` methods shrunk to a single `@SuppressWarnings("exports")` each; instance-cast removed
- `TypedContainerPathTest.java` — `asListIdempotent` (which asserted `assertSame`) replaced by `asListBehaviorPreserving` which round-trips through the rewrapped path. Same semantics; no instance-identity assumption
- `Mapper.java` — folds in the javadoc fix from PR #35 (`{@link #Mapper(...)}` → `{@link Mapper#Mapper(...)}` inside the nested `PatchEntry` record) so this branch passes `./gradlew :core:javadoc` standalone. If #35 lands first this becomes a no-op; if this lands first #35 can close as redundant

## Test plan

- [x] `./gradlew :core:test` — full suite green (including `TypedContainerPathTest`)
- [x] `./gradlew :core:javadoc` — green
- [x] `./gradlew spotlessApply --rerun-tasks` — clean

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -161,30 +161,26 @@ public sealed class Telescope<
    * final Telescope<Company, Department> elements = Telescope.asList(built).each();
    * }</pre>
    */
-  @SuppressWarnings({ "unchecked", "exports" })
+  @SuppressWarnings("exports")
   public static <S, X> ListPath<S, X> asList(final Telescope<S, List<X>> path) {
-    if (path instanceof ListPath<?, ?> lp) return (ListPath<S, X>) lp;
     return new ListPath<>(path.optic, path.fieldOptics, path.chain);
   }
 
   /** Pre-built-fragment companion to {@link #asList} for {@code Set&lt;X&gt;} paths. */
-  @SuppressWarnings({ "unchecked", "exports" })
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #37, opened 2026-06-07
feat(codegen)!: Phase D — <X>Telescope.construct holder bypasses reflective construct (ADR-0006)

eschizoid:
## Summary

Closes the forward branch of `Reflective#structuralIso` for annotated types so both directions of deep-mapping decomposition are reflection-free.

- **Codegen.** The sibling `<X>Telescope` holder gains `public static <X> construct(Function<String, Object> values)`. For records, the body calls the canonical constructor directly with cast-pulled values. For beans, the body mirrors whatever write strategy `emitBeanNavigator` picked (builder chain or no-arg ctor + setters). Lombok inherits via the existing `emitBeanNavigator` path; the round-deferred `processingOver()` emission already guards against AST visibility.
- **Runtime.** `MetadataHolderProbe.HolderRef` adds an optional `Function<Function<String, Object>, Object> constructor` field bound once via `LambdaMetafactory` at probe time (no `privateLookupIn` needed — the holder is plain public Java). `Reflective#structuralIso(cls)`'s forward branch (`Map → instance`) routes through it when present; legacy holders without `construct` surface `null` and the engine falls back to today's reflective `Records.construct` / `Beans.BeanWriter` path. Pre-resolution-or-nothing posture matches `resolveHolderReaders` — one branch outside the `Iso`'s hot loop, not inside.
- **Net effect.** For annotated type pairs, both forward (`construct`) and backward (`read`) branches of `structuralIso` are reflection-free. Only `SerializedLambda` decode on user accessor method references remains — matching ADR-0006's closing claim.
- **Drive-by.** `Mapper.PatchEntry` javadoc `{@link #Mapper(...)}` → `{@link Mapper#Mapper(...)}` so `:core:javadoc` compiles clean. The nested-record context needed the outer-class qualifier.

## Test plan

- [x] `:core:test` — new `DeepMapPhaseDConstructTest` covers the holder-driven forward branch round-trip (annotated/plain/mixed cases), exercises the bound constructor directly, and verifies the graceful `null`-constructor degradation via a hand-written legacy-shape holder fixture (`PhaseDLegacyHolderTarget` + `PhaseDLegacyHolderTargetTelescope`).
- [x] `:codegen:test` — new `MetadataHolderConstruct` nested classes in `FocusProcessorTest` and `BeanFocusProcessorTest` verify the emitted `construct(Function)` shape: signature, `@SuppressWarnings("unchecked")`, primitive boxing, container-generic casts, builder vs setter rebuild bodies, and the no-construct-on-rejected-holder guard.
- [x] `:lombok:test` — new Phase D tests in `LombokFocusProcessorTest.MetadataHolder` call the emitted `DataUserTelescope.construct` and `BuilderUserTelescope.construct` end-to-end via reflection, confirming Lombok inherits Phase D through `emitBeanNavigator` without round-ordering issues.
- [x] `:core:javadoc` clean.
- [x] `spotlessCheck --rerun-tasks` clean (stale-cache guard from MEMORY).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,15 @@ story.
   probes for a sibling `<X>Telescope` holder per side. When the holder covers every component, every per-component read
   during instance decomposition routes through the pre-baked `Lens` constants instead of `Records.read` /
   `Beans.readProperty`. Holder-absent types and partial holders fall through to the reflective path unchanged.
+- **Phase D.** `<X>Telescope` holders gain a `public static <X> construct(Function<String, Object> values)` method that
+  mirrors the same write strategy the `<X>Path<R>` lens setters chose (canonical constructor for records; builder chain
+  or no-arg ctor + setters for beans). `MetadataHolderProbe.HolderRef` adds an optional
+  `Function<Function<String, Object>, Object> constructor` field bound once via `LambdaMetafactory` at probe time.
+  `Reflective#structuralIso(cls)`'s forward branch (`Map → instance`) routes through it when present, bypassing the
+  reflective `Records.construct` / `Beans.BeanWriter` path. Legacy holders without a Phase D `construct` method degrade
+  gracefully (the `constructor` field is `null` and the reflective fallback runs). Combined with Phase C, both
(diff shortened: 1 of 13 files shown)


eschizoid:
@copilot please review

PR #38, opened 2026-06-07
chore(examples): replace Main.java orchestrator with per-demo Gradle tasks

eschizoid:
## Summary

`Main.java` existed solely to call the ten demo classes' static `run()` methods in sequence. That orchestration is what Gradle is for, not Java.

Each demo class now carries its own `static void main()` (Java 25 unnamed-arg form, matching the project's existing style) that delegates to `run()`. `examples/build.gradle.kts` registers one `JavaExec` task per demo plus an aggregator `runAllDemos` that depends on all of them:

- `./gradlew :examples:runAllDemos` — full smoke test
- `./gradlew :examples:runCodegenDemo` (and friends) — targeted single-demo run
- A failing demo trips its own task name (`runDeepMappingDemo FAILED`) instead of getting buried in a single-JVM stack trace

Trade-off: ten JVM startups instead of one (~5s overhead total). For a smoke-test workload that runs in CI, that's acceptable; for local iteration the targeted-task win is the bigger value.

The `application` plugin and `mainClass.set(...)` wiring were the only consumers of `Main.java` and have been removed.

## Test plan

- [x] `./gradlew :examples:compileJava` green
- [x] `./gradlew :examples:runAllDemos` exercises all ten demos end-to-end (all printed "Before / After" traces still visible per-task)
- [x] `./gradlew :examples:runCodegenDemo` (sampled) runs the single demo in isolation
- [x] `./gradlew :examples:spotlessApply :examples:javadoc` green


eschizoid:
@copilot please review

PR #39, opened 2026-06-07
chore(benchmarks): JMH benchmarks for ADR-0006 Phases B + C holder dispatch

eschizoid:
## Summary

JMH benchmarks that quantify the per-call cost of ADR-0006's holder-routed dispatch path against the LMF substrate it short-circuits. Benchmarks-only PR — no code changes outside `:benchmarks`.

- **Phase B — per-field dispatch.** `field_holder` vs `field_lmf` measures whether `MetadataHolderProbe.lensFromHolder` shortens the `Telescope.of(X).field(X::name).update(...)` hot path. `field_holder_constant` (direct `<X>Telescope.NAME.update(...)`, probe bypassed) is the codegen-direct ceiling the holder path can approach but not exceed.
- **Phase C — deep-mapping backward branch.** `mapForward_*` and `mapBackward_*` measure whether the `Reflective#structuralIso(cls)` backward branch's holder-lens reads land measurable savings across a 3-level record-tree conversion.
- Fixtures are paired: `_holder` carry `@Focus` (sibling `<X>Telescope` holder emitted); `_lmf` are the same shape without the annotation. Only the dispatch path differs.

## Results

Captured at 5 warmup + 10 measurement × 3 forks on JDK 25 / Apple Silicon:

| Benchmark               |  ns/op | ±error |                                                vs LMF |
| ----------------------- | -----: | -----: | ----------------------------------------------------: |
| `field_holder`          |   25.6 |   ±0.3 |                              **3.54x vs `field_lmf`** |
| `field_holder_constant` |   26.0 |   ±1.1 |        direct holder — within error of `field_holder` |
| `field_lmf`             |   90.7 |   ±4.5 |                                LMF substrate baseline |
| `mapForward_holder`     |  808.3 |  ±38.8 |                            ~1.04x vs `mapForward_lmf` |
| `mapForward_lmf`        |  837.1 |  ±43.7 |                                LMF substrate baseline |
| `mapBackward_holder`    | 1163.0 | ±334.5 | _noisy: concurrent JMH on same host, wide error band_ |
| `mapBackward_lmf`       | 1374.3 | ±522.2 | _noisy: concurrent JMH on same host, wide error band_ |

### Honest read

- **Phase B is a clear win.** Holder = 25.6 ns vs LMF = 90.7 ns is a real 3.54x speedup, error bars don't overlap. And `field_holder` is within the same error band as `field_holder_constant` — so the `ClassValue`-backed probe adds essentially zero overhead and the holder-routed path reaches the codegen-direct ceiling.
- **Phase C forward is comparable.** ~3-4% faster on the means, but error bars overlap. The construction side dominates the per-call cost; Phase D (constructor holder) is where the forward branch becomes measurable.
- **Phase C backward is comparable but noisy.** A concurrent sibling JMH workload was running on the same machine during that portion of the measurement, so the wide error bars don't pin a meaningful ratio. The trend matches Phase C forward (~0.85x), so there's no surprise regression.

The architectural story holds even where the per-call delta is small: the holder gives the runtime a zero-config-codegen handshake, and Phase D will use the same probe infrastructure for the forward path. The Phase B win is the headline; Phase C is parity with a forward-looking foundation.

## Files

- `benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/HolderDispatchBenchmark.java` — the new bench.
- `BenchHolderRec.java` / `BenchPlainRec.java` — paired flat record fixtures for the per-field rows.
- `BenchHolder{Src,Tgt,DeptSrc,DeptTgt,AddressSrc,AddressTgt}.java` and the matching `BenchPlain*` set — 3-level record-tree fixtures for the deep-mapping rows. `@Focus` on the holder side, plain records on the LMF side; structurally identical at every level.
- `benchmarks/build.gradle.kts` — property-based knobs (`-Pjmh.warmup`, `-Pjmh.iterations`, `-Pjmh.fork`, `-Pjmh.includes=…`) so dev cycles and full publication-grade runs share the same config without editing the file.
- `benchmarks/README.md` — new "Hybrid dispatch (ADR-0006 Phases B + C)" section with descriptive table, results table, and the honest-read subsection.

## Out of scope (Phase D)

Phase D (constructor holder for the forward branch of `structuralIso`) is parallel agent work and will land separately. The `mapForward_*` rows already exercise the source-side backward Iso branch under today's `@Focus` holders; when Phase D lands its forward-branch numbers can be added as a follow-up row in the same table.

## Test plan

- [x] `./gradlew :benchmarks:compileJmhJava` clean — generated `BenchHolderRecTelescope` + sibling holders confirmed in `benchmarks/build/generated/sources/annotationProcessor/java/jmh/...`
- [x] Full 5+10×3 fork run completed, numbers pasted into `benchmarks/README.md`

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -34,6 +34,23 @@ POJO benchmarks walk an identical mutable-bean mirror. The conversion benchmarks
 | `fromBeanForwardRead`      | `Telescope.fromBean(...).viaFields().read(...)` POJO→record bridge.                                  |
 | `bridgeForwardRead`        | Same POJO→POJO conversion via the generated `@Bridge` constant — reflection-free codegen.            |
 
+### HolderDispatchBenchmark — sibling metadata holder routing (ADR-0006 Phases B + C)
+
+These benchmarks isolate the holder-routed dispatch path that ADR-0006 layers on top of the LMF substrate. Each axis has
+paired `_lmf` (no `@Focus`, probe misses, LMF path runs) and `_holder` (sibling `<X>Telescope` holder on the classpath,
+probe hits, holder constant returned) rows. Fixtures are structurally identical between the two flavours — same field
+names, same field types, same 3-level tree — so the only difference at dispatch time is whether the holder is present.
+
+| Benchmark               | What it does                                                                                                                                                              |
+| ----------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `field_lmf`             | `Telescope.of(BenchPlainRec.class).field(BenchPlainRec::name).update(...)` against an unannotated record — Phase B fallback to the LMF-backed `Records.fieldLens(name)`.  |
+| `field_holder`          | Same dispatch against a `@Focus`-annotated record — `MetadataHolderProbe` hits and returns the pre-baked `BenchHolderRecTelescope.name` constant.                         |
+| `field_holder_constant` | The generated constant invoked directly (`BenchHolderRecTelescope.name.update(...)`) — skips the probe entirely; the pure codegen-direct ceiling.                         |
+| `mapForward_lmf`        | `Telescope.mapper(BenchPlainSrc.class, BenchPlainTgt.class).forward(...)` — Phase C fallback. Source-side backward Iso branch reads each component via `Reflective.read`. |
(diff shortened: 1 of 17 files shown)


eschizoid:
@copilot please review

PR #42, opened 2026-06-07
docs(readme): add honest telescope-vs-MapStruct positioning section

eschizoid:
## Summary

The original "Not a MapStruct competitor" framing from v0.1 was accurate at the time but undersells what ADR-0005 (LMF substrate) + ADR-0006 Phases A-D (codegen-runtime hybrid) actually delivered. With Phase D landed and the [holder-dispatch numbers](benchmarks/README.md#holderdispatchbenchmark--sibling-metadata-holder-routing-adr-0006-phases-b--c--d) showing measurable wins in both deep-mapping directions, the project is competitive for the use cases where it overlaps with MapStruct — but the **centre of gravity** is still different, and the README should say so honestly.

## Changes

- **Softened the existing "Not a MapStruct competitor" bullet** in the "What it is _not_" section from "competitor" → "replacement", with a forward-link to the new positioning section.
- **Added a new "How it compares to MapStruct" section** with:
  - A side-by-side capability table (bidirectional, deep navigation, effectful update, codegen, runtime fallback, sealed types, expression-based mappings, lifecycle hooks, ecosystem integration, maturity, perf)
  - "When MapStruct is the right pick" subsection
  - "When telescope is the right pick" subsection
  - **"Performance honesty" paragraph** that's explicit about what the 3.23× ratio does and doesn't compare — it's telescope's `@Focus` codegen path vs its own reflective fallback, NOT vs MapStruct. We haven't run apples-to-apples vs MapStruct and the README says so.

## Scope

- Docs only, no source changes
- No version bump
- No claim that telescope is faster than MapStruct (we haven't measured it)

## Test plan

- [x] Markdown anchor (`#how-it-compares-to-mapstruct`) verified by inspection
- [x] No emoji, no AI patterns, no "stands as a testament" prose

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,14 +119,72 @@ That's the library. No `Iso`, `Lens`, `Prism`, `Affine`, `Traversal`, `Getter`, 
 
 ## What it is _not_
 
-- Not a MapStruct competitor. MapStruct owns compile-time bean mapping. For flat `Entity → Dto` work, write a static
-  method or use MapStruct.
+- Not a MapStruct replacement. MapStruct is a purpose-built bean-mapping framework with deep ecosystem integration
+  (Spring, Quarkus, MicroProfile) and ten years of edge-case fixes. Telescope is an optics DSL that includes
+  bidirectional mapping as one capability — see [How it compares to MapStruct](#how-it-compares-to-mapstruct) below for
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #43, opened 2026-06-08
feat(codegen): emit constants() on <X>Telescope + drop pre-1.0 legacy fallbacks

eschizoid:
## Summary

Two related changes that close the remaining cold-path reflection gap and tighten the codegen ↔ runtime contract:

**1. `constants()` method on `<X>Telescope`** — `FocusProcessor` + `AbstractTelescopeProcessor` (BeanFocus + Lombok) emit a `public static Map<String, Telescope<?, ?>> constants()` returning the pre-baked name → lens map. The runtime probe consumes it directly: one `getDeclaredMethod` + one `Method.invoke` regardless of holder size, vs the prior `getDeclaredFields()` scan + per-field `field.get(null)` loop. Cold-path probe goes from `O(N)` reflective ops to `O(1)`.

**2. Drop pre-1.0 legacy fallback paths.** Pre-1.0 telescope has no downstream users locked into older codegen output; the legacy `getDeclaredFields()` scan (for holders without `constants()`) and the null-constructor branch (for holders without `construct(Function)`) were carrying their weight only for hypothetical users. Removed both. A holder that's missing either required member now trips a precise `IllegalStateException` pointing at "re-run the @Focus / @BeanFocus processor" rather than silently falling back. `HolderRef.constructor` is now guaranteed non-null when a holder is present.

### Codegen

- `FocusProcessor#emitRecordConstantsMap` — new emitter, called after `emitRecordConstruct`
- `AbstractTelescopeProcessor#emitBeanConstantsMap` — same, called after `emitBeanConstruct`
- Both emit `Map.of("only", only)` for single-entry holders; `Map.ofEntries(Map.entry(...), ...)` for multi-entry

### Runtime

- `MetadataHolderProbe.readConstants` calls `constants()` directly; throws when missing
- `MetadataHolderProbe.bindConstructor` throws when `construct(Function)` is missing (was returning null)
- `MetadataHolderProbe.readConstantsLegacy` deleted
- `Reflective#structuralIso` javadoc updated — no more "older Phase A holder" branch
- `LinkedHashMap` import removed (last consumer was the legacy field-scan loop)

### Tests

- `FocusProcessorTest.MetadataHolderConstantsMap` (new) — verifies emission shape for multi-field and single-field records
- `BeanFocusProcessorTest.MetadataHolderConstantsMap` (new) — same for beans
- `MetadataHolderProbeTest.legacyHolderWithoutConstantsMethodFallsBackToFieldScan` — **deleted**
- `DeepMapPhaseDConstructTest.legacyHolderHasNullConstructor` — **deleted**
- `PhaseELegacyHolderTarget(+Telescope)` fixtures — **never committed**, scrubbed before the PR
- `PhaseDLegacyHolderTarget(+Telescope)` fixtures — **deleted**

### Docs

- ADR-0006 — Phase E bullet updated to reflect the no-fallback stance
- `CHANGELOG.md` — Phase E entry updated likewise
- Scrubbed ADR-XXXX and Phase-X labels from javadoc / comments / `@DisplayName` strings per project convention; the ADR file, CHANGELOG, and PR bodies remain the right venues for those cross-references

## Test plan

- [x] `./gradlew :core:test :codegen:test :lombok:test` all green
- [x] `./gradlew :core:javadoc` green
- [x] Behavior: re-running `MetadataHolderProbeTest` on a hand-authored holder without `constants()` now throws `IllegalStateException` (manually verified by reverting `FocusProcessor`)


eschizoid:
@copilot please review

PR #44, opened 2026-06-08
chore(core,codegen): clear IntelliJ warnings — getFirst, method refs, redundant suppression

eschizoid:
## Summary

Five small cleanups driven by IntelliJ inspection results post-Phase-E. All are net-line-removing or net-zero; no behaviour change.

1. **`Reflective#resolveHolderConstructor`** — collapse the `if (maybeHolder.isEmpty()) return null;` guard + `.get().constructor()` into a single `Optional.map(HolderRef::constructor).orElse(null)` expression
2. **`FocusProcessor#emitRecordConstantsMap`** — drop unused `recordName` parameter (the new emission uses `components.getFirst().getSimpleName()` directly); replace `components.get(0)` with `components.getFirst()` (Java 21+ idiom)
3. **`AbstractTelescopeProcessor#emitBeanConstantsMap`** — replace `props.get(0)` with `props.getFirst()`
4. **`Beans#getter`** — `source -> reader.apply(source)` becomes `reader::apply` (method reference)
5. **`Iso#liftOptional`** — drop redundant `OptionalUsedAsFieldOrParameterType` suppression; the `OptionalAssignedToNull` suppression is the only one that actually fires (verified by post-edit IntelliJ run)

## Intentional flags left alone

- **`{@link Class#forName}` / `Method#invoke` "Symbol inaccessible" errors** — IntelliJ JPMS confusion on `java.base` exports; `:core:javadoc` compiles cleanly
- **Unused type-inference parameters** on `Telescope.of(Class<S>)`, `Telescope.ofBean(Class<P>)`, `From.to(Class<B>)`, `Telescope.fieldByName(String, Class<B>)` — drive `<S>` / `<P>` / `<B>` inference at call sites; documented as such
- **Phantom type parameters** on `Mapping<A, B>` / `Kind<F, A>` / `EitherK<L>` / `ValidatedK<E>` — load-bearing for the sealed-permit and HKT-emulation patterns
- **Unused public-API surface** on `Validated#isInvalid` / `#fold` / `#combine` and `Mapper#asTelescope` — exposed for downstream consumers, may not be called inside the library
- **Duplicated `Iso.liftList` / `liftSet` fragments** — different `Collector` instances per container shape; refactoring would obscure intent
- **`Mapper` constructor + `DeepMap#resolve` "Class Iso<A, B> is not exported"** — both already carry `@SuppressWarnings("exports")`; IntelliJ's separate module-export inspection doesn't honour the javac suppression

## Test plan

- [x] `./gradlew :core:test :codegen:test :lombok:test` all green
- [x] `./gradlew :core:javadoc` green
- [x] `mcp__idea__get_file_problems` re-run on every changed file shows the targeted warning gone

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,10 +35,10 @@ story.
   `SerializedLambda` decode on user accessor method references.
 - **Phase E.** `<X>Telescope` holders gain a `public static Map<String, Telescope<?, ?>> constants()` method returning
   the name → lens map directly. `MetadataHolderProbe.probe(...)` calls this method as the only path; a holder that's
-  missing the method (out-of-date codegen on the classpath) trips a precise `IllegalStateException` rather than
-  silently falling back. Same posture for the `construct(Function)` method emitted in Phase D — holder presence is a
-  full contract, not a probe of independent optional methods. Cuts the cold-path probe from `~3 + N` reflective ops
-  (N = holder field count) to `3` ops regardless of N. The probe is already `ClassValue`-cached, so this is a
+  missing the method (out-of-date codegen on the classpath) trips a precise `IllegalStateException` rather than silently
+  falling back. Same posture for the `construct(Function)` method emitted in Phase D — holder presence is a full
+  contract, not a probe of independent optional methods. Cuts the cold-path probe from `~3 + N` reflective ops (N =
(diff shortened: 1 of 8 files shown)


eschizoid:
@copilot please review

PR #45, opened 2026-06-08
fix(ci): compute next version explicitly via -Prelease.forceVersion

eschizoid:
## Summary

axion-release 1.18.7's `-Prelease.incrementer=<patch|minor|major>` flag is being silently ignored on telescope's main branch — the `release` task resolves the current version to the latest tag (`v0.3.0`) instead of bumping per the incrementer, then tries to recreate that tag and crashes:

```
> Task :release FAILED
Creating tag: v0.3.0
> RefAlreadyExistsException: tag '...tag v0.3.0...' already exists
```

Two consecutive Release workflow runs (runs `27111925151` and `27112719174`) hit the same failure. The first three releases (v0.1.0, v0.2.0, v0.3.0) used the same workflow and command and succeeded — so this is plausibly an axion-release edge case on `0.x` versions or on a specific git state past the last tag. The sibling kpipe repo with identical config also still works for it. Diagnosis is incomplete; what's certain is that telescope's `release -Prelease.incrementer=minor` no longer increments.

## What changes

- **New `Compute next version` step** parses `git tag --sort=-version:refname` to find the latest `vMAJ.MIN.PATCH`, then applies the dropdown selection (`patch` / `minor` / `major`) to produce the next version explicitly.
- **`Update Version and Create Tag` step** now passes that version via `-Prelease.forceVersion=$NEXT` — axion-release honours this flag unconditionally and tags at the literal version.
- `RAW_VERSION` is now sourced from the computed value rather than `./gradlew currentVersion` (which was returning `0.3.0-SNAPSHOT`, the source of the original mistag).

## Trade-off

The workflow no longer depends on axion-release's auto-increment logic; the dropdown selection drives the version computation deterministically. Slight loss: axion-specific incrementer customisation (e.g. branch-specific incrementers configured in `scmVersion {}`) is bypassed. For this repo's main-only release flow that's not a concern.

## Test plan

- [x] Workflow YAML is syntactically valid
- [x] Shell snippet matches `vMAJ.MIN.PATCH` (regex-anchored) — won't pick up `v0.1.0-rc1` or similar
- [ ] **Re-trigger Release workflow after merge** with `releaseType=minor`; expect a successful tag at `v0.4.0`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -45,15 +45,33 @@ jobs:
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           passphrase: ${{ secrets.GPG_PASSPHRASE }}
 
+      - name: Compute next version
+        id: nextver
+        run: |
+          # axion-release's auto-incrementer (`-Prelease.incrementer=<patch|minor|major>`) was
+          # silently skipping the bump on this repo — `release` kept resolving the current version
+          # to the latest tag (e.g. `v0.3.0`) and tried to recreate it. Compute the next version
+          # explicitly from `latest-tag + dropdown selection` and pass it via
+          # `-Prelease.forceVersion`, which axion honours unconditionally.
+          LATEST=$(git tag --sort=-version:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | head -1 | sed 's/^v//')
+          IFS='.' read -r MAJ MIN PATCH <<< "$LATEST"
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

eschizoid:
Testing the :examples-as-cause theory directly on main per the diagnostic narrowing — temporarily removing the :examples subproject. If a release succeeds, that confirms the cause; we can decide how to bring :examples back without breaking releases.

PR #48, opened 2026-06-09
feat(codegen,lombok): @Bridge deep recursion + Lombok nested + same-module Path emission

eschizoid:
## Summary

Three substantive fixes addressed together — all align with the lattice-first mantra (changes in dispatch/timing/naming; the lattice composition rules and emitted Iso semantics are untouched).

- **`@Bridge` codegen achieves full feature parity with `DeepMap.autoIso`** (PLAN item 7). Queue-based emission with cycle-safe `seen`-set. Per-field `FieldPlan` dispatches across 8 shapes mirroring the runtime: IDENTITY, RECURSE, LIST, SET, MAP_VALUES, OPTIONAL, OPTIONAL_TO_NULLABLE, NULLABLE_TO_OPTIONAL. Auto-generated sub-bridges use `<Source>To<Target>Bridge` naming; user-declared `<Source>Bridge` wins by simple-name convention. Self-referencing types compile-recurse exactly once via the seen-set; runtime recursion terminates on `Optional.empty()`. Map key validation: source/target keys must `isSameType` exactly. Static `forward(Source)` / `backward(Target)` methods on each bridge so child bridges call parents directly — the `Telescope` BRIDGE constant wires both via `from/to/using` method refs (no `Telescope.set(dummy, value)` gymnastics).
- **Lombok-emitted `<X>Path` visible to same-module main code** (PLAN item 10). `LombokFocusProcessor` switched from `processingOver()`-only emission to retry-each-round: emit when `beanProperties()` returns non-empty (Lombok has finished its lazy AST patches that round); defer to the next round otherwise; last-resort fallback on `processingOver()` so anything that never became readable still surfaces its real `no readable properties` error. The Path symbol is now visible in time for same-module main-source binding.
- **Lombok nested static classes supported.** `LombokFocusProcessor` no longer rejects nested classes; `AbstractTelescopeProcessor` flattens enclosing-class hierarchy (`Outer.Inner` → `OuterInnerPath` at package level) and uses dotted form for in-source type references. Removes the v0.4.1 demo workaround of splitting `PartnerCustomer`/`PartnerAddress`/`PartnerLineItem` into top-level siblings.

## Test plan

All net-new tests pass on top of every pre-existing test.

- [x] `:codegen:test` — 7 new tests in `BridgeProcessorTest`:
  - `nestedRecordPairAutoBridge` — parent `@Bridge` auto-emits a `CustomerToCustomerDtoBridge` for nested mismatch
  - `userDeclaredSubBridgeWins` — sub-pair with user `@Bridge` keeps simple-name, no duplicate emission
  - `listContainerAutoLifts` — `List<X>↔List<Y>` lifts element-wise via stream
  - `setContainerAutoLifts` — `Set<X>↔Set<Y>` lifts via `LinkedHashSet`
  - `optionalContainerAutoLifts` — `Optional<X>↔Optional<Y>` via `.map(::forward)`
  - `mapValuesAutoLift` — `Map<K, V>↔Map<K, V'>` lifts values, preserves keys (`LinkedHashMap`)
  - `optionalToNullableCrossParadigm` — Optional↔nullable bridge (mirror of `Iso.liftOptionalToNullable`)
  - `cyclicTypeEmitsOnce` — self-referencing `Node { Optional<Node> child }` emits one bridge; recursion at runtime
  - All 6 pre-existing tests still pass (existing assertions updated for the new `Bridge::forward` / `Bridge::backward` method-ref shape).
- [x] `:lombok:test` — 2 new tests in `LombokFocusProcessorTest`:
  - `nestedStaticDataClassEmitsFlattenedPath` — `OuterWithNested.Inner` yields `OuterWithNestedInnerPath` + holder at package level
  - `sameRoundConsumerCanReferenceEmittedPath` — `SameRoundConsumer` in `src/test/java` references `DataUserPath.start().email().update(...)` directly; runs end-to-end
  - All pre-existing tests still pass
- [x] `:core:test` — green, no changes to core
- [x] Full `./gradlew check` — green across every module

## Architecture notes

- `FieldPlan.IDENTITY_ELEMENT_SENTINEL` (`__IDENTITY__`): for containers with same-typed elements (e.g. `List<Foo>` ↔ `List<Foo>` reached via a parent type difference), the per-element lift becomes `e -> e` instead of calling a sub-bridge. Saves a layer of indirection.
- `emitMethodBody` handles the two output shapes from `buildExpr`: expression form (`new Foo(...)` — prefix with `return ... ;`) vs block form (`{ var out = ...; ... return out; }` — emit statements directly).
- Cycle handling: at compile time, the `seen` set blocks re-queuing a `TypePair`; at runtime, the BRIDGE constants reference each other via method refs (lambdas in `Telescope.from(...).to(...).using(...)`) so static-init order is irrelevant.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
@@ -285,6 +285,40 @@ public abstract class AbstractTelescopeProcessor extends AbstractProcessor {
   }
 
   /**
+   * Flatten the enclosing-class hierarchy of {@code element} into a single concatenated base name.
+   * For a top-level class {@code Foo} returns {@code "Foo"}; for a nested class {@code Outer.Inner}
+   * returns {@code "OuterInner"}; for double-nested {@code A.B.C} returns {@code "ABC"}. Used to
+   * derive collision-free file-level names for emitted Path / Step / Telescope classes when the
+   * source type is nested — telescope-codegen always emits at the package level, never as a nested
+   * sibling, so we encode the outer hierarchy into the name to keep things unique.
+   */
+  protected static String flattenedNameOf(final Element element) {
+    final var sb = new StringBuilder(element.getSimpleName().toString());
+    var enclosing = element.getEnclosingElement();
+    while (enclosing != null && enclosing.getKind() != ElementKind.PACKAGE) {
+      sb.insert(0, enclosing.getSimpleName().toString());
+      enclosing = enclosing.getEnclosingElement();
+    }
+    return sb.toString();
+  }
+
+  /**
+   * Build a Java type reference for {@code element} as it would appear in source code sitting in
+   * the same package — {@code "Foo"} for top-level, {@code "Outer.Inner"} for nested. Use this
+   * anywhere the emitted Path / Step source needs to spell the source type's class identifier (the
+   * {@code Telescope<R, X>} parameter, method references like {@code X::getY}, etc).
+   */
+  protected static String packageRelativeTypeRefOf(final Element element) {
+    final var sb = new StringBuilder(element.getSimpleName().toString());
+    var enclosing = element.getEnclosingElement();
+    while (enclosing != null && enclosing.getKind() != ElementKind.PACKAGE) {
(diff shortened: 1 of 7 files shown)


eschizoid:
@copilot please review

PR #49, opened 2026-06-09
fix: 5 dispatch-tier bugs surfaced by the 13-slice Tier A/B/C bug hunt

eschizoid:
## Summary

The 13-slice parallel demo-coverage audit (the workflow PR #48's branch ran in worktrees) surfaced exactly the dispatch-tier bugs PLAN.md predicted. **Every fix is in the dispatch/timing/cache-key tier — none touch the optic composition lattice.** All 5 fixes have regression tests; every pre-existing test still passes.

### Bug → fix matrix

| # | Surfaced by | Site | Fix |
|---|---|---|---|
| 1 | Tier A2 (sealed types + paradigm hop) | `Telescope#field/each/eachValue/whenPresent/list/setField/mapField/optional` (8 sites) | New `lensForAccessor(getter)` helper re-picks `FieldOptics` per call based on the accessor's declaring class via `SerializedLambda`. Closes PLAN item 4. |
| 2 | Tier A3 (JPA cycles) | `DeepMap#lazyCacheIso` | `ThreadLocal<IdentityHashMap>` per-traversal seen-set (separate forward/backward). Re-entry on the same instance returns `null` instead of overflowing. |
| 3 | Tier A1 (LAZY/HibernateProxy) | `Beans#readProperty` | New `Beans.persistentClassOf(Object)` — reflection-based Hibernate-proxy unwrap, no hard dependency. Routes cache lookup to persistent class. |
| 4 | Tier B9 (Set navigation) | `AbstractTelescopeProcessor#writeInstanceClass` | One-line: add `import java.util.Set;` next to the other `java.util.*` imports in the eager preamble. |
| 5 | Tier C11 (fieldByName) | `Records#noField` | Append `— known fields: [...]` from `componentNames(cls)` to the error message. |

### Bug #1 detail — paradigm-hop dispatch

The chain `.field(...).then(mapperBridge).as(BeanSubtype.class).field(BeanSubtype::getX)` threw `IllegalArgumentException: Not a record: BeanSubtype`. Root cause: `Telescope.field()` routed through `this.fieldOptics`, locked-in at `Telescope.of(...)` entry. After the paradigm hop the focus type switched to a bean but `fieldOptics` stayed `RecordFieldOptics`.

Fix: `lensForAccessor(getter)` resolves the accessor's declaring class via `LambdaIntrospection.implClassOf` and picks `RecordFieldOptics` if it's a record, otherwise `BeanFieldOptics`. The stored `fieldOptics` is now strictly a fallback for `fieldByName(String)` which has no accessor.

Regression test: `TelescopeTest.ParadigmHopDispatch.fieldAfterParadigmHop` — `Telescope.of(BoxRoot.class).field(...).then(recToBeanMapper).field(BoxBean::getEmail).update(...)` reads and updates the bean leaf correctly through the mapper hop.

### Bug #2 detail — DeepMap value-cycle SOE

`mapper.backward(hibernateHydratedTree)` recursed indefinitely when the tree had a bidirectional self-association (`bob.manager == alice && alice.reports.contains(bob)`). DeepMap's cycle detection was type-level only.

Fix: per-thread `IdentityHashMap` seen-sets in `lazyCacheIso`. Re-entry severs the cycle to `null`. Cleans up when the outermost call unwinds so subsequent independent traversals start fresh.

Acknowledged trade-off: the second occurrence becomes `null` (structure is lost on revisit). Documented in the cycle guard's javadoc. The graph is finite by construction; no SOE.

Regression test: `DeepMappingTest.Cycles.valueCycleSeversCleanly` — bidirectional cycle (`alice.ref == bob && bob.ref == alice`) is severed at the 4th level instead of overflowing.

### Bug #3 detail — HibernateProxy cache key churn

`GETTER_INVOKERS` keyed by `pojo.getClass()`. For LAZY-fetched entities, each distinct `HibernateProxy` bytecode subclass created a separate `ClassValue` cache entry + triggered its own `LambdaMetafactory` build. `MetadataHolderProbe` similarly missed the `<persistentClass>Telescope` holder.

Fix: `Beans.persistentClassOf(Object)` unwraps `HibernateProxy#getHibernateLazyInitializer().getPersistentClass()` via reflection — without forcing proxy initialization. No hard Hibernate dependency in `:core`; the interface is loaded via `Class.forName` at class-init, null when Hibernate isn't on the classpath, falls through to `getClass()` (zero-cost).

Regression test: `BeansTest.PersistentClass` (4 tests) — null-safe, plain-POJO fall-through, subclass without HibernateProxy fall-through, `readProperty` on a subclass instance.

## Test plan

- [x] `:core:test` — all 280+ tests pass including 5 new regression tests
- [x] `:codegen:test` — all 47+ tests pass including the new `setOfScalarsCompiles` regression
- [x] `:lombok:test` — all 14 tests pass
- [x] `:examples:library:check` — green
- [x] `:examples:springboot:order-jpa:test` — green
- [x] `:examples:springboot:product-starter:test` — green
- [x] `:spring-boot-starter:test` — 7 tests pass
- [x] `:benchmarks` compile — clean
- [x] `./gradlew spotlessCheck` — green

## Lattice integrity

Every fix is structural: changes the **timing** (cycle guard), the **cache key** (proxy unwrap), the **dispatch resolver** (per-accessor optics), the **emitted preamble** (Set import), or the **error message** (known fields list). The optic composition surface — `Iso`, `Lens`, `Prism`, `Affine`, `Traversal`, `.then(...)` rules — is untouched. The lattice carried these fixes through reliably; no `Function<Object, Object>` plumbing introduced.

## What's NOT in this PR

The 8 demo slices that shipped clean from the bug-hunt workflow (Map auto-lift, mapper.patch Lombok→record, updateValidated + ControllerAdvice, multi-edit HTTP, @Bridge entity, redact lossy Iso, @Convert AttributeConverter, writeBean per-class) are valuable demo additions but separate from the bug fixes. They can land as a follow-up PR once these fixes are reviewed and merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,17 @@
 <p align="center">
-  <img src="img/logo.png" alt="telescope — deep-copy DSL for Java records and POJOs" width="320" />
+  <img src="img/logo.png" alt="telescope — optics-based DSL for Java records and POJOs" width="320" />
 </p>
 
 # telescope
 
-**Deep-copy DSL for Java records and POJOs.**
+**Optics-based DSL for Java records and POJOs.**
 
-One type. No category-theory jargon. Update fields deep inside immutable records — through lists, sets, maps, optionals,
-and sealed-type variants — without writing copy constructors by hand. Got POJOs? Navigate them natively or bridge them
-to records; the same DSL applies.
+One type drives deep navigation, immutable update, bidirectional mapping, and effectful update across records, plain
+POJOs, and Lombok `@Data` classes — no category-theory jargon, no hand-written copy constructors. Reach fields nested
+through lists, sets, maps, optionals, and sealed-type variants in one chain. Skip codegen for the runtime path, or add
+`@Focus` / `@BeanFocus` / `@Bridge` annotations for compile-time-bound navigators with deep recursion through
+containers. Drop-in `telescope-spring-boot-starter` for autoconfig + a typed `Mapper<A,B>` bean registry — one
+dependency, zero wiring.
 
 [![JVM 25+](https://img.shields.io/badge/JVM-25%2B-brightgreen.svg?&logo=openjdk)](https://openjdk.org/projects/jdk/25/)
 [![Build](https://github.com/eschizoid/telescope/actions/workflows/ci.yaml/badge.svg)](https://github.com/eschizoid/telescope/actions/workflows/ci.yaml)
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
@@ -608,6 +608,7 @@ public abstract class AbstractTelescopeProcessor extends AbstractProcessor {
         out.println("import java.util.List;");
         out.println("import java.util.Map;");
         out.println("import java.util.Optional;");
+        out.println("import java.util.Set;");
         out.println("import java.util.concurrent.CompletableFuture;");
         out.println("import java.util.concurrent.Executor;");
         out.println("import java.util.function.BiFunction;");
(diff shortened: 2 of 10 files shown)


eschizoid:
@copilot please review

PR #50, opened 2026-06-09
feat(examples): land 13-slice Tier A/B/C bughunt demo bundle

eschizoid:
## Summary

The 13 worktree-isolated demo slices from the multi-agent Tier A/B/C bughunt workflow, flattened into `examples/springboot/order-jpa/.../bughunt/<slice>/`. Each slice is a self-contained Spring Boot scenario exercising one corner of telescope on a realistic record+entity pair, with a Spring-integration test asserting the user-facing flow.

## Slices

| Slice | What it pins |
|---|---|
| `bridge` | `@Bridge` round-trip on invoice header + line item |
| `bulkupdate` | `Telescope.all(over(...))` across N optional patch fields |
| `inspect` | DSL introspection controller (read / find / count / exists) |
| `jpaconverter` | JPA `AttributeConverter` + Telescope nav over the converted field |
| `jpacycle` | Self-referencing Employee entity — type-level + value-level cycle resolution |
| `lazyfetch` | Hibernate LAZY proxy unwrap on the read path |
| `maps` | `Map<K, V>` auto-lift via `DeepMap.autoIso` |
| `partnerpatch` | `patch` on `Mapper<A, B>` against a partial-update PATCH endpoint |
| `redacted` | Sensitive-field rewriting via `Telescope.all` + bridge |
| `sealed` | Sealed-type `Payment ↔ PaymentEntity` narrow after paradigm hop |
| `setfield` | `Set<X>` typed terminal via `SetPath.each()` |
| `validated` | `updateValidated` accumulating errors through a nested traversal |
| `writebean` | Per-class `WriteHint` override (`CONSTRUCTOR` vs `SETTERS`) |

## Follow-on lattice fix

The `jpacycle` slice surfaced one small composition bug between PR #49's cycle interruption and `Iso.liftOptionalToNullable`:

```java
// before — NPEs on cycle-severed null
y -> y == null ? Optional.empty() : Optional.of(element.from(y))
// after — cycle-severed null collapses to Optional.empty()
y -> y == null ? Optional.empty() : Optional.ofNullable(element.from(y))
```

Composes correctly with `DeepMap.cycleSafe`'s null-return strategy. Javadoc updated.

## Base branch

Based on `fix/dispatch-tier-bugs-from-bug-hunt` (#49) — 4 of the 13 slices need its fixes to pass green. Once #49 merges to main, this PR's base auto-rebases and the diff drops to slices + the one Iso fix.

## Test plan

- [x] `./gradlew check` green across all modules locally
[...]

PR #51, opened 2026-06-09
Fix/dispatch tier bugs from bug hunt

eschizoid:


eschizoid:
Superseded — contents already on main. After PR #50 merged into the wrong base (fix/dispatch-tier-bugs-from-bug-hunt instead of main) earlier today, I cherry-picked its squash commit `ddc2177` directly onto main as `df7015c`. Every file this branch claims to change is byte-identical to main now; the diff GitHub shows is stale (computed against the merge-base from when the PR was opened, before the cherry-pick landed). Closing.

PR #53, opened 2026-06-09
fix(core): rename Edit.overIfPresent(BiFunction) → Edit.mapIfPresent

eschizoid:
## Summary

Quick follow-up to PR #52. While wiring \`Edit.overIfPresent(...)\` into the order-jpa \`BulkUpdateController\` I hit a real overload-resolution ambiguity at the call site:

\`\`\`java
// fails to compile: "reference to overIfPresent is ambiguous"
overIfPresent(LINE_ITEM_QUANTITIES, req.lineItemQuantityDelta(), (delta, q) -> q + delta)
\`\`\`

Even though only the \`BiFunction\` form has the right arity, Java's overload resolution can't pick between the two same-name 3-arg overloads while target-typing the lambda. Casting the lambda to \`BiFunction\` works but defeats the "reads clean" goal of the DSL.

## Change

Rename the BiFunction form from \`overIfPresent\` to \`mapIfPresent\`. The two read-modes are now distinguished by name, not by lambda arity:

| Method | Use when |
|---|---|
| \`overIfPresent(path, value)\` | DTO field type matches the leaf — direct replace |
| \`overIfPresent(path, value, mapper)\` | DTO type differs — map then replace |
| \`mapIfPresent(path, value, transform)\` | New leaf depends on the current leaf + carried value |

Mnemonic: **over** = the value yields the new leaf; **map** = the new leaf depends on the current leaf.

## Behavior

Unchanged. Same identity-on-null short-circuit, same lazy mapper invocation.

## Test plan

- [x] AllOverTest renames the BiFunction cases to \`mapIfPresent\`, adds a null-skip test for the new name
- [x] \`./gradlew :core:test\` green

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Edit.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Edit.java
@@ -83,10 +83,15 @@ public interface Edit<S> {
    * delta or an instruction rather than a replacement.
    *
    * <pre>{@code
-   * Edit.overIfPresent(LINE_ITEM_QUANTITIES, req.quantityDelta(), (delta, q) -> q + delta)
+   * Edit.mapIfPresent(LINE_ITEM_QUANTITIES, req.quantityDelta(), (delta, q) -> q + delta)
    * }</pre>
+   *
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #54, opened 2026-06-09
feat(core): Mapping.drop(srcAccessor) — explicit source-field exclusion

eschizoid:
## Summary

Adds a fourth row type to the deep-mapper varargs: `drop(srcAccessor)`. Claims a single source field with no target counterpart so the strict deep-mapping factory accepts the pair without requiring a same-name target property.

## Motivating example (from the order-jpa demo refactor)

```java
Telescope.mapper(
    Order.class, PartnerShippingLabel.class,
    to(Order::orderNumber, PartnerShippingLabel::getTrackingReference),
    via(Order::lineItems,  PartnerShippingLabel::getItems, partnerItemMapper),
    drop(Order::metadata));   // Order has internal metadata that
                              // mustn't leak to the partner DTO
```

Without `drop`, adding any field to the source side breaks every consumer mapper that doesn't have a target for it — wrong direction of pressure for a deep-mapping DSL.

## Behavior

- **Forward** (`mapper.forward(src)`): the dropped field is omitted from the target — no entry in `byTargetName`, the structural-iso never sees it.
- **Backward** (`mapper.backward(dst)`): the source-reconstructor produces `null` for the dropped slot. If your source records `null`-reject in their canonical ctor, you'll see that surface here.
- **Scoping**: a drop row applies only to the specific (source, target) pair the user passed to `Telescope.mapper(...)`. Drops on the same source class don't bleed across to other mappers.
- **Duplicate guard**: two `drop(src)` rows on the same source field fail-fast with the same `IllegalArgumentException` the duplicate-`to(...)` guard produces.

## Implementation

- New package-private `Drop<A, B, X>` record implements `Mapping<A, B>` + `MappingInternals<A, B>`. Carries only the src accessor; `targetClass()` and `targetField()` return `null` to mark "no target counterpart."
- `Mapping` sealed permits + `MappingInternals` sealed permits widen to include `Drop`.
- `DeepMap.groupOverridesByPair` binds `Drop` rows to the top-level `(source, target)` `TypePair` the user passed to `Telescope.map(...)`; a drop is scoped to the specific mapper that declares it, not every pair the recursion lands on.
- `DeepMap.populateIso` short-circuits on `row instanceof Drop` — adds the source field to `claimedSrc` (satisfying the strict-source pass) and registers a backward-only `FieldStep` that produces `null` in the source-reconstructor under the source-field key.
- `DeepMap.fieldIsoOf` gets a `Drop` case that throws — `Drop` never reaches `fieldIsoOf`, but the switch needs exhaustiveness for the sealed hierarchy.

## Test plan

- [x] 5 new `DropRow` cases under `DeepMappingTest`:
  - strict mode rejects unmapped source without `drop`
  - `drop` makes forward omit the field
  - `drop` makes backward leave the source slot `null`
  - `drop` composes with `to(...)` renames in the same mapper
  - duplicate `drop` is rejected with the same fail-fast guard as duplicate `to(...)`
- [x] `./gradlew :core:check` green
- [x] spotless clean

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/mapping/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/mapping/DeepMap.java
@@ -93,7 +93,7 @@ public final class DeepMap {
       hintMap.isEmpty() && defaultWriterFactory == null
         ? Reflective.BEANS
         : Reflective.beansWithHints(hintMap, defaultWriterFactory);
-    final var overrideTable = groupOverridesByPair(overrides.toArray(Mapping<?, ?>[]::new));
+    final var overrideTable = groupOverridesByPair(overrides.toArray(Mapping<?, ?>[]::new), source, target);
     final var cache = new HashMap<TypePair, Iso<?, ?>>();
     final var topSteps = new LinkedHashMap<String, FieldStep>();
     populateIso(source, target, overrideTable, beanRefl, cache, topSteps);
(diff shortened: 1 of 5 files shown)


eschizoid:
@copilot please review

PR #55, opened 2026-06-09
feat(core): two-arg Mapping.drop(srcAcc, target) for nested pair scope

eschizoid:
## Summary

Follow-up to PR #54. The single-arg `drop(srcAcc)` scopes the elision to the top-level mapper pair. The nested case — drop a field on a type the recursion lands on with a specific nested target — needs an explicit target class.

## Motivating example (order-jpa demo refactor, in progress)

```java
Telescope.mapper(
    Order.class, PartnerShippingLabel.class,
    to(Order::orderNumber,  PartnerShippingLabel::getTrackingReference),
    via(Order::lineItems,   PartnerShippingLabel::getItems, partnerItemMapper),
    drop(Order::metadata),                       // top-level (PR #54)
    drop(Customer::tags, PartnerCustomer.class));// nested pair (this PR)
```

Without the two-arg overload, the only ways to handle `Customer.tags` in this mapper would be (a) put `tags` on `PartnerCustomer` (semantic leak) or (b) reject the model extension.

## Change

- `Drop` record gains an `explicitTarget: Class<B>` component (nullable to distinguish single-arg vs two-arg).
- New static factory `Mapping.drop(Accessor<A, ?> src, Class<?> target)`.
- `DeepMap.groupOverridesByPair` uses `explicitTarget` when non-null, falls back to the top-level target otherwise. No behavioral change to single-arg call sites.

## Test plan

- [x] 3 new tests under `DeepMappingTest$DropRow`:
  - `twoArgDropAppliesAtNestedPair` — forward elides the nested field
  - `twoArgDropBackwardLeavesNull` — backward leaves the nested slot null
  - `oneArgDropOnNestedSourceDoesNotReachNestedPair` — documents why the two-arg form is needed
- [x] `./gradlew :core:test --tests '*DeepMappingTest*'` green
- [x] spotless clean

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/mapping/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/mapping/DeepMap.java
@@ -223,13 +223,14 @@ public final class DeepMap {
     final var grouped = new HashMap<TypePair, List<Mapping<?, ?>>>();
     for (final var row : overrides) {
       final var internals = internalsOf(row);
-      // Drop rows have no target accessor (targetClass() == null). Bind them to the top-level
-      // (source, target) pair the user passed to Telescope.map(...); a Drop row is scoped to the
-      // specific mapper that declares it, not to every pair the recursion lands on.
-      final var key =
-        row instanceof Drop<?, ?, ?>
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #57, opened 2026-06-09
refactor(core): flatten public package, drop @SuppressWarnings("exports")

eschizoid:
## Summary
- Move 12 files from `conversion/` + `mapping/` sub-packages into the root `io.github.eschizoid.telescope` package; `module-info.java` exports drop from 4 to 2. Every cross-class caller is now same-package, so `Telescope.wrap`, `Mapper(Iso,…)` constructor, and `DeepMap.resolve` tighten to package-private and every `@SuppressWarnings("exports")` goes away.
- The last seam (`Telescope.optic()`) is replaced by `public static Object opticOf(Telescope<?,?>)` — the `Object` return keeps `Iso`/`Lens`/`Traversal` out of every public signature; `MetadataHolderProbe` casts the opaque value to `Lens` at the consume site. Lattice mantra preserved: users still never type `Iso`, `Lens`, `Prism`, `Affine`, `Traversal`.
- Bundled: new `AdminTouchController` + test in `order-jpa` exercising `fieldByName(String)` as a runtime WRITE escape hatch (PLAN Tier 1 item 2) + 12 IDE-warning cleanups (type-param shadowing, redundant casts, `args.getFirst()`, unused params, etc.).

**Breaking change for downstream:** import paths flatten — `io.github.eschizoid.telescope.conversion.Mapper` → `io.github.eschizoid.telescope.Mapper`, same for every `mapping.*` type. Pre-1.0 with no known adopters per YAGNI; acceptable.

## Test plan
- [x] `./gradlew compileJava compileTestJava` — green across core, codegen, lombok, spring-boot-starter, all 4 example submodules, benchmarks
- [x] `./gradlew test` — full suite green (exit 0)
- [x] Order-jpa fieldByName tests: happy-path persist via `shippingAddress.city`, bad-leaf-name 400 with declaring class + known-fields list, bad-top-level 400 against `Order` components
- [x] IDE re-scan via IntelliJ inspections: 0 `@SuppressWarnings("exports")` remaining in `core/src/main` and `codegen/src/main`; `Telescope.java` carries only the documented type-witness false-positive warnings (`Class<S>` param for inference per CLAUDE.md)
- [ ] Copilot review pass (auto-triggered on this PR)


eschizoid:
@copilot please review

PR #60, opened 2026-06-10
feat(benchmarks): apples-to-apples MapStruct comparison + perf-honesty README update

eschizoid:
Closes PLAN.md Tier 1 item 1 — the last open hedge before the 1.0 cut.

## Summary
- New `MapStructComparisonBenchmark` — 18 rows (3 depth tiers × 2 directions × 3 engines). Same input fixtures held in `@State(Scope.Benchmark)` so all three engines time the SAME work; only the dispatch path varies.
- 12 fixture files (`Mc*` prefix) + 3 MapStruct `@Mapper` interfaces. MapStruct 1.6.3 added as `jmhAnnotationProcessor` + `jmhImplementation` (off main classpath).
- `--add-exports` flag fixed for the post-split internal module name.
- `benchmarks/README.md` Results section + new "MapStruct comparison" section.
- Root `README.md` "Performance honesty" paragraph **rewritten** with measured numbers — replaces the "we haven't run an apples-to-apples vs MapStruct benchmark" hedge.

## Results (JDK 25, Apple Silicon, 3 warmup + 5 measurement × 1 fork)

| Tier   | Direction       | MapStruct (ns/op) | Telescope codegen (ns/op) | Telescope runtime (ns/op) |
| ------ | --------------- | ----------------: | ------------------------: | ------------------------: |
| flat   | bean → record   |     3.66 ± 0.19  |              15.37 ± 3.27 |             371.85 ± 39.06 |
| flat   | record → bean   |     3.56 ± 0.12  |               6.85 ± 3.12 |             507.21 ± 23.47 |
| nested | bean → record   |     5.41 ± 0.56  |              17.86 ± 1.45 |             564.47 ± 59.91 |
| nested | record → bean   |     5.71 ± 0.80  |              11.53 ± 2.36 |             762.93 ± 37.40 |
| deep   | bean → record   |    53.88 ± 5.62  |             189.43 ± 4.62 |           2304.62 ± 119.68 |
| deep   | record → bean   |    70.81 ± 68.75 |             191.64 ± 14.09 |          2678.48 ± 592.70 |

## Honest read

**MapStruct wins on every row.** No spin. On the closest comparison (codegen vs codegen) telescope sits **2–4× behind** across the three tiers. The reflective runtime path is **30–100× slower** than MapStruct's generated bytecode.

**Why MapStruct wins on dispatch.** It emits one hand-templated method body per pair — every getter / setter / nested call is fully monomorphic and the JIT inlines the whole conversion into a single basic block. Telescope's `@Bridge` emits an `Iso` chain composed via the lattice's `.then(...)` rules. The composition is what makes the lattice reusable (Iso round-trips, threads through `Telescope` paths, lifts through containers via `Iso.liftList`) but adds dispatch indirection at the call site. On a tight bench loop the JIT inlines through it; the residual is the 2–4× delta.

**What this means.** Telescope's case isn't perf — it's the capabilities MapStruct doesn't compose to: sealed-narrow paradigm hop, effectful update (`updateAsync` / `updateValidated` / `updateEither` / `updateOptional`), JPA cycle handling, Hibernate LAZY proxy unwrap, deep navigation as a primitive. All demoed end-to-end in `examples/springboot/`. Both tools clear sub-microsecond on flat/nested and stay single-microsecond on deep — for typical request-handling code the 2–4× codegen delta is ~10–140 ns absolute. It matters when conversion is the hot loop in a batch pipeline; it doesn't when it's one of N steps in a REST → JPA flow.

## Test plan
- [x] `./gradlew :benchmarks:compileJmhJava` — clean
- [x] `./gradlew :benchmarks:jmh -Pjmh.includes=MapStructComparisonBenchmark` — all 18 rows green, exit 0
- [x] `./gradlew check -x :benchmarks:jmh` — full multi-module check passes (tests + spotless)
- [x] MapStruct `*Impl` classes + telescope `*Bridge` classes both generated (verified in `benchmarks/build/generated/sources/annotationProcessor/`)
- [ ] Copilot review (auto-triggered)

## Caveats logged in the README
- 5 measurement × 1 fork is the project's standard config; not publication-grade. `deep_mapstruct_backward` has a ±68.75 error bar wider than the score — backward direction on the deep tier was noisy this run. A 10 measurement × 3 fork run would tighten it; numbers and direction are reliable, tight CIs are a follow-up.
- Telescope `_codegen_backward` rows use `BRIDGE.set(placeholderBean, rec)` which discards the placeholder and invokes the underlying `Iso.from(rec)`. The `set` adds small constant overhead vs a direct `iso.from` call.


eschizoid:
@copilot please review

PR #62, opened 2026-06-10
perf(codegen): BridgeProcessor emits for-loop helpers for LIST/SET/MAP_VALUES

eschizoid:
Hotspot #5 from the dispatch analysis — the biggest single driver of the deep-tier gap to MapStruct.

## What changed

The `@Bridge` annotation processor was emitting **Stream pipelines** for container recursion:

```java
public static McCompanyRec forward(final McCompanyBean s) {
  return new McCompanyRec(
    s.getName(),
    s.getDepartments().stream().map(McDeptBeanToMcDeptRecBridge::forward).toList()
  );
}
```

Per call: Spliterator + ReferencePipeline + `Function::apply` per element, on lists of typically ≤4 elements. The deep-tier `@Bridge` benchmark composes through **two** such list-lifts (`List<Department>` → `List<Team>`), and that pipeline overhead is the dominant gap to MapStruct (MapStruct emits a presized `ArrayList` + for-loop directly).

This PR makes the processor emit the same shape:

```java
package demo;

import io.github.eschizoid.telescope.Telescope;
import java.util.ArrayList;
import java.util.List;

public final class McCompanyBeanBridge {

  private McCompanyBeanBridge() {}

  public static McCompanyRec forward(final McCompanyBean s) {
    return new McCompanyRec(s.getName(), __fwd_departments(s.getDepartments()));
  }

  public static McCompanyBean backward(final McCompanyRec t) {
    return new McCompanyBean(t.name(), __bwd_departments(t.departments()));
  }

  public static final Telescope<McCompanyBean, McCompanyRec> BRIDGE =
    Telescope.from(McCompanyBean.class).to(McCompanyRec.class).using(
      McCompanyBeanBridge::forward,
      McCompanyBeanBridge::backward);

  private static List<McDeptRec> __fwd_departments(final List<McDeptBean> src) {
    if (src == null) return null;
    final var out = new ArrayList<McDeptRec>(src.size());
    for (final var x : src) out.add(McDeptBeanToMcDeptRecBridge.forward(x));
    return out;
  }

  private static List<McDeptBean> __bwd_departments(final List<McDeptRec> src) {
    if (src == null) return null;
    final var out = new ArrayList<McDeptBean>(src.size());
    for (final var x : src) out.add(McDeptBeanToMcDeptRecBridge.backward(x));
    return out;
  }
}
```

One private static helper per **LIST / SET / MAP_VALUES** field whose element type carries a sub-bridge. Identity-element containers (same element type both sides) collapse to an inline `new ArrayList<>(readExpr)` copy — no helper emitted, since there's nothing to transform.

The processor also now collects the `java.util.*` imports each bridge needs and emits them as proper `import` statements instead of using FQNs inline (cleaner generated code, matches the convention of every other Java file in the codebase).

## Why this design over the alternatives

Considered three approaches:

1. **Shared utility class** (`BridgeContainers.copyList(src, fn)`) — adds a new public type to the surface; users would see it. The lattice mantra is "users never see internals," and a public copy-helper weakens that even labeled "internal."
2. **Single generic helper per bridge file** (`__copyList(src, fn)`) — keeps `Function::apply` per element (~1-3 ns/element JIT-inlined). Same engineering cost as per-field; smaller perf win.
3. **Per-field helpers (this PR)** — no `Function::apply` at all, direct static call inside the loop. Generated code matches MapStruct's pattern exactly. Slight bytecode increase per bridge (2 short helpers per list/set/map field).

Picked (3) because no `Function::apply` is the right ceiling, and the naming logic (`__fwd_<fieldName>`) is trivially deterministic.

## What's preserved

- **Public DSL surface: zero changes.** `Telescope.from(...)`, `@Bridge`, `Mapping.to/via` — all identical.
- **No new public types.** Helpers are `private static` on each generated bridge class.
- **Optional / Optional↔nullable container kinds: unchanged.** They already use `Optional.map(...)`, which doesn't have the Stream-pipeline overhead.

## README also updates

Added a new **"Per-field source/target mapping — side by side"** subsection in `How it compares to MapStruct`, walking the reader through the `@Mapping(source=..., target=...)` → `Mapping.to(srcAcc, tgtAcc)` equivalence and the safety trade-off. MapStruct trades the typed-ref ergonomics for path strings like `"user.address.street"`; telescope trades the string-path brevity for `javac`-time guarantees that an accessor still exists and types still match. The side-by-side makes that explicit.

## Test plan

- [x] `./gradlew :codegen:check :core:check :lombok:check` — green.
- [x] `./gradlew check` (full multi-module incl. four Spring Boot examples) — green.
- [x] `:codegen` tests for `BridgeProcessor` updated to assert the new emission: helper-call shape in forward/backward, helper signature, size-presized container allocation, and direct sub-bridge call inside the loop. Old stream-pipeline string assertions removed.
- [x] JMH `MapStructComparisonBenchmark` rerun deferred to a quiet machine (machine has been at load avg 60+ on 12 threads through this session; tracked in #61).

## Expected impact when the rerun lands

| Tier   | Direction       | PR #60 baseline | After PR #61 expected | After this PR expected |
|--------|-----------------|----------------:|----------------------:|-----------------------:|
| deep   | bean → record   |         189.43  |                  ~100 |                  ~70   |
| deep   | record → bean   |         191.64  |                  ~120 |                  ~90   |
| nested | bean → record   |          17.86  |                  ~10  |                   ~8   |
| nested | record → bean   |          11.53  |                 11.53 |                   ~8   |

Deep tier should close to within ~1.3-1.5× of MapStruct's 54-70 ns. Flat tier unaffected — no list recursion.


eschizoid:
@copilot please review

PR #64, opened 2026-06-11
docs(readme): rewrite "Performance honesty" with measured numbers + refresh "What it is not"

eschizoid:
Extracted from PR #60 (https://github.com/eschizoid/telescope/pull/60) so the narrative ships independently of the bench code.

## What changes

Two README-only edits in the same neighborhood of the file:

### 1. "Performance honesty" — replaces the hedge with measured numbers

Old paragraph said: *"we haven't run an apples-to-apples vs MapStruct benchmark"*. PR #60 fixed that by adding the benchmark; this PR updates the narrative around it.

The new section drops the full results table (3 tiers × 2 directions × 3 engines = 18 rows from `MapStructComparisonBenchmark`) plus an honest read:

- **MapStruct wins every row.** No spin.
- **Codegen vs codegen:** telescope sits **2-4× behind** across tiers.
- **Reflective runtime path:** **30-100× slower** than MapStruct's generated bytecode.

And reframes telescope's case explicitly as the capabilities MapStruct doesn't compose to — sealed-narrow paradigm hop, effectful update, JPA cycles, Hibernate LAZY unwrap, deep navigation as a primitive — not raw dispatch speed.

The comparison table row that previously read *"Per-field dispatch perf (codegen path, ADR-0006)"* is now *"Dispatch perf — codegen vs codegen"* with a direct pointer to the new section.

### 2. "What it is not" — refreshed bullets

Option-A "sharp pivot" rewrite:

- **"Not a MapStruct replacement"** — credits what shipped since the original bullet was written (`telescope-quarkus` Arc extension, four Spring Boot examples), states the asymmetry up front (MapStruct wins on dispatch, telescope wins on composable capabilities), and links the measured table.
- **"Not a fuzzy auto-mapper"** — unchanged (still accurate).
- **"Not category theory"** — explicitly names the package-private lattice types (`Iso`, `Lens`, `Prism`, `Affine`, `Traversal`) and the JPMS boundary that keeps them off the public surface.

## What this depends on

This PR references `MapStructComparisonBenchmark` and `benchmarks/README.md#mapstruct-comparison-apples-to-apples`. Those land on main with **PR #60** (the bench code itself). Merge order: **#60 first, then this PR.** The README anchor links will be dead until #60 lands, but the file itself is independent of the bench.

## Test plan

- [x] No code changes; nothing to test.
- [x] Markdown renders correctly (tables, anchors, code blocks).
- [ ] Cross-link check after #60 merges (the `#mapstruct-comparison-apples-to-apples` anchor lives in `benchmarks/README.md` from #60).

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,16 +144,22 @@ capability lists, vs-MapStruct callouts, and benchmark cross-links.
 
 ## What it is _not_
 
-- Not a MapStruct replacement. MapStruct is a purpose-built bean-mapping framework with deep ecosystem integration
-  (Spring, Quarkus, MicroProfile) and ten years of edge-case fixes. Telescope is an optics DSL that includes
-  bidirectional mapping as one capability — see [How it compares to MapStruct](#how-it-compares-to-mapstruct) below for
-  the actual side-by-side.
-- Not a fuzzy auto-mapper. `Telescope.map(...)` matches fields by exact name and type, nothing more — no fuzzy name
+- **Not a MapStruct replacement — and not trying to be.** MapStruct emits one hand-tuned method body per type pair and
+  wins on dispatch speed (measured — 2-4× on flat conversion; see [Performance honesty](#performance-honesty)).
+  Telescope's case is the capabilities MapStruct doesn't compose to: deep navigation, effectful update (`updateAsync` /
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #65, opened 2026-06-11
feat(mapping): Mapper.withPath family — closes the @Mapping(source/target nested) gap

eschizoid:
Closes the **`@Mapping(source="a.b.c", target="x.y.z")`** gap with MapStruct that surfaced as a feature ask earlier today.

Three new methods on `Mapper`, each with a documented cardinality semantic:

| Method | Cardinality | What it does |
|---|---|---|
| `withPath(srcPath, tgtPath)` | single → single | Phase 1. Both paths terminate at a single value. Forward stamps source value at the target leaf; backward reverses it. Round-trip preserved. |
| `withBroadcastPath(srcPath, tgtPath)` | single → many | Phase 2. Source is a single value; target traverses (e.g. `.each(...)`). Forward writes the value to every focus. Backward assumes foci agree — round-trip law caveat documented. |
| `withZipPath(srcPath, tgtPath)` | many → many | Phase 2. Both paths traverse collections. Forward writes positionally (source's Nth value → target's Nth focus). Cardinality mismatch throws — silent truncation would be a footgun. |

Plus convenience overloads accepting an `Accessor` on either side, lifted to a one-hop `Telescope` via the existing `sourceClass` / `targetClass` on `Mapper`.

## Why this shape

- **Path is a real `Telescope<S, A>`** — composed with the same `.field(...)` / `.each(...)` navigation users already use for read / update. IDE refactors follow each hop and `javac` catches missing accessors everywhere.
- **No DeepMap changes, no new sealed permits.** Pure composition layer over an existing `Mapper`. Users compose:
  ```java
  Telescope.mapper(A.class, B.class, /* flat rules */)
      .withPath(srcPath, tgtPath);
  ```
- **Three distinct method names** (vs. one overloaded method) so the call site signals cardinality intent to a reader. `withZipPath(...)` reads differently from `withBroadcastPath(...)` — that's the point.

## Side by side with MapStruct

```java
// MapStruct
@Mapper
public interface OrderMapper {
  @Mapping(source = "customerName", target = "shipping.recipient.fullName")
  @Mapping(source = "createdAt",    target = "audit.createdDate")
  OrderDto toDto(Order order);
}
```

```java
// telescope — chain withPath calls on the base Mapper
final var mapper = Telescope.mapper(Order.class, OrderDto.class)
    .withPath(
        Order::getCustomerName,
        Telescope.of(OrderDto.class).field(OrderDto::getShipping).field(Shipping::getRecipient).field(Recipient::getFullName))
    .withPath(
        Order::getCreatedAt,
        Telescope.of(OrderDto.class).field(OrderDto::getAudit).field(Audit::getCreatedDate));

final OrderDto dto  = mapper.forward(order);
final Order    back = mapper.backward(dto);  // path-rules round-trip both ways
```

## Test plan

- [x] `./gradlew check` (full multi-module incl. four Spring Boot examples) — **green**.
- [x] `NestedPathMapperTest` — 7 scenarios across both phases:
  - flat-source accessor → nested-target path (forward + backward round-trip)
  - nested-source path → nested-target path
  - layering multiple `withPath` calls
  - broadcast: single → many (every focus stamped)
  - zip: positional N:N with matching cardinality
  - zip: cardinality mismatch throws `IllegalStateException`
- [x] All 7 tests **PASSED**.

## Round-trip notes (Phase 2)

- **Broadcast:** backward reads the **first focus** of the target path and writes it back to the source. If the foci disagree on the actual value at backward time, the iso law breaks. Use this semantic only when you know all foci hold the same value. The README documents this caveat alongside the example.
- **Zip:** cardinality must match at apply time, both forward and backward. Throws `IllegalStateException` with the actual N vs M counts when violated.

## Cleanup also included

Removed a duplicate "Per-field source/target mapping" section that had landed twice in `main` (merge artifact from #62 → #64 sequencing). The remaining section's "Nested path" row in the comparison table is updated to point at `Mapper.withPath` (previously claimed `Mapping.via` covered this, which it doesn't perfectly).

## What this does NOT include

- **No new `Mapping.to(srcPath, tgtPath)` factory.** Full integration with the `Mapping` sealed interface + `DeepMap` engine would require additional design work (target field claiming, conflict resolution, post-fixup ordering). The Mapper-level composition gets the feature to users now; sugar can land in a follow-up.
- **No path arity detection.** `withPath` works on single-focus paths; the broadcast/zip methods exist as separate named entry points. A single overloaded `withPath` that auto-dispatches on the underlying optic shape is a possible future refinement.

## Stacked / merge order

Off `main` directly. Independent of the still-open PRs (#60 / #64 are README+bench; can land in any order with this).

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,7 +227,7 @@ Telescope.map(Order.class).to(OrderDto.class)
 | **Typo / type-mismatch caught at**   | Annotation-processor run                                | **`javac` compile time** — the wrong-type accessor doesn't compile         |
 | **Survives a rename (IDE refactor)** | String breaks; processor re-runs and surfaces the error | IDE refactor follows the accessor everywhere                               |
 | **Reverse direction**                | A second method with `@InheritInverseConfiguration`     | Same `Mapping.to(...)` row works both ways                                 |
-| **Nested path** (`source = "a.b.c"`) | Expression-string                                       | `Mapping.via(srcAcc, tgtAcc, nestedMapper)` — typed at every hop           |
+| **Nested path** (`source = "a.b.c"`) | Expression-string                                       | `mapper.withPath(srcPath, tgtPath)` — fully typed path on either side      |
 | **Custom expression**                | `@Mapping(expression = "java(...)")`                    | `Mapping.via(srcAcc, tgtAcc, customMapper)` — plain Java mapper, type-safe |
 | **`condition = "..."` predicate**    | Annotation attribute                                    | `Edit.overIfPresent(...)` for updates; `Mapping.drop(...)` for mappings    |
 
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

eschizoid:
Closing — what shipped here is not the API we agreed on. The Option A you picked was `Mapping.to(srcAcc, tgtPath)` as a varargs row in `Telescope.mapper(A.class, B.class, ...rows)`. What I shipped (`Mapper.withPath` / `withBroadcastPath` / `withZipPath`) is a different shape — chained methods on `Mapper` instead of declarative rows, with two of the three method names using FP jargon. Redoing as proper Option A in a fresh PR.

PR #66, opened 2026-06-11
feat(mapping): Mapping.to(srcAcc, Telescope) for nested-target rows + Path → Telescope rename across the repo

eschizoid:
Bundled rebuild of the closed [#65](https://github.com/eschizoid/telescope/pull/65) with the **actual Option A** you picked plus the **`Path` → `Telescope` rename** scrub.

## 1. `Mapping.to(srcAcc, Telescope)` — proper Option A

New static factory on the sealed `Mapping` interface. Lives in the same varargs row position as `Mapping.to(srcAcc, tgtAcc)` / `Mapping.via(...)` — no chained method on `Mapper`, no jargon names.

```java
    to(Order::getName, OrderDto::getFullName),
    to(Order::getCustomerName,
       Telescope.of(OrderDto.class)
           .field(OrderDto::getShipping)
           .field(Shipping::getRecipient)
           .field(Recipient::getFullName)));
```

Closes the gap with MapStruct's `@Mapping(source = "flat", target = "a.b.c")` via the lattice's public surface (`Telescope`) for the nested write — no new "Path" abstraction introduced.

**Implementation:** new sealed permit `TelescopeTo<A, B, X>` is the row's data carrier. `DeepMap` recognises it at the outer pair only — segregates path rows from the flat field-claim assembly, applies them as a post-fixup overlay on top of the auto-recursion result:

- **Forward:** `tgtTelescope.set(b, srcAcc.apply(a))` after the base produces `b`.
- **Backward:** read `tgtTelescope.read(b)`, then rebuild `a` via the source-side `Reflective.construct(...)` with the source field overridden.

`Telescope.set` and `Telescope.read` from the lattice's public surface do all the actual work — no new lattice machinery introduced.

## 2. `Path` → `Telescope` rename across the whole repo

Every previous `Path` surface renamed to use lattice-consistent vocabulary:

| Today | After this PR |
|---|---|
| `Telescope.ListPath` (public typed subclass) | `Telescope.ListTelescope` |
| `Telescope.SetPath` | `Telescope.SetTelescope` |
| `Telescope.MapPath` | `Telescope.MapTelescope` |
| `Telescope.OptionalPath` | `Telescope.OptionalTelescope` |
| `<X>Path<R>` (codegen navigator) | `<X>Telescope<R>` |
| `<X>Telescope` (codegen metadata holder, ADR-0006) | `<X>FieldOptics` |
| `.start()` (codegen-emitted static factory) | `.focus()` |
| `OrderPathController` (Spring example) | `OrderTelescopeController` |
| `@RequestMapping("/orders/path")` | `/orders/telescope` |

`MetadataHolderProbe` updated to look up `<X>FieldOptics`. All tests, fixtures, examples, READMEs, and the per-module READMEs updated to use the new names.

## Why bundled

The `Mapping.to(srcAcc, tgtTelescope)` API requires `Telescope` (not "Path") to be the public-facing name of the lattice's navigator type. Renaming `<X>Path` → `<X>Telescope` is the load-bearing change that makes the new factory's signature read consistently with the rest of the codebase. Doing them in one PR keeps the rename atomic with the API change it enables — no "use `Mapping.to(srcAcc, Telescope)` … by passing a Path" hand-wave in the docs during the transition.

## Test plan

- [x] `./gradlew check` (full multi-module) — **GREEN**
  - `:internal` (incl. MetadataHolderProbeTest verifying the renamed `<X>FieldOptics` lookup)
  - `:core` (incl. new TelescopeTo integration tests via DeepMappingTest)
  - `:codegen` (in-memory ProcessorHarness against all 4 processors with renamed emission)
  - `:lombok` (file-based integration tests against the actual renamed `<X>Telescope` navigators)
  - `:examples:library`, `:examples:springboot:invoicing`, `:examples:springboot:order-jpa`, `:examples:springboot:org-chart`, `:examples:springboot:product-starter` (real Spring Boot apps with H2 + Hibernate exercising the renamed navigators end-to-end)

## What this PR does NOT include

- **No collection-side `Mapping.to(srcAcc, tgtTelescope)` semantics** (Phase 2 broadcast/zip from the closed #65). The flat-source → nested-scalar-target case ships now; collection paths can land in a follow-up once we've stared at the round-trip semantics more.
- **No path-on-source `Mapping.to(srcTelescope, tgtAcc)` factory** (the nested-source case). Same reason as above — design needs another pass before shipping.
- **No `:benchmarks` rerun** of the MapStruct comparison. Still deferred to a quiet machine (the load-avg-60+ issue from earlier in the session).

## Closes

- Closes the gap behind [#65](https://github.com/eschizoid/telescope/pull/65)
- Implements the `<X>Path` rename surface the user asked for in the linked discussion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,13 +133,14 @@ showcase. All changes are additive — no breaking changes vs 0.4.0.
 
 - Internal: holder-aware dispatch sites prepared in `Records` / `Beans` ahead of Phase B wiring.
 - **`Telescope.asList` / `.asSet` / `.asMap` / `.asOptional` no longer short-circuit on `instanceof`.** The four typed-
-  container promotion methods previously had `if (path instanceof ListPath<?, ?> lp) return (ListPath<S, X>) lp;`
-  branches that saved an allocation when the caller already held a typed subclass. The branches required
+  container promotion methods previously had
+  `if (path instanceof ListTelescope<?, ?> lp) return (ListTelescope<S, X>) lp;` branches that saved an allocation when
+  the caller already held a typed subclass. The branches required
   `@SuppressWarnings({"unchecked", "exports", "CastConflictsWithInstanceof"})` because of the wildcard projection; in
-  practice they almost never fired (`.list(getter)` returns `ListPath` directly; `asList(...)` callers are promoting
-  fresh `Telescope<S, List<X>>` builds that need the allocation regardless). Dropped — each method now shrinks to a
-  single `@SuppressWarnings("exports")`. Behaviour-equivalent; observable difference is only instance identity on the
-  rare already-typed-subclass case.
+  practice they almost never fired (`.list(getter)` returns `ListTelescope` directly; `asList(...)` callers are
+  promoting fresh `Telescope<S, List<X>>` builds that need the allocation regardless). Dropped — each method now shrinks
+  to a single `@SuppressWarnings("exports")`. Behaviour-equivalent; observable difference is only instance identity on
+  the rare already-typed-subclass case.
 - **Examples module reshaped from a single `Main.java` orchestrator to per-demo Gradle tasks.** Each demo class
   (`CodegenDemo`, `DeepMappingDemo`, etc.) carries its own `static void main()`; `examples/build.gradle.kts` registers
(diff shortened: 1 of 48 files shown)


eschizoid:
@copilot please review

PR #67, opened 2026-06-11
perf(records): cache reader at Lens construction in Records.fieldLens

eschizoid:
## Summary

The runtime `Records.fieldLens(String)` overload re-resolved `(class, name) -> idx` on every `lens.get(source)` call — a linear scan over the cached `RecordComponent[]` to translate the field name into the array index used to pick the LMF reader. For the method-reference path (`Telescope.of(User.class).field(User::name)`), that scan was pure overhead: the declaring class is captured at `.field(...)` construction time via `Telescope.implClassOf(getter)`, so the index can be baked in once.

This PR:

- Adds `Records.fieldLens(Class<S> cls, String fieldName)` — resolves the per-class `RecordInfo`, the component index, and the LMF-built `Function<Object, Object>` reader once at construction. The returned Lens's `get` / `set` / `modify` dispatch through the captured handles directly.
- Refactors the rebuild path so the class-aware lens reuses the existing canonical-constructor invoker via a new private `rebuildWith(info, idx, source, fn)` helper. `updateField(...)` (the string-only path) delegates to the same helper after its own resolution step, so there's a single rebuild implementation.
- Wires `RecordFieldOptics.lensFor(...)` in `Telescope.java` to call the new class-aware overload via the already-captured `implClass`. Falls back to the string-only overload only when `implClassOf(...)` returns `null` (a pathological lambda case the runtime already handles defensively).
- Keeps `Records.fieldLens(String)` working unchanged. It remains the fallback for `fieldByName(String)`, where the source class genuinely isn't known until call time.

The LMF dispatch primitive itself (`reader.apply`) is unchanged. This PR only eliminates the per-call `indexOf(fieldName)` linear scan and the per-call `info(source.getClass())` `ClassValue` lookup that the `set` / `modify` paths used to repeat.

Hotspot #7 of the bughunt-fold-architecture batch. Sibling agent ships Hotspot #8 (`Beans.java`).

## Test plan

- [x] `./gradlew check` green across all modules (`:internal`, `:core`, `:codegen`, `:lombok`, `:benchmarks`, `:spring-boot-starter`, `:quarkus`, `:examples`)
- [x] `OpticLawsTest` get-set / set-get / set-set Lens laws pass
- [x] `TelescopeTest`, `FocusCodegenTest`, `BeanFocusCodegenTest`, `PojoOpticsTest` pass
- [x] `HybridDispatchIntegrationTest` (verifies the unannotated-record fallthrough into `Records.fieldLens`) passes
- [x] `spotlessApply` clean

## Caveats

- The bridge-bench numbers won't move — different path (codegen-generated `Iso<A, B>`). This PR helps `Telescope.of(Class).field(...)` runtime navigation (the `ofBeanFieldUpdate` / `reflectionFieldUpdate` tier rows in the benchmark table).
- No new JMH benchmark added; the existing `HolderDispatchBenchmark.fallbackPath` exercises this path and should see a tighter number, but per the agent brief the JMH suite wasn't run here to avoid contaminating numbers under shared load.
- Pre-1.0 — no `@Deprecated` shim on the string-only overload per the project's no-deprecated-aliases convention. The string-only `fieldLens(String)` remains a first-class API for the `fieldByName(...)` escape hatch.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -1164,7 +1164,11 @@ public sealed class Telescope<
       final Class<A> implClass = Telescope.implClassOf(getter);
       final var holderLens = MetadataHolderProbe.<A, B>lensFromHolder(implClass, name);
       if (holderLens != null) return holderLens;
-      return Records.fieldLens(name);
+      // Pass the declaring class so the lens captures (info, idx, reader) at construction —
+      // eliminates the per-call (class, name) → idx scan that the string-only fieldLens(name)
+      // overload pays. The string-only overload remains the fallback for fieldByName(String),
+      // where the source class isn't known until call time.
+      return implClass != null ? Records.fieldLens(implClass, name) : Records.fieldLens(name);
     }
   }
 
--- a/internal/src/main/java/io/github/eschizoid/telescope/internal/Records.java
+++ b/internal/src/main/java/io/github/eschizoid/telescope/internal/Records.java
@@ -50,7 +50,12 @@ public final class Records {
    * A {@link Lens} over a record component, identified by name. The {@code get} reads the
    * component; {@code set}/{@code modify} return a copy of the record with that one component
    * replaced. Backs {@link io.github.eschizoid.telescope.Telescope}'s {@code .fieldByName(String)}
-   * overload.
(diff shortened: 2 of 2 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #68, opened 2026-06-11
perf(internal): Beans.lens captures LMF reader at construction

eschizoid:
## Summary

`Beans.lens(pojoClass, property, writer)` returned a Lens whose every
`get(source)` call paid the same per-call lookup overhead:

1. `Beans.readProperty(source, property)` →
2. `persistentClassOf(source)` (HibernateProxy unwrap check) →
3. `GETTER_INVOKERS.get(beanClass)` (ClassValue probe) →
4. `.get(name)` (HashMap lookup) →
5. `reader.apply(source)` (the LMF-built getter — the only useful step).

The `(pojoClass, property)` pair is fixed at construction time, so the
LMF-built `Function<Object, Object>` reader is too. The per-call probe
was ~20-30 ns of pure waste before the actual getter ran.

This change captures the reader (and the full per-class reader map,
for the off-path carry-over reads in `set` / `modify`) once at
construction. Per-call dispatch is a single instanceof + direct
`Function.apply` — no ClassValue probe, no HashMap lookup, the JIT
inlines through the captured `final` reference.

Misconfigured lenses now fail at build time with `IllegalArgumentException`
("No getter for property 'X' on Y") rather than at first read.

## Polymorphism preserved

- **HibernateProxy unwrap.** `persistentClassOf(pojo)` returns the parent
  class — by construction that's `pojoClass`. The proxy's runtime class
  is a generated subclass, so it takes the slow path and `readProperty`
  unwraps correctly. The captured `pojoClass` is the right key.
- **Subclass instances.** A `Lens<User, X>` applied to a `SuperUser`
  instance takes the slow path (`source.getClass() != pojoClass`) and
  routes through `readProperty`, which looks up the `SuperUser` entry in
  `GETTER_INVOKERS` (cached on first miss). Same correctness as before,
  same one-time per-class build cost, only the monomorphic case skips
  the per-call lookup.

The fast path is `source.getClass() == pojoClass`. The common
`Telescope.ofBean(User.class).field(User::getX)` usage hits this every call.

## Test plan

- `./gradlew check` — all modules green (full repo build, including
  `:internal`, `:core`, `:codegen`, `:lombok`, examples).
- Added two `BeansTest.LensFactory` tests:
  - `lensRejectsUnknownPropertyEagerly` — pin the new build-time IAE.
  - `lensWorksOnSubclassInstance` — `Lens<ParentBean, String>` reads
    `id` correctly when applied to a `ChildBean` instance (forces the
    slow path).
- Existing lens / writer / optic-laws tests unchanged and green.

The bridge-bench numbers in `:benchmarks` won't move (that path goes
through `Reflective.structuralIso` + `Beans.readProperty`, not through
`Beans.lens`). The benefit lands on `Telescope.ofBean(Class).field(...)`
runtime navigation — the typical user-facing reflective bean path.

### DIFF
--- a/internal/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
+++ b/internal/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
@@ -505,6 +505,23 @@ public final class Beans {
    * properties carried over) via {@code writer}. Immutable — it never mutates {@code source}.
    * Powers {@code Telescope.ofBean(...).field(...)}.
    *
+   * <p>The {@code (pojoClass, property)} pair is constant at construction, so the LMF-built {@link
+   * Function Function&lt;Object, Object&gt;} reader is resolved <em>once</em> via {@link
+   * #GETTER_INVOKERS} and captured by the returned {@code Lens}. Per-call {@code get(source)}
+   * dispatches directly through the captured reader — no per-call ClassValue probe, no per-call
+   * HashMap lookup. The lens fails fast at build time (rather than at first read) if {@code
+   * property} has no getter on {@code pojoClass}.
+   *
+   * <p>Subclass / proxy polymorphism is preserved through a single-instanceof fast-path: when
+   * {@code source.getClass() == pojoClass} the captured reader runs directly; otherwise the call
+   * falls through to {@link #readProperty(Object, String)}, which handles both Hibernate-proxy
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #69, opened 2026-06-11
perf(codegen): @Bridge dispatch goes through per-bridge BridgeFn impl (Hotspot #6)

eschizoid:
## Summary

The hotspot: the generated `<X>Bridge.BRIDGE` constant routes the per-call dispatch through `Iso.of(...)`'s `Function`-typed fields. javac synthesises one anonymous `Iso` class for that factory's `new Iso<>(){...}` body — every bridge in the JVM shares it, and the `forward.apply(source)` call site inside it goes megamorphic across many bridges, where the JIT can't inline through it.

Fix: codegen emits a per-bridge concrete dispatch type so each bridge constant sees one concrete implementing class at its call site.

- New public 2-method `BridgeFn<A, B>` in `core/.../conversion/` (forward + backward).
- New `Telescope.bridge(BridgeFn)` static factory that wraps an `Iso` calling `fn.forward(s)` / `fn.backward(t)` directly — no `Function::apply` hop.
- `BridgeProcessor` emits a `private static final class Fn implements BridgeFn<...>` per `@Bridge` whose `forward`/`backward` delegate to the existing static `<Bridge>.forward` / `<Bridge>.backward`. The `BRIDGE` constant is built via `Telescope.bridge(new Fn())`.

## Design — why Option A (`BridgeFn` interface), not Option B (expose `Iso`)

Option B would have codegen emit a concrete `Iso` subclass per bridge, but that requires exporting the internal `Iso` lattice type to user packages. CLAUDE.md and ADR-0001 are explicit: the lattice (`Iso` / `Lens` / `Prism` / `Affine` / `Traversal`) stays internal, users never type those names. So Option A: keep the lattice internal, surface a tiny 2-method `BridgeFn` interface that the codegen-emitted concrete dispatch type implements. The wrapping `Iso` lives once on `Telescope.bridge(...)`.

The internal lattice types are untouched — no changes to `Iso` / `Lens` / `Prism` / `Affine` / `Traversal` files. Same composition lattice, same `.then(...)` rules, same effectful-update plumbing.

## Test plan

- `./gradlew check` green across all modules: `:internal`, `:core`, `:codegen`, `:lombok`, `:spring-boot-starter`, `:quarkus`, `:benchmarks`, `:examples:library`, and all four Spring Boot example modules (`order-jpa`, `product-starter`, `invoicing`, `org-chart`).
- `BridgeProcessorTest` assertions migrated to the new emission shape: `Telescope.bridge(new Fn())`, the nested `Fn` impl, `BridgeFn` import, the static delegate bodies.
- New `TelescopeTest.bridgeFnFactory` test exercises the public `Telescope.bridge(BridgeFn)` factory end-to-end (read, update round-trip, composition through `.each(...).then(bridge).field(...)`).
- The MapStruct comparison benchmark rerun is deferred to a quiet machine — load average has been 60+ all session.

## Generated bridge — before / after

Before:
```java
public static final Telescope<UserEntity, UserDto> BRIDGE =
  Telescope.from(UserEntity.class).to(UserDto.class).using(
    UserEntityBridge::forward,
    UserEntityBridge::backward);
```

After:
```java
private static final class Fn implements BridgeFn<UserEntity, UserDto> {
  @Override public UserDto forward(final UserEntity s) { return UserEntityBridge.forward(s); }
  @Override public UserEntity backward(final UserDto t) { return UserEntityBridge.backward(t); }
}

public static final Telescope<UserEntity, UserDto> BRIDGE = Telescope.bridge(new Fn());
```


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #70, opened 2026-06-11
feat(mapping): nested-path Mapping factories + codegen .of() entry rename

eschizoid:
## Summary

Closes the gap with MapStruct's `@Mapping(source = "a.b.c", target = "x.y.z")` by letting `Mapping` rows accept multi-hop `Telescope`s on either side. All four shapes route through the existing optic lattice via `Telescope.set` / `read` / `toList` / `updateIndexed` — no new internal primitives.

### New public `Mapping` factories

| Factory | Shape | Engine |
|---|---|---|
| `to(Accessor<A, X>, Telescope<B, X>)` | flat src → nested tgt | `tgtT.set(t, srcAcc.apply(a))` |
| `to(Telescope<A, X>, Accessor<B, X>)` | nested src → flat tgt | rebuild `t` with `targetField = srcT.read(a)` |
| `to(Telescope<A, X>, Telescope<B, X>)` | both nested (broadcast on many-focus) | `tgtT.set(t, srcT.read(a))` |
| `zip(Telescope<A, X>, Telescope<B, X>)` | positional N:N, cardinality-checked | `tgtT.updateIndexed(t, …)` against `srcT.toList(a)` |

### Engine

- `DeepMap` groups overrides by `(sourceClass, targetClass)`, now with a **soft/strict claim split** so multiple `Mapping.to` rows reading the same source field compose (MapStruct's multi-`@Mapping` shape).
- Telescope rows' top-level field claim is recovered via `firstHopName` — a new package-private string on `Telescope` captured by `Telescope.lens(Accessor, BiFunction)`, the overload Java picks for codegen-emitted method-refs. `.then(...)` propagates the first non-null `firstHopName`, so composition chains preserve it.
- `Mapping` is now sealed over `TelescopeTo`, `FromTelescopeTo`, and `TelescopeToTelescope` (with a `Kind` enum discriminating `BROADCAST` vs `ZIP`).

### Ergonomic rename: `<X>Telescope.focus()` → `<X>Telescope.of()`

Symmetric with runtime `Telescope.of(Class)`, and removes the "focus" doubling that made chains like `PersonTelescope.focus().address().city()` read awkwardly. ~50 call sites swept across `codegen`, `lombok`, examples, READMEs, and ADR-0006.

```java
// Before
PersonTelescope.focus().address().city()
// After — symmetric with runtime Telescope.of(Person.class)…
PersonTelescope.of().address().city()
```

### Codegen 1:1 verification

`TelescopeMappingTest.CodegenOnePointOne` exercises real `@Focus`-generated navigators (`Person` / `Address` top-level test fixtures) as both the source AND target side of `Mapping.to(...)`, proving the codegen navigator value is interchangeable with a runtime `Telescope.of(...).field(...)` chain at the `Telescope<S, A>` value level — not at the source-text level.

`Mapping.java` javadocs now document both runtime and codegen call shapes side-by-side so adopters see `PersonTelescope.of().address().city()` as a first-class argument to `Mapping.to`.

## Test plan

- [x] 13 telescope-mapping tests pass (`TelescopeMappingTest` covers all four factories, codegen 1:1, cardinality enforcement on zip, multiple rows sharing source field, deeply-nested telescopes on both sides)
- [x] `:core` / `:codegen` / `:lombok` / `:benchmarks` all green
- [x] All four `examples/springboot/*` and `examples/library` compile after the `.focus()` → `.of()` sweep
- [x] Lombok integration tests pass — `<X>Telescope.of()` is reachable by reflection from `LombokFocusProcessorTest`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1101,7 +1101,7 @@ Telescope.of(Company.class)
   .update(company, String::toLowerCase);
 
 // Compile-time, reflection-free — same Telescope, generator-built
-CompanyPath.focus()
+CompanyPath.of()
   .departments().each().teams().each()
   .users().each().email()
   .update(company, String::toLowerCase);
@@ -1118,7 +1118,7 @@ import io.github.eschizoid.telescope.annotations.Focus;
 // Generated: <X>Path<R> per annotated type plus a step class per collection-shaped component.
 // Usage reads like the reflective DSL — but every hop is type-checked by javac and every read /
 // rebuild is a direct method-ref + constructor call (no reflection):
-final Telescope<Company, String> userNames = CompanyPath.focus()
+final Telescope<Company, String> userNames = CompanyPath.of()
   .teams().each()        // step over List<Team> → TeamPath<Company>
   .users().each()        // step over List<User> → UserPath<Company>
   .name();               // terminal Telescope<Company, String>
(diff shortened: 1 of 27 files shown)


eschizoid:
@copilot please review

PR #71, opened 2026-06-11
chore: post-PR-#70 cleanup + javadoc fixes + 30-sec Mapping hero

eschizoid:
Follow-up to **#70** (telescope-aware `Mapping` factories + `focus()` → `of()` rename), which merged at `5403f8d`. These two commits were stranded on `feat/mapping-telescope-factories` after the squash-merge.

### Javadoc @link rot blocks the release pipeline

The release workflow stages javadoc artifacts to Sonatype via JReleaser. Today `:core:javadoc` fails with 16 unresolved `@link` errors — most are pre-existing rot, three are new from PR #70. CI didn't catch this because `gradle check` doesn't run `:core:javadoc`, only the release task does. If we cut 1.0 without fixing this, the JReleaser deploy step fails on stage.

Fixes are surgical — add the missing imports per file:

- `mapping/{Drop,FromTelescopeTo,MappingInternals,TelescopeTo,TelescopeToTelescope}` → import `DeepMap`
- `mapping/{MapStep,WriteHint}` → import `Telescope`
- `effects/{Either,Validated}` → import `Telescope`
- `mapping/Mapping` → import `Edit`
- `runtime/instances/ValidatedK` → fix stale FQN `io.github.eschizoid.telescope.Either` → `effects.Either` (Either was relocated to `effects/` in a prior refactor; the FQN rotted)

### Cleanup of PR #70 leftovers

- **Drop the redundant `Mapping.drop(Src::displayCity)`** in `TelescopeMappingTest.realFocusNavigatorWorksAsTgtTelescope`. The engine's existing telescope-permissive mode (`DeepMap.populateIso:425`) already lets unclaimed source fields fall through when telescope fixups are present — the `drop` was defensive and unnecessary. Removing it tightens the codegen-1:1 example.
- **Document the intermediate-allocation limitation** in `Mapping.to(Accessor, Telescope)` javadoc. The overlay requires the target's intermediate structure to be reachable via auto-recursion. When the source is genuinely flat with no same-name field driving an intermediate, the overlay NPEs descending into a null intermediate. Workaround: `Mapping.via(srcAcc, tgtAcc, smallMapper)` with explicit intermediate allocation. v1.1 will close this gap by synthesizing intermediates from the telescope's hop chain.
- **Fix two stale `TelescopeZip` mentions** in `DeepMap` comments left over from the `Kind`-enum refactor (`TelescopeZip` was folded into `TelescopeToTelescope` with `Kind.BROADCAST` / `Kind.ZIP`).
- **Delete `CHANGELOG.md` and `CONTEXT.md`** — redundant with `git log` and unused at pre-1.0.

## Test plan

- [x] `:core:javadoc` now clean (was 16 errors before)
- [x] `:core:check` / `:codegen:check` / `:lombok:check` all green
- [x] All 13 telescope-mapping tests still pass after removing the redundant `Mapping.drop`


eschizoid:
@copilot please review

PR #72, opened 2026-06-11
feat(mapping): Mapping.constant + Mapping.compute + intermediate allocation — close the last MapStruct gaps pre-1.0

eschizoid:
Three MapStruct-parity gaps closed before 1.0 in one slice. Two commits, each independently green:

### `Mapping.constant(Tgt::field, value)` + `Mapping.compute(Tgt::field, Supplier<X>)` — forward-only target injection

Closes MapStruct's `@Mapping(constant = "...")` and `@Mapping(expression = "java(...)")` without splitting across `@InheritInverseConfiguration` interfaces. Both rows live next to the rest in the same `Telescope.mapper(...)` call.

```java
  to(Order::id,                OrderDto::id),
  constant(OrderDto::tenant,   "production"),     // eager literal
  compute(OrderDto::createdAt, Instant::now),     // fresh per call
  compute(OrderDto::traceId,   UUID::randomUUID),
  compute(OrderDto::metadata,  HashMap::new));    // fresh mutable container per call
```

**Semantics:** forward stamps the value; backward silently drops the slot (the rebuilt source carries the type default at the dual position — exact dual of `Mapping.drop` on the source side). `forward(backward(t))` does NOT round-trip a `compute` value by design — a supplier like `Instant::now` can't be inverted. Documented in javadoc + a JUnit `@DisplayName` pins the retraction behavior so a future contributor doesn't try to "fix" it.

**Why the eager/lazy split:** `Mapping.constant(Tgt::metadata, new HashMap<>())` would share one map reference across every forward call (Java mutable-default-argument trap). `Mapping.compute(Tgt::metadata, HashMap::new)` allocates a fresh container per call. MapStruct's `constant` vs `expression` distinction conflates these because their `constant` only accepts literals; ours is type-checked by `javac` at the call site.

### `Mapping.to(srcAcc, tgtTelescope)` — recursive intermediate allocation

Closes the v1.0 limitation previously documented in `Mapping.to(Accessor, Telescope)`'s javadoc. A flat source field routed through a multi-hop target telescope now lifts into the deep structure without per-hop allocation glue from the user:

```java
record Slim(String displayCity) {}        // genuinely flat — no `address` field
record Address(String city, String zip) {}
record Person(String name, int age, Address address) {}

Telescope.mapper(Slim.class, Person.class,
  to(Slim::displayCity,
     Telescope.of(Person.class).field(Person::address).field(Address::city)));

mapper.forward(new Slim("Brooklyn"));
// → Person(name=null, age=0, address=Address(city="Brooklyn", zip=null))
```

Previously this NPE'd because the auto-recursion left `address` null and the telescope overlay's `tgtT.set(t, value)` descended into it. Now the engine synthesizes a recursive default-tree instance — primitives get JLS defaults (0 / false / 0.0), records recurse via canonical-ctor reflection with the same scheme, everything else is null.

**Type-driven depth.** The engine doesn't walk the telescope's hop chain at all — once the FIRST hop is allocated as a default-tree, every deeper hop already exists in the tree (because the recursion is over field TYPES, not nav steps). Three-hop test (`outer.mid.inner.value`) passes with zero per-hop config.

**Records only for v1.0.** Bean intermediates without a no-arg ctor or builder still need a `Mapping.via(...)` workaround; deferred to v1.1 if it shows up in real usage. Cycles between record types can't arise (canonical ctors need every referenced type already constructible at compile time), so `recursiveDefault` doesn't need cycle detection.

**Sub-issue surfaced:** the existing `NULLING_ISO` returned `null` even for primitive-typed slots, NPEing when canonical-ctor reflection tried to unbox into `int` / `boolean` / etc. New `placeholderIsoFor(fieldType, claimedByTelescopeWrite)` unifies the three cases — record allocator, primitive default, plain null reference — into one type-aware dispatch.

## Engine cost

~50 LOC for constant/compute (two new permits mirror `Drop` on the target side, two iso factories). ~100 LOC for intermediate allocation (recursiveDefault + primitiveDefault + placeholderIsoFor + rawClassOf + defaultAllocatorIso). No new internal optic primitives; everything composes on the existing `Iso` surface.

## Tests

`MappingConstantComputeTest` (5):
- `constant` stamps literal forward; backward drops it (companion `drop(Sbase::junk)` verifies dual-direction symmetric handling)
- `compute(HashMap::new)` produces fresh container identity per call
- `compute(UUID::randomUUID)` produces distinct values per call
- counter-based supplier explicitly demonstrates `forward(backward(t))` does NOT round-trip

`MappingIntermediateAllocationTest` (3):
- one-hop intermediate from flat source (the original wart)
- two-hop nested (`address.geocode.lat`) — mixed primitive/reference depth
- three-hop nested (`outer.mid.inner.value`) — pins the "type-driven depth" claim

## Docs

- `Mapping.constant` / `Mapping.compute` javadocs explain the forward-only semantics, the eager-vs-lazy distinction, and the MapStruct parity (declared once, no `@InheritInverseConfiguration`).
- `Mapping.to(Accessor, Telescope)` javadoc rewritten — the v1.0 limitation note is gone, the recursive-allocation behavior and the bean caveat are in.
- PLAN.md Tier 2 #1 marked shipped pre-1.0.

## Test plan

- [x] All 8 new tests pass (`MappingConstantComputeTest` + `MappingIntermediateAllocationTest`)
- [x] Full `:core` test suite remains green (no regressions in existing `TelescopeMappingTest`, `DeepMappingTest`, `OpticLawsTest`, etc.)
- [x] `:codegen` / `:lombok` checks remain green

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -4,6 +4,8 @@ import io.github.eschizoid.telescope.conversion.Mapper;
 import io.github.eschizoid.telescope.internal.Beans;
 import io.github.eschizoid.telescope.internal.Reflective;
 import io.github.eschizoid.telescope.internal.optics.Iso;
+import io.github.eschizoid.telescope.mapping.Compute;
+import io.github.eschizoid.telescope.mapping.Constant;
 import io.github.eschizoid.telescope.mapping.Drop;
 import io.github.eschizoid.telescope.mapping.FromTelescopeTo;
 import io.github.eschizoid.telescope.mapping.MapStep;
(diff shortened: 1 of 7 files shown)


eschizoid:
@copilot please review

PR #73, opened 2026-06-11
feat(mapping): nested Mapping.constant/compute + unify Constant/Compute + codegen-1:1 tests across all three features

eschizoid:
Three threads in one PR — each cheap on its own, valuable as a unit.

## 1) Drop the Constant/ConstantTelescope split (and Compute/ComputeTelescope)

The post-PR-#72 surface had two internal record types per feature — one for the flat `constant(Accessor, X)` factory, one for the nested-target factory. Two engine handlers, two paths to keep in sync, for a feature users see as one concept ("stamp a value somewhere on the target").

**Unified**: one `Constant<A, B, X>` record carrying `Telescope<B, X>`. The flat factory wraps the bare accessor into a single-hop telescope at construction time via `Telescope.of(LambdaIntrospection.implClassOf(tgt)).field(tgt)`. Engine sees one shape, runs one handler. Compute follows the same pattern.

**Side effect**: the flat case now goes through the telescope-fixup pass instead of `byTargetName` / `FieldStep`. Per-call cost is one extra closure dispatch — negligible vs the rest of the engine. Half the surface, one path to reason about.

## 2) New nested-target overloads

```java
Mapping.constant(Telescope<B, X> targetTelescope, X value)
Mapping.compute(Telescope<B, X> targetTelescope, Supplier<? extends X> supplier)
```

Closes MapStruct's `@Mapping(target = "a.b.c", constant = "...")` and `@Mapping(target = "a.b.c", expression = "java(...)")` for nested targets:

```java
Telescope.mapper(Slim.class, Person.class,
  to(Slim::name, Person::name),
  constant(PersonTelescope.of().address().zip(),     "11201"),
  compute (PersonTelescope.of().audit().createdAt(), Instant::now));
```

Intermediate allocation (PR #72) kicks in automatically — a constant or compute at `PersonTelescope.of().address().zip()` allocates the `Address` intermediate before stamping. Records only; bean intermediates still need `Mapping.via`.

## 3) Codegen-1:1 verification across all three features

`MappingCodegenOneToOneTest` — 8 tests, each of the three target-side features exercised in three shapes:

| Feature | Flat accessor | Runtime telescope | Codegen navigator |
|---|---|---|---|
| **Intermediate allocation** (`Mapping.to(srcAcc, tgtTelescope)`) | n/a (single-hop) | ✓ | ✓ |
| **`Mapping.constant`** | ✓ | ✓ | ✓ |
| **`Mapping.compute`** | ✓ | ✓ | ✓ |

Plus a composed end-to-end test that combines all three features in one mapper through real `@Focus`-generated `PersonTelescope` navigators:

```java
Telescope.mapper(FlatSrc.class, Person.class,
  drop(FlatSrc::displayName),
  drop(FlatSrc::displayCity),
  compute (PersonTelescope.of().name(),                () -> "computed-name"),
  to      (FlatSrc::displayCity, PersonTelescope.of().address().city()),
  constant(PersonTelescope.of().address().zip(),       "11201"));
```

Pins the codegen-1:1 promise for the v1.0 MapStruct-parity slice end-to-end against the actual on-disk generated navigator classes.

## Engine changes

- `DeepMap.applyForward` gets two new switch cases (Constant, Compute) — each is a one-liner calling `tgtT.set(t, value)` / `tgtT.set(t, supplier.get())`.
- The PR-#72 `constantIso` / `computeIso` helpers are dropped (no longer used — flat-path-via-`byTargetName` is gone).
- Sealed permits collapse: `ConstantTelescope` / `ComputeTelescope` deleted; the unified `Constant` / `Compute` remain.

## Test plan

- [x] All 8 new tests in `MappingCodegenOneToOneTest` pass
- [x] Existing 5 `MappingConstantComputeTest` + 3 `MappingIntermediateAllocationTest` still pass after the Constant/Compute unification
- [x] Full `:core` test suite green (no regressions in `TelescopeMappingTest`, `DeepMappingTest`, `OpticLawsTest`, etc.)
- [x] `:codegen` / `:lombok` checks green

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -325,39 +325,24 @@ public final class DeepMap {
         telescopeFixups.add(ttRow);
         continue;
       }
-      // Constant / Compute rows are the target-side mirror of Drop: they claim a target field
-      // with no source counterpart, stamping a fixed (Constant) or freshly-supplied (Compute)
-      // value into the rebuilt target at forward time. Backward direction silently drops the
-      // slot — the rebuilt source carries the type default at the dual position (same retraction
-      // semantics as Drop, in the inverse direction). Register only under byTargetName so the
-      // forward pass picks them up; absence from bySourceName means the source rebuilder ignores
-      // the slot entirely on backward.
+      // Constant / Compute rows are target-injection telescope fixups: claim the target
+      // telescope's first hop as written, register the row in telescopeFixups, and let the
+      // applyForward pass stamp value (Constant) or supplier.get() (Compute) at the location the
+      // telescope navigates to. The flat factories (constant(Accessor, X) / compute(Accessor,
+      // Supplier)) wrap the accessor in a single-hop telescope at construction time so this loop
+      // sees only one shape per kind — no separate flat handler. Backward direction is a no-op:
+      // these rows don't contribute to bySourceName, so the source rebuilder ignores the slot
+      // entirely on backward (same retraction semantics as Drop on the source side).
       if (row instanceof Constant<?, ?, ?> cRow) {
-        final var tgtFieldName = tgtRefl.normalize(internals.targetField());
-        if (!claimedTgt.add(tgtFieldName)) throw new IllegalArgumentException(
-          "Deep map " +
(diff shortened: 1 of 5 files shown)


eschizoid:
@copilot please review

PR #75, opened 2026-06-12
refactor(internal): eliminate the MetadataHolderProbe static-init bridge (PLAN.md Tier 2 #2)

eschizoid:
Closes **PLAN.md Tier 2 #2**: replace the `MetadataHolderProbe` static-init bridge.

## What was the debt

`internal/MetadataHolderProbe` needed the underlying `Lens` from a codegen-emitted `Telescope` constant. `:internal` can't `requires` `:core` (the reverse dep already exists), so the workaround was a static-init callback: `Telescope`'s `<clinit>` registered a `Function<Object, Object>` extractor in `MetadataHolderProbe.opticExtractor`, then `HolderRef.lensFor(name)` called it to unwrap the constant.

Three things were wrong with that:

- **Mutable global state** — `volatile Function<Object, Object> opticExtractor`
- **Implicit load-order contract** — `Telescope`'s `<clinit>` must fire before any probe. Fine in practice (every probe site lives inside `Telescope`) but invisible to the reader and not enforced by anything.
- **Object-typed callback** — `Function<Object, Object>` hid a real type relationship: the input is always a `Telescope` and the output is always a `Lens`. This violates the lattice-first principle in `CLAUDE.md` ("a `Function<Object, Object>` field that could have been an `Iso<X, Y>` is a smell").

## What changes

The cast site moves to the side of the module boundary that owns the type. `:internal` becomes truly oblivious to `:core` — no callback, no global state, no static-init bridge.

```
Before                                          After
------                                          -----
:internal/MetadataHolderProbe                   :internal/MetadataHolderProbe
  - opticExtractor: volatile Function             (gone)
  - setOpticExtractor(fn)                         (gone)
  - HolderRef.lensFor(name): Lens<?,?>            (gone)
  - lensFromHolder(cls, name): Lens<S,A>          (gone)

:core/Telescope                                 :core/Telescope
  - static { setOpticExtractor(...) }             (gone)
                                                  + singleHolderLens(cls, name)
                                                  + holderReadersFor(cls, names)
                                                  + holderConstructorFor(cls)

:internal/Reflective                            :internal/Reflective
  - structuralIso(cls): builds holder data        + structuralIso(cls, holderReaders,
    internally via the bridge                       holderConstructor) — pre-resolved
                                                    by the caller in :core
                                                  - resolveHolderReaders(cls, names)  (gone)
                                                  - resolveHolderConstructor(cls)     (gone)
```

### File-by-file

- **`internal/Reflective.java`** — `structuralIso(cls)` becomes a thin delegate to a new holder-aware overload `structuralIso(cls, holderReaders, holderConstructor)`. The original no-arg form still works for any in-module callers that don't have holder data. Drops `resolveHolderReaders` + `resolveHolderConstructor`.
- **`internal/MetadataHolderProbe.java`** — drops `opticExtractor`, `setOpticExtractor`, `HolderRef.lensFor`, `lensFromHolder`, and the `import Lens`. `HolderRef` becomes a pure data carrier: `(holderClass, constantsByName, constructor)`. `constantsByName` was already typed `Map<String, Object>` — the raw constants are exactly what `:internal` can see.
- **`core/Telescope.java`** — drops the `static { MetadataHolderProbe.setOpticExtractor(...) }` block. Three new package-private helpers do the cast on the `:core` side:
  - `singleHolderLens(cls, name)` — per-field dispatch, used by `RecordFieldOptics` / `BeanFieldOptics`
  - `holderReadersFor(cls, names)` — all-or-nothing map for `Reflective.structuralIso`'s backward branch
  - `holderConstructorFor(cls)` — bound constructor for `structuralIso`'s forward branch
- **`core/DeepMap.java`** — `assembleIso` pre-resolves holder data via the three helpers and passes the results into the new `structuralIso` overload. The per-pair Iso composition still reads as pure lattice `.then()`.
- **`internal/MetadataHolderProbeTest.java`** — three `lensFromHolder*` assertions removed (the logic now lives in `:core`); two new tests assert the `HolderRef.constantsByName` shape that `:internal` can actually see. End-to-end holder-dispatch behavior is covered by the existing `Telescope` tests in `:core`.

## Invariants preserved

- **Zero `@SuppressWarnings("exports")`** — module-info exports unchanged.
- **No public API change** — `RecordFieldOptics` and `BeanFieldOptics` keep the same dispatch shape; the `FieldOptics` interface is unchanged.
- **All-or-nothing holder-readers semantics** — one branch outside the Iso's hot loop, not N branches inside.
- **Lattice-first** — nothing Object-typed remains. `holderReadersFor` returns `Map<String, Lens<Object, Object>>`; the constructor is the typed `Function` shape the holder emits.

## Perf

Full JMH suite ran post-change (48m 41s). Headline result: **`lensConstantUpdate` is perf-flat at 45.15 ns/op** vs. ~45 baseline — the codegen holder-dispatch path is exactly what we want.

Two runtime-deep-mapping benchmarks (`mapBeanForwardRead`, `fromBeanForwardRead`) show ~2× regression vs the historical CLAUDE.md baseline. Many other benchmarks show ~2× improvement (`bridgeForwardRead`, `recordComponentRead_lmf`, `reflectionFieldUpdate`, etc.) in the SAME run, which strongly suggests the historical baseline numbers were taken under different machine state (JDK patch level, background load) rather than a real regression in this change.

**Confirmation cycle pending**: a focused re-run on just `mapBeanForwardRead` + `fromBeanForwardRead` is kicking off in parallel to this PR. If the regression reproduces, the cause is most likely the two `Reflective.names(cls)` calls now made in `DeepMap.assembleIso` (vs. once inside `structuralIso` previously) — though theoretically those should be cached at the `Records` / `Beans` reflective layer. I'll post the focused numbers as a comment and update `CLAUDE.md`'s JMH table once we have stable confirmation.

## Test plan

- [x] `:core` test suite green (no regressions in `TelescopeTest`, `DeepMappingTest`, `TelescopeMappingTest`, etc.)
- [x] `:internal` test suite green (incl. `MetadataHolderProbeTest` with the rewritten shape assertions)
- [x] `:codegen` + `:lombok` test suites green
- [x] Full JMH suite completed in 48m 41s; `lensConstantUpdate` (the actually-changed dispatch path) is perf-flat
- [x] Focused re-run on `mapBeanForwardRead` + `fromBeanForwardRead` confirms perf-neutral — apples-to-apples on same machine state shows `main` and PR branch land at the same ballpark (~220–350 ns with wide JIT variance). The historical CLAUDE.md ~114/~142 numbers were taken under different machine state; full evidence in the PR comment above.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -897,8 +897,26 @@ public final class DeepMap {
     final Map<String, FieldStep> byTargetName,
     final Map<String, FieldStep> bySourceName
   ) {
-    final Iso<S, Map<String, Object>> srcReader = srcRefl.structuralIso(source).reverse();
-    final Iso<Map<String, Object>, T> tgtBuilder = tgtRefl.structuralIso(target);
+    // Pre-resolve the holder data on the :core side (Telescope is visible here) and pass into
+    // Reflective.structuralIso so :internal stays compile-time-oblivious to :core — no callback,
+    // no global state, no static-init bridge. holderReadersFor returns null when the holder is
+    // missing OR doesn't cover every component; the Iso then falls back to the reflective read
+    // path (matches the prior dispatch-shape invariant: one branch outside the Iso, not N
+    // branches inside).
+    final var srcNames = srcRefl.names(source);
+    final var srcHolderReaders = Telescope.holderReadersFor(source, srcNames);
+    final var srcHolderConstructor = Telescope.holderConstructorFor(source);
+    final var tgtNames = tgtRefl.names(target);
+    final var tgtHolderReaders = Telescope.holderReadersFor(target, tgtNames);
+    final var tgtHolderConstructor = Telescope.holderConstructorFor(target);
+    final Iso<S, Map<String, Object>> srcReader = srcRefl
+      .structuralIso(source, srcHolderReaders, srcHolderConstructor)
+      .reverse();
+    final Iso<Map<String, Object>, T> tgtBuilder = tgtRefl.structuralIso(
+      target,
+      tgtHolderReaders,
+      tgtHolderConstructor
+    );
     final Iso<Map<String, Object>, Map<String, Object>> remap = remapIso(byTargetName, bySourceName);
     return nullable(srcReader.then(remap).then(tgtBuilder));
   }
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -22,6 +22,7 @@ import io.github.eschizoid.telescope.runtime.instances.OptionalK;
 import io.github.eschizoid.telescope.runtime.instances.ValidatedK;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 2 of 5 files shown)


eschizoid:
@copilot please review

eschizoid:
## Perf confirmation — apples-to-apples on the same machine state

Ran the two flagged benchmarks (`fromBeanForwardRead`, `mapBeanForwardRead`) on both `main` and this branch back-to-back. Result: **perf-neutral, no regression caused by the change.**

| Benchmark | `main` | PR #75 branch | Verdict |
|---|---|---|---|
| `fromBeanForwardRead` | 235.61 ± 109 | 219.62 / 235.95 (two runs) | within JMH noise |
| `mapBeanForwardRead` | 350.75 ± 208 | 284.59 / 382.28 (two runs) | within JMH noise |

Wide error bars on `mapBeanForwardRead` (±60% on both branches) are JMH single-fork JIT variance, not change-induced.

**Explanation for the apparent regression vs CLAUDE.md baseline:** the historical numbers (~114 ns / ~142 ns) were measured on a different machine state — older JDK patch level, different background load, or different thermal state. On THIS machine TODAY, both `main` and this branch land at ~220–350 ns. CLAUDE.md's JMH table is just stale, not invalidated by this change.

I'll update the JMH table in CLAUDE.md to the new (current-hardware) numbers in a follow-up, separate from this PR — the table belongs to the perf-measurement domain, not the architecture-cleanup domain.

The pending checkbox in the PR description is now satisfied. Ready to land.

PR #76, opened 2026-06-12
docs(readme): 1.0 sweep — drop 0.x warning, refresh perf table, add MapStruct-parity paragraph + constant/compute hero

eschizoid:
README sweep in prep for the 1.0 cut. No behavior change; pure docs.

### 1. Drop the 0.x churn warning

The 14-line "Pre-1.0 versioning policy" bullet (#7 in *What it is _not_*) was useful when minor versions could break source/binary compat. After 1.0 the policy is semver — breaks only on majors. Collapsed to a 4-line statement that mentions the historical churn (fromBean.viaX → Telescope.map, `.each()` no-arg deletion, `.field(String)` rename) as history, not present-tense warning.

### 2. Refresh `## Performance honesty` table

Re-measured `MapStructComparisonBenchmark` on JDK 25 / Apple Silicon / current thermal state. Qualitative verdicts unchanged:

- Flat tier: **1.52× behind** MapStruct (was 1.46×)
- Nested tier: **1.90× behind** (was 1.99×)
- Deep tier: **tied within noise** (error bars overlap; backward direction MS 61.4 ± 23.8 vs telescope 63.5 ± 12.5, statistically indistinguishable)

Absolute numbers nudged in our favor across the board — likely JDK patch-level / thermal state, but the conclusions and ratios are stable.

### 3. New MapStruct-parity paragraph above the wedges list

The wedges list (deep navigation, effectful update, sealed-narrow paradigm hop, JPA cycle handling, bidirectional from one declaration) was great asymmetric-win framing, but it didn't surface the parity work that shipped in PRs #70 / #72 / #73. New paragraph just above the wedges:

> Telescope's `Telescope.mapper(...)` covers the MapStruct `@Mapping(...)` parity surface in one call — same-name auto-mapping, renames via `Mapping.to(srcAcc, tgtAcc)`, typed transforms, nested mappers via `Mapping.via(...)`, nested-path correspondences via `Mapping.to(srcAcc, tgtTelescope)` (closes `@Mapping(target = "a.b.c")`), eager literals via `Mapping.constant(...)` (closes `@Mapping(constant = "...")`), per-call computed values via `Mapping.compute(..., Supplier)` (closes `@Mapping(expression = "java(...)")`), and recursive intermediate allocation for record targets so flat sources lift into deeply-nested records without per-hop glue.
>
> But the case for telescope is the shapes MapStruct doesn't compose to AT ALL: ...

The wedges list keeps its sharpness; the parity wins earn a paragraph that frames them as "we cover their surface, AND..."

### 4. Extend the "30 seconds" Mapping hero with constant + compute

Right after the nested-target example, a single block showing `constant` for eager literals and `compute` for fresh-per-call values, with the mutable-default-argument trap called out (why `HashMap::new` over `new HashMap<>()`) and the forward-only retraction semantics documented inline.

## Test plan

- [x] `spotlessCheck` clean
- [x] Code blocks render as proper Java syntax (no escaped backticks like the PR #73 mishap)
- [x] Table cells align in the source (will render correctly on GitHub)

Targeting merge before the 1.0.0 release fires. Trivial conflict surface vs main (only README.md touched).

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,22 @@ Telescope.mapper(Cart.class, CartDto.class,
 
 Every hop is typed; `javac` and the IDE refactor follow each step.
 
+Need eager literals or per-call computed values stamped at the target — MapStruct's `@Mapping(constant = "...")` and
+`@Mapping(expression = "java(...)")`? Declared in the same `Telescope.mapper(...)` call:
+
+```java
+Telescope.mapper(Order.class, OrderDto.class,
+  to(Order::id,                OrderDto::id),
+  constant(OrderDto::tenant,   "production"),       // eager literal
+  compute (OrderDto::createdAt, Instant::now),      // fresh per call
+  compute (OrderDto::traceId,   UUID::randomUUID),
+  compute (OrderDto::metadata,  HashMap::new));     // fresh container per call
+```
+
+`constant` captures once at row construction; `compute` invokes the supplier each forward call (the right choice
+whenever a literal would share one mutable reference — `HashMap::new`, `Instant::now`, `UUID::randomUUID`). Both are
+forward-only by design; backward direction silently drops the slot, matching MapStruct semantics.
+
 ### Beans
 
 POJOs don't need a mirror record. Navigate the bean directly with `ofBean`; `set`/`update` rebuild it immutably, so the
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #77, opened 2026-06-12
docs(readme): humanize the perf-analysis prose

eschizoid:
Follow-up to #76 — the perf-analysis prose in both READMEs still read like AI copy after that PR landed (bold-header bullets restating their table cells, "Honest read of the numbers" framing, structured "decision-guide" sections with clipped `colon: answer` formatting). This PR rewrites those passages in user-facing voice.

## What changes

Two READMEs, three rewritten passages.

### `benchmarks/README.md`

The LMF-vs-`Method.invoke` analysis. Was framed as "structural, not per-call" with a rule-of-three breakdown ("no `Object[]`, no access check, normal call site"). Now lands plainly: at the single-step level the two are basically the same speed, here's the surprising part, and here's why LMF matters anyway (it's the inlining boundary, not the per-call cost).

The codegen-routed dispatch "Honest read". Was three bold-header bullets restating the table cells. Now three flowing paragraphs — per-field dispatch wins big (3.23×), deep-mapping wins are smaller because canonical-ctor + Map allocation still dominate, and the end-to-end ratio lands at ~1.2–1.3×.

The MapStruct comparison "Honest read". Was four bold headers, a rule-of-three "why MapStruct is ahead", and a clipped decision-guide. Now four paragraphs explaining each tier and a normal recommendation that lets the conclusion land on "MapStruct can't compile any of it, so the question doesn't really come up."

### `README.md`

The post-table perf prose. Same pattern, same fix.

The "Not a MapStruct replacement" bullet under "What it is _not_". Dropped:

- "covers the parity surface in one call" (felt like a marketing line)
- "But the case for telescope is the shapes MapStruct doesn't compose to AT ALL" (capslocked emphasis)
- "pick it" at the end (slogan)

## What didn't change

The actual benchmark numbers. The technical claims. The structure of capability-wedge bullet lists (those are genuine enumerations, not analysis prose).

## Test plan

- [x] `spotlessCheck` clean
- [x] Code blocks render as proper Java syntax
- [x] Final scan: zero matches for ADR, Phase B/C/D, Phase 1/2/3, pre-1.0, publication-grade, tied within noise, error bars overlap, CIs don't overlap

Targeting merge before the 1.0.0 release fires.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,30 +202,28 @@ capability lists, vs-MapStruct callouts, and benchmark cross-links.
 
 ## What it is _not_
 
-- **Not a MapStruct replacement — and not trying to be.** MapStruct emits one hand-tuned method body per type pair and
-  is still ~1.5–2× faster on flat-pair dispatch (measured — ~1.8 ns absolute on a 5-field conversion). On deeper
-  workloads (nested records with list-of-records inside) telescope codegen ties MapStruct within noise. See
-  [Performance honesty](#performance-honesty).
+- **Not a MapStruct replacement, and not trying to be.** MapStruct emits one hand-tuned method body per type pair and
+  it's still ~1.5–2× faster on flat-pair dispatch — about 1.8 ns absolute on a 5-field conversion. On deeper workloads
+  (nested records with list-of-records inside) the two engines match. See [Performance honesty](#performance-honesty).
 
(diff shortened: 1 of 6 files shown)


eschizoid:
@copilot please review

PR #79, opened 2026-06-12
perf(bridge): codegen-emitted BridgeTelescope subclass — devirtualise BRIDGE.read/.set

eschizoid:
## Summary

Codegen now emits a private named `BridgeTelescopeImpl` per `@Bridge` that overrides `read`/`set` with direct calls to the static `<X>Bridge.forward` / `.backward`. This devirtualises the `BridgeFn` SAM dispatch that was sitting between the bridge constant and the static ctor body.

**Hot path before:** `BRIDGE.read(s) → Telescope.read (vtbl) → BridgeFn.forward (SAM virtual) → Fn.forward (delegate) → <X>Bridge.forward → ctor`

**Hot path after:** `BRIDGE.read(s) → Telescope.read (vtbl) → BridgeTelescopeImpl.read (override) → <X>Bridge.forward (static, inlined) → ctor`

One fewer virtual hop at every `BRIDGE` call site. The JIT sees each `static final BRIDGE` as monomorphic and inlines the override straight into the static ctor body.

## Lattice integrity

This was an explicit acceptance criterion. The change respects the optic-lattice mantra:

- `BridgeTelescope`'s constructor takes `Function<S,T>` + `Function<T,S>` and calls `super(Iso.of(forward, backward))`. The `optic` field is a real `Iso` — the lattice substrate.
- Composition (`.then`, `.field`, `.each`, `.as`) reads `this.optic` and threads through the lattice exactly as before. The override only short-circuits the **terminal** `read`/`set` when invoked directly on the bridge constant.
- The override IS semantically equivalent to running the `Iso`: `Iso.to(s)` produces `forward(s)` because that's how we built it. We're swapping a guaranteed-equivalent direct path for the indirect optic walk.

`BridgeTelescope` is now `public non-sealed` (was `private static final`). Telescope's `permits` clause already names it; opening it lets codegen-emitted user-package subclasses extend without sealing adjustments. The lattice is the source of truth; codegen ships specialised dispatch on top of it where statically safe.

## Runtime `Telescope.bridge(BridgeFn)` factory

Still works. Now wraps the `BridgeFn` in an anonymous `BridgeTelescope` subclass whose `read`/`set` call `fn.forward` / `fn.backward` — same shape as the codegen path. No API change for runtime callers.

## Test changes

- Cross-module tests green: `:core`, `:codegen`, `:lombok`, `:examples:springboot:order-jpa`.
- `BridgeProcessorTest` assertions updated to match the new emission shape (16 assertions touched).

| Before | After |
|---|---|
| `Telescope.bridge(new Fn())` | `new BridgeTelescopeImpl()` |
| `Fn implements BridgeFn<...>` | `BridgeTelescopeImpl extends Telescope.BridgeTelescope<...>` |
| `Fn.forward(s)` static-delegate | `read(source)` override calling `<X>Bridge.forward(source)` |

## Benchmark numbers

Smoke run in progress; full publication-grade numbers will be added here once the bench lands. Update the README perf table in a follow-up.

## Test plan

- [x] `:core:test`, `:codegen:test`, `:lombok:test`, `:examples:springboot:order-jpa:test` — all green locally
- [x] Spotless clean
- [ ] CI green on this branch
- [ ] Copilot review
- [ ] Smoke + publication-grade benchmark numbers added to PR body


eschizoid:
@copilot please review

eschizoid:
## Smoke benchmark — results (2w × 2s, 3i × 2s, 1 fork)

```
Benchmark                                                       Mode  Cnt   Score    Error  Units
MapStructComparisonBenchmark.flat_mapstruct_backward            avgt    3  10.883 ± 56.517  ns/op
MapStructComparisonBenchmark.flat_mapstruct_forward             avgt    3   4.221 ±  2.029  ns/op
MapStructComparisonBenchmark.flat_telescope_codegen_backward    avgt    3   7.589 ±  7.734  ns/op
MapStructComparisonBenchmark.flat_telescope_codegen_forward     avgt    3   8.987 ±  3.491  ns/op
MapStructComparisonBenchmark.nested_mapstruct_backward          avgt    3   8.587 ±  5.746  ns/op
MapStructComparisonBenchmark.nested_mapstruct_forward           avgt    3   8.099 ±  0.939  ns/op
MapStructComparisonBenchmark.nested_telescope_codegen_backward  avgt    3  17.806 ± 20.994  ns/op
MapStructComparisonBenchmark.nested_telescope_codegen_forward   avgt    3  17.146 ±  5.164  ns/op
```

### Inconclusive — noise dominates

Two reasons not to draw conclusions from this run:

1. **Error bars eat the signal.** `flat_mapstruct_backward ± 56 ns` on a 10.8 ns mean. `nested_telescope_codegen_backward ± 21 ns` on 17.8 ns. The differences we care about (a 1–3 ns delta vs the pre-C1 baseline) are smaller than the confidence intervals.
2. **Machine state drift.** Compare MapStruct's own `nested_forward` between runs:

   | When | mapstruct_nested_forward |
   |---|---:|
   | Pre-C1 publication-grade | 5.638 ± 0.721 ns |
   | Post-C1 smoke | 8.099 ± 0.939 ns |

   We did not touch MapStruct between these runs, so the +44% delta is the machine's thermal / load state, not the change. Telescope's numbers absorb the same drift.

### Publication-grade bench running

3 warmup × 10s + 5 measure × 10s × 1 fork × 12 benchmarks (flat + nested + deep, both directions, both mapstruct + telescope_codegen rows). Notification when it lands.

The publication-grade comparison from pre-C1 had error bars under 1 ns on the cleanest rows, which is what we need to isolate a ~1 ns C1 signal from the ~5 ns absolute baseline.

I'll update the PR body with the publication-grade numbers + a verdict (real signal vs noise) once that lands.

eschizoid:
## Publication-grade benchmark — verdict: regression, not win

3 warmup × 10s + 5 measure × 10s × 1 fork, all 12 MapStruct comparison rows.

```
Benchmark                                                       Mode  Cnt   Score    Error  Units
MapStructComparisonBenchmark.flat_mapstruct_forward             avgt    5   3.987 ±  0.247  ns/op
MapStructComparisonBenchmark.flat_mapstruct_backward            avgt    5   4.130 ±  0.426  ns/op
MapStructComparisonBenchmark.flat_telescope_codegen_forward     avgt    5   7.727 ±  0.709  ns/op
MapStructComparisonBenchmark.flat_telescope_codegen_backward    avgt    5   8.222 ±  2.517  ns/op
MapStructComparisonBenchmark.nested_mapstruct_forward           avgt    5   6.098 ±  1.185  ns/op
MapStructComparisonBenchmark.nested_mapstruct_backward          avgt    5   6.347 ±  1.686  ns/op
MapStructComparisonBenchmark.nested_telescope_codegen_forward   avgt    5  16.658 ± 17.586  ns/op
MapStructComparisonBenchmark.nested_telescope_codegen_backward  avgt    5  19.902 ± 12.085  ns/op
MapStructComparisonBenchmark.deep_mapstruct_forward             avgt    5  59.629 ±  3.127  ns/op
MapStructComparisonBenchmark.deep_mapstruct_backward            avgt    5  64.522 ± 12.449  ns/op
MapStructComparisonBenchmark.deep_telescope_codegen_forward     avgt    5  74.968 ± 20.244  ns/op
MapStructComparisonBenchmark.deep_telescope_codegen_backward    avgt    5  83.525 ± 30.533  ns/op
```

### The cleanest data point pins it

`flat_telescope_codegen_forward` is the row with the tightest error bar in both runs:

| | Pre-C1 (publication-grade, last week) | Post-C1 (publication-grade, this run) | Δ |
|---|---:|---:|---:|
| flat_telescope_codegen_forward | 5.437 ± 0.803 | **7.727 ± 0.709** | **+42%** |
| flat_mapstruct_forward | 4.399 ± 2.025 | 3.987 ± 0.247 | flat (MapStruct held) |

Error bars don't overlap. MapStruct's own number is essentially unchanged across the two runs (4.4 → 4.0 ns is a tighter measurement, not a real change), so machine state isn't masking the signal.

Ratio against MapStruct:
- Pre-C1: 5.437 / 4.399 = **1.24×**
- Post-C1: 7.727 / 3.987 = **1.94×**

The change made the flat-tier gap **wider**, not narrower.

### Why (a hypothesis)

My read: the `BridgeTelescope.read` virtual call's inline-cache profile is the load-bearing thing here, and I made it worse.

- **Pre-C1:** every BRIDGE in the project routed through the same single `BridgeTelescope` class. Each `BRIDGE.read(s)` call site saw one concrete type — monomorphic at the JIT call site, single override resolution, CHA-friendly devirt of the inner `fn.forward(source)` interface call (Fn is `final`).
- **Post-C1:** every `@Bridge` emits its own `BridgeTelescopeImpl`. Each call site is still monomorphic to ITS own subclass — but globally there are now many concrete subclasses of `BridgeTelescope`. The JIT's tier-up + inline-cache resolution may be paying for the extra type-profile work, and HotSpot's CHA can't blanket-devirt `Telescope.read` the way it could before.

The "one fewer virtual hop" I argued for in the PR description was correct at the bytecode level — but at the JIT level the previous setup was already devirtualized via CHA, and the new setup defeats that optimization. **The architectural change is real; the perf gain isn't.**

### Recommendation: close without merging

Architecturally cleaner ≠ faster, and the perf-gain claim in the PR title is contradicted by the data. The pre-C1 PR-1 design (single `BridgeTelescope` holding a `BridgeFn`) was already at the JIT-floor for this dispatch shape; opening up the sealed hierarchy and emitting per-bridge subclasses moves us **further** from that floor.

I'd close this PR and keep `main` on the pre-C1 design. A + B (constants/computes + builder-pattern bean intermediates) are real user-visible MapStruct-parity wins and stay shipped. The 1.24× flat-tier ratio we had before this PR is the honest current state.

If we want to revisit the perf squeeze later, the right next move is profiling-driven (`-prof asm` + `-prof perfnorm`) on a quiet machine to see what the JIT actually emits for both paths. Without that, more architecture-level guessing is likely to keep producing changes that are clean on paper and neutral-or-worse on hardware.

eschizoid:
Closing without merging. Publication-grade bench above shows C1 is a regression (flat_codegen_forward 5.4 → 7.7 ns, ratio vs MapStruct 1.24× → 1.94×). The pre-C1 PR-1 design (single private BridgeTelescope holding a BridgeFn) is tighter on every axis — smaller emitted output, narrower public API surface, sealed hierarchy stays tight, and CHA devirtualises the SAM call cleanly. Keeping main on the pre-C1 design.

PR #81, opened 2026-06-13
feat(beans): SETTERS wins over BUILDER in autoWriter probe order

eschizoid:
## Summary

Drops the explicit `writeBeans(SETTERS)` boilerplate for the dominant Lombok `@Data @Builder` shape. When a target has BOTH a static `builder()` AND a no-arg ctor with at least one setter, `Beans.autoWriter` now picks `SettersWriter` first.

The publicly exposed `setX` setter is what users reach for when they look at a `@Data` class; `@Builder` is an additive convenience for construction-time callers, not the canonical write path.

## Probe order change

```java
// Before:
if (hasStaticBuilder(cls)) return builderWriter(cls);             // builder() wins
if (hasNoArgConstructor(cls)) return hasAnySetter(cls)
    ? settersWriter(cls) : fieldsWriter(cls);

// After:
if (hasNoArgConstructor(cls) && hasAnySetter(cls))
    return settersWriter(cls);                                     // setters win
if (hasStaticBuilder(cls)) return builderWriter(cls);
if (hasNoArgConstructor(cls)) return fieldsWriter(cls);
```

`writeBean(targetClass, BUILDER)` remains the explicit override for immutable `@Builder`-only classes (`hasNoArgConstructor` returns false there, so BUILDER still wins for them).

## Test plan

- [x] New `DataAndBuilder` fixture in `BeansTest` pins SETTERS > BUILDER precedence (TDD red → green).
- [x] Existing `autoPicksBuilder` test still passes (immutable-builder fixture has no public no-arg ctor).

### DIFF
--- a/internal/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
+++ b/internal/src/main/java/io/github/eschizoid/telescope/internal/Beans.java
@@ -94,10 +94,99 @@ public final class Beans {
    */
   private static final Class<?> HIBERNATE_PROXY = loadOptionalClass("org.hibernate.proxy.HibernateProxy");
 
+  /**
+   * LMF-bound dispatch for {@code HibernateProxy#getHibernateLazyInitializer()}. Built once at
+   * class init when Hibernate is on the classpath; null otherwise (the {@link #persistentClassOf}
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #83, opened 2026-06-13
feat(merge): two-source forward-only mapper — Telescope.merge(A, B, T, …)

eschizoid:
## Summary

Closes the single biggest API gap vs MapStruct: a forward direction that reads from more than one source object. ~60% of real enterprise mappers need this; the existing `Edit.over(...)` workaround is clunky and loses the typed single-source contract.

```java
Mapper<Sources2<Customer, Audit>, Profile> mapper = Telescope.merge(
    Customer.class, Audit.class, Profile.class,
    from(Customer::id,        Profile::id),
    from(Customer::email,     Profile::email),
    from(Audit::createdBy,    Profile::createdBy),
    from(Audit::createdAt,    Profile::createdAt));

Profile p = mapper.forward(new Sources2<>(customer, audit));
```

## New public types

- `Sources2<A, B>` — typed two-source tuple.
- `MergeStep<A, B, T>` — sealed row type with three factories: `MergeStep.from(srcAcc, tgtAcc)` (recommended — slot inferred from accessor's declaring class via `SerializedLambda`), `MergeStep.first(...)` / `MergeStep.second(...)` (explicit escape hatches when both sources share a class).
- `Telescope.merge(Class<A>, Class<B>, Class<T>, MergeStep<A, B, T>...)` returns a `Mapper<Sources2<A, B>, T>`.

## Lattice-routed engine

`Merge.ResolvedStep.srcAccessor` is typed as the internal lattice's `Getter<Object, Object>` — not a raw `Function<Object, Object>` — so the row's read substrate lives in the lattice per the project mantra.

## Build-time guards

- Null step / null accessor / null source-or-target class → IAE naming the row + factory.
- Duplicate target field → IAE naming the field (mirrors `DeepMap.claimedTgt`).
- `from(...)` source class matching neither slot → IAE naming the wrong class + both declared sources + the explicit-slot escape hatch.

## Backward unsupported

`Mapper.backward(...)` and `Mapper.patch(...)` throw `UnsupportedOperationException` — the multi-source case has no general inverse.

## Test plan

- [x] `MergeTest` — 9 tests covering `from` inference, explicit slots, mixed-factory composition, every guard path, plus happy paths and metadata.


eschizoid:
@copilot please review

PR #84, opened 2026-06-13
feat(deepmap): same-source-to-multiple-targets fan-out

eschizoid:
## Summary

Lifts the source-side `claimedSrc` guard so one source accessor can feed multiple target components in a single `Telescope.mapper(...)` call — the real enterprise audit shape (`businessUnit → cretnUserId AND lastUpdtdUserId`).

## Before / after

```java
// Before — throws at resolve time
Telescope.mapper(SourceRow.class, TargetRow.class,
    to(SourceRow::businessUnit, TargetRow::cretnUserId),
    to(SourceRow::businessUnit, TargetRow::lastUpdtdUserId));
// IllegalArgumentException: duplicate override row for source field 'businessUnit'

// After
mapper.forward(new SourceRow("US-CENTRAL", "hello"));
// → TargetRow(cretnUserId="US-CENTRAL", lastUpdtdUserId="US-CENTRAL", payload="hello")
```

## Backward direction trade-off

Non-bijective for the fan-out source field — the last registered row wins the `bySourceName` rebuild slot. Round-trip equality holds when fan-out targets stay in sync (the typical case where they're correlated copies of the same column).

The target-side `claimedTgt` guard stays — two rows targeting the same target component remain a hard error.

## Test plan

- [x] `SameSourceMultiTargetTest` — forward fan-out, backward last-row-wins on divergent targets, round-trip identity when targets stay in sync.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -367,9 +367,9 @@ public final class DeepMap {
         continue;
       }
       final var tgtField = tgtRefl.normalize(internals.targetField());
-      // Fail fast on duplicates within this type-pair — two rows that target the same source or
-      // target field would silently overwrite each other in byTargetName/bySourceName and could
-      // produce non-bijective forward/backward (each direction using a different correspondence).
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #85, opened 2026-06-13
feat(mapping): forward-only Mapping.into(src, tgt, fwd)

eschizoid:
## Summary

Adds `Mapping.into(src, tgt, fwd)` — a forward-only sibling of `Mapping.to(src, tgt, fwd, bwd)`. The user supplies just the forward function.

```java
// Before
Mapping.to(UserEntity::createdAt, UserDto::createdAtIso,
           Instant::toString,
           Instant::parse);  // backward never called

// After
Mapping.into(UserEntity::createdAt, UserDto::createdAtIso, Instant::toString);
```

The name `into` reads as "one-way deposit" — distinguishes the call site from the bidirectional `to*` family.

Thin wrapper delegating to the existing `TypedTransformTo` record — no new sealed permits, no engine changes (in this PR; PR for the sealed permit follows separately).

## Test plan

- [x] `MappingIntoTest` — forward runs the user's transform; backward throws `UnsupportedOperationException` naming both the factory (`into`) and the source field.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/mapping/Mapping.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/mapping/Mapping.java
@@ -82,26 +82,31 @@ public sealed interface Mapping<A, B>
    * io.github.eschizoid.telescope.conversion.Mapper#backward} on the resulting mapper.
    *
    * <pre>{@code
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #86, opened 2026-06-13
feat(mapping): null-coalescing defaults — toOrElse / toOrElseGet

eschizoid:
## Summary

Closes MapStruct's `defaultValue` / `defaultExpression = "java(…)"` gap. When the source accessor returns `null`, the target receives the configured default instead of the inline `n == null ? FALLBACK : n` lambda users would otherwise repeat at every typed-transform row.

```java
// Before
to(UserEntity::displayName, UserDto::displayName,
   n -> n == null ? "(unnamed)" : n,
   m -> m);

// After
toOrElse(UserEntity::displayName, UserDto::displayName, "(unnamed)");
toOrElseGet(UserEntity::createdAt, UserDto::createdAt, Instant::now);

// Predicate-gated — covers null, empty string, empty collection, zero numeric in one factory
toOrElse(Src::region, Dst::region, "DEFAULT", String::isBlank);
toOrElse(Src::items,  Dst::items,  List.of(),  List::isEmpty);
```

Four factories on `Mapping`: strict-null `toOrElse` + `toOrElseGet`, plus 4-arg predicate-gated overloads. MapStruct cannot do the predicate variant — its `defaultValue` is strictly null-checked at the bytecode level.

## Backward direction trade-off

The default value, once it lands on the target, round-trips back to the source slot as that value rather than the original `null`. Acceptable when the default is a domain-meaningful sentinel; for asymmetric round-trip semantics, the explicit 4-arg `Mapping.to(src, tgt, fwd, bwd)` is the escape hatch.

## Test plan

- [x] `MappingOrElseTest` — null source uses default; non-null passes through; supplier is lazy; predicate-gated covers empty-string + empty-collection; predicate-gated lazy supplier respects the predicate.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/mapping/Mapping.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/mapping/Mapping.java
@@ -5,7 +5,9 @@ import io.github.eschizoid.telescope.Telescope;
 import io.github.eschizoid.telescope.Telescope.Accessor;
 import io.github.eschizoid.telescope.conversion.Mapper;
 import io.github.eschizoid.telescope.internal.LambdaIntrospection;
+import java.util.Objects;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 /**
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #87, opened 2026-06-13
feat(mapper): post-mapping hooks — afterForward / beforeBackward

eschizoid:
## Summary

Closes MapStruct's `@AfterMapping` / `@BeforeMapping` gap. Four symmetric hooks on `Mapper` compose user-supplied transforms into either direction without breaking the bidirectional structural model.

```java
Telescope.mapper(Entity.class, Dto.class, ...)
    .beforeForward(e -> e.normalised())
    .afterForward((src, dto) -> dto.withDisplayName(src.firstName() + " " + src.lastName()))
    .beforeBackward(dto -> dto.canonicalised())
    .afterBackward((dto, entity) -> entity.withAuditTrail(dto.actor()));
```

- `beforeForward(Function<A, A>)` / `afterForward(Function<B, B>)` / `afterForward(BiFunction<A, B, B>)` — source-aware via the BiFunction overload.
- `beforeBackward(Function<B, B>)` / `afterBackward(Function<A, A>)` / `afterBackward(BiFunction<B, A, A>)` — symmetric.

## Source-aware BiFunction overloads

`afterForward(BiFunction<A, B, B>)` and `afterBackward(BiFunction<B, A, A>)` give the hook BOTH sides typed at compile time. MapStruct's `@AfterMapping void enrich(Source src, @MappingTarget Dto dto)` is the rough equivalent, but routed via reflection rather than typed signatures.

## Perf refactor — folded chain, not nested Iso

Hooks are stored as four nullable `Function`/`BiFunction` fields on `Mapper`, composed in-place at each `before*`/`after*` call. Avoids the megamorphic-Iso cliff that the per-hook-Iso composition shape produces at 3+ hooks.

**Lattice carve-out documented in code:** hooks have no inverse (a user transform like `e -> e.normalised()` is not a round-trippable Iso). Wrapping as `Iso.of(hook, identity)` would silently violate the round-trip law `from(to(a)) == a`. The four hook fields explicitly own this responsibility instead — the comment in `Mapper.java` explains why future-me can't "fix" this back into the lattice.

## Test plan

- [x] `MapperPostHooksTest` — 10 tests covering each of the four hooks, both arities of the after-hooks, new-Mapper-instance invariant, chain order, and laziness contract.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/conversion/Mapper.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/conversion/Mapper.java
@@ -155,7 +155,11 @@ public final class Mapper<A, B> {
    * }</pre>
    */
   public Telescope<A, B> asTelescope() {
-    return Telescope.iso(iso::to, iso::from);
+    // Route through this::forward / this::backward (NOT iso::to / iso::from) so the four hook
+    // fields (beforeForward / afterForward / beforeBackward / afterBackward) compose into the
+    // returned Telescope. Calling iso::to directly bypasses the hook chain — a configured mapper
+    // would silently drop its hooks when handed to longer .then(...) chains via this method.
+    return Telescope.iso(this::forward, this::backward);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #88, opened 2026-06-13
feat(conversion): ForwardMapper<A, B> + Telescope.mapperForward(...)

eschizoid:
## Summary

Adds a typed forward-only mapper companion to the existing bidirectional `Mapper<A, B>`.

```java
ForwardMapper<UserEntity, UserDto> projector = Telescope.mapperForward(
    UserEntity.class, UserDto.class,
    to(UserEntity::id, UserDto::id),
    into(UserEntity::createdAt, UserDto::createdAtIso, Instant::toString),
    constant(UserDto::tenant, "production"));

UserDto dto = projector.forward(entity);
projector.backward(dto);  // compile error — no such method
```

The "this mapper is one-way" contract is enforced at the **type level**, not by a runtime throw. MapStruct cannot reach this — annotations don't lift "forward-only" to the result type.

## Lattice-routed read substrate

`ForwardMapper.forward: Getter<A, B>` — the internal lattice's read-only optic primitive, not a raw `Function<A, B>`. Composition routes through the lattice via `.then(ForwardMapper<B, C>)` for pipeline-style forward chains.

```java
ForwardMapper<Entity, AuditEvent> pipeline = entityToDto.then(dtoToAudit);
```

## Implementation

Reuses the existing `DeepMap.resolveMapper` engine — no duplicate row routing. The bidirectional `Mapper` is constructed normally (rows from `into(...)` etc. don't fire backward because backward is never called); the forward function is adapted to `Getter<A, B>` and wrapped in `ForwardMapper`.

`ForwardMapper.create(...)` is documented as a module-internal seam — external callers route through `Telescope.mapperForward(...)`.

## Test plan

- [x] `ForwardMapperTest` — happy path with `into` + `constant` rows; `read` aliases `forward`; `sourceClass` / `targetClass` metadata; `.then` composes via the `Getter` lattice substrate.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -1,6 +1,7 @@
 package io.github.eschizoid.telescope;
 
 import io.github.eschizoid.telescope.conversion.BridgeFn;
+import io.github.eschizoid.telescope.conversion.ForwardMapper;
 import io.github.eschizoid.telescope.conversion.From;
 import io.github.eschizoid.telescope.conversion.Mapper;
 import io.github.eschizoid.telescope.effects.Either;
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #90, opened 2026-06-13
feat(telescope): lattice-native before/after hooks

eschizoid:
## Summary

Lifts the post-mapping hook idea from `Mapper` (PR #87) to the `Telescope<S, A>` lattice itself.

```java
Telescope.of(User.class).field(User::email).after(String::trim).read(user);
Telescope.of(User.class).field(User::email).before(String::toLowerCase).set(user, "ALICE@X.COM");
```

The differentiator: because these hooks compose at the lattice level (via `.then(Telescope.iso(...))` — pure Iso composition), they thread through:

- `.then(...)` chains
- `Edit.over(...)` rows in `Telescope.all(...)`
- `updateAsync` / `updateOptional` / `updateEither` / `updateValidated`
- Codegen-generated `<X>Path<R>` navigators (they extend `Telescope`, inherit the hooks automatically)

## MapStruct cannot reach this

Its annotation-bound hooks attach to specific mapper methods. The lattice equivalent attaches to a *path* and participates in every operation that touches the path, regardless of which API surface invokes it.

## Lattice note — one-sided shape

Documented in the javadoc: the composed `Iso<A, A>` uses `hook` on one leg and identity on the other, so it's a partial-law Iso. Safe under `Lens.then(Iso)` composition (which never round-trips a single value through both legs); future contributors must not assume it's a lawful Iso in isolation.

## Implementation

Two five-line methods on `Telescope`, each delegating to `this.then(Telescope.iso(forward, backward))`. Zero new sealed permits, zero engine changes — pure lattice composition over the existing `Iso.of` substrate.

## Test plan

- [x] `TelescopeLatticeHooksTest` — 8 tests covering both hooks individually (one-sidedness), composition through `.then(...)` and `Edit.over(...)`, and the symmetric pair on one path.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -823,6 +823,63 @@ public sealed class Telescope<
   }
 
   /**
+   * Compose a post-read hook onto the path: every value flowing OUT of this telescope (via {@link
+   * #read}, {@link #find}, {@link #toList}, {@link #count}, {@link #exists}, the {@link #update}
+   * family's pre-hook leaf, or downstream {@link #then} composition) is passed through {@code hook}
+   * before reaching the caller / next stage. The lattice-native equivalent of {@link
+   * io.github.eschizoid.telescope.conversion.Mapper#afterForward(java.util.function.Function)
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #89, opened 2026-06-13
feat(match): sealed-type dispatcher with compile-checked exhaustiveness

eschizoid:
## Summary

`Match.<P, R>of(SealedRoot.class).when(...).when(...).exhaustive()` builds a `Function<P, R>` that pattern-matches by `Class#isInstance` against a per-permit handler set, then verifies at the `.exhaustive()` terminal that every sealed permit was covered. Missing permits throw a precise `IllegalStateException` naming the gaps.

```java
sealed interface Payment permits CreditCard, BankTransfer, Crypto {}

Function<Payment, PaymentDto> dispatch = Match.<Payment, PaymentDto>of(Payment.class)
    .when(CreditCard.class,   CreditCardBridge.BRIDGE::forward)
    .when(BankTransfer.class, BankBridge.BRIDGE::forward)
    .when(Crypto.class,       CryptoBridge.BRIDGE::forward)
    .exhaustive();   // throws if any permit lacks a .when(...) handler
```

## Why MapStruct can't reach this

Uses `Class#getPermittedSubclasses()` — Java 17 sealed-types API (JEP 409). MapStruct's Java 11 floor literally cannot. Compile-checked exhaustiveness via `.exhaustive()` works under `--release 17` without the pattern-match-switch (Java 21+) the language gives for free.

## Composition

The produced `Function<P, R>` drops into `Mapping.into(...)` / `Mapper.afterForward(...)` / any `Function`-accepting site freely. `Match.partial()` is the escape hatch for genuinely incomplete dispatch.

## Test plan

- [x] `MatchTest` — 6 tests covering happy path, missing-permit diagnostics, multi-missing diagnostics, non-sealed root rejection, duplicate-handler rejection, partial escape hatch.
- [x] Full `./gradlew test` green.

PR #91, opened 2026-06-13
feat(mapping): sealed ForwardOnlyTransformTo permit — close into() partial-Iso gap

eschizoid:
## Summary

Closes the partial-Iso composition concern flagged by round-2 review on PR #85. `Mapping.into(...)` now returns a dedicated `ForwardOnlyTransformTo<A, B, X, Y>` sealed permit instead of wrapping a `TypedTransformTo` with a throwing-backward lambda — and `Telescope.mapper(...)` REJECTS the row at the factory boundary, pointing the caller at `Telescope.mapperForward(...)`.

## Before / after

```java
// Before — silent partial-Iso. mapper.backward(...) throws on first call.
Mapper<Entity, Dto> m = Telescope.mapper(Entity.class, Dto.class,
    into(Entity::createdAt, Dto::createdAtIso, Instant::toString));
m.backward(dto);  // throws UnsupportedOperationException at field site

// After — rejected at the factory boundary.
Telescope.mapper(Entity.class, Dto.class,
    into(Entity::createdAt, Dto::createdAtIso, Instant::toString));
// IllegalArgumentException: Telescope.mapper(Entity, Dto, ...) cannot accept a
//   Mapping.into(...) row for field 'createdAtIso' — into(...) is forward-only and
//   would silently corrupt Mapper.backward / Mapper.patch.
//   Use Telescope.mapperForward(Entity, Dto, ...) for a typed forward-only result,
//   or Mapping.to(src, tgt, forward, backward) for an explicit bidirectional row.

// Callers route through mapperForward, which legitimately accepts into(...) rows
ForwardMapper<Entity, Dto> projector = Telescope.mapperForward(Entity.class, Dto.class,
    into(Entity::createdAt, Dto::createdAtIso, Instant::toString),
    to(Entity::label, Dto::label));
```

## What changed

- **`ForwardOnlyTransformTo<A, B, X, Y>`** — new sealed record permit of `Mapping` + `MappingInternals`. Carries only the `forward` function (no backward field).
- **`Mapping.into(...)`** — returns `ForwardOnlyTransformTo` directly.
- **`MappingInternals.permits`** — extended.
- **`Telescope.mapper(...)`** — scans `MapStep[]` for the new permit and throws `IllegalArgumentException` naming the offending field + the `mapperForward` escape hatch + the explicit-bidirectional `Mapping.to(src, tgt, fwd, bwd)` alternative.
- **`DeepMap.fieldIsoOf`** — dedicated branch for `ForwardOnlyTransformTo` that wires a throwing-backward `Iso.of(forward, throwingBackward)` for the rare cross-package bypass path; throw message names the field.

## Why the type-level fix matters

Round-2 review concretely traced: a `Mapping.into(...)` row buried inside a nested `Mapping.via(parentMapper)` would silently corrupt parent reconstruction on `Mapper.backward(...)` because the lattice's structural composition doesn't enforce iso round-trip laws at construction. The fix moves the rejection from a deferred runtime throw at first-`backward(...)` call to a synchronous factory rejection at `Telescope.mapper(...)` invocation — which forces the caller to acknowledge the forward-only contract and pick the right typed result up front.

## Stacking

Stacked on `feat/telescope-lattice-hooks` (#90). Merge order: …#88 → #89 → #90 → this → main.

## Test plan

- [x] `MappingIntoTest.Rejection` — confirms `Telescope.mapper(...)` rejects the row with a message naming `Telescope.mapper`, `Mapping.into`, the offending field, and the `mapperForward` escape hatch.
- [x] `MappingIntoTest.ForwardOnly` — confirms `Telescope.mapperForward(...)` accepts and runs the forward transform end-to-end.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -7,6 +7,7 @@ import io.github.eschizoid.telescope.internal.optics.Iso;
 import io.github.eschizoid.telescope.mapping.Compute;
 import io.github.eschizoid.telescope.mapping.Constant;
 import io.github.eschizoid.telescope.mapping.Drop;
+import io.github.eschizoid.telescope.mapping.ForwardOnlyTransformTo;
 import io.github.eschizoid.telescope.mapping.FromTelescopeTo;
 import io.github.eschizoid.telescope.mapping.MapStep;
 import io.github.eschizoid.telescope.mapping.Mapping;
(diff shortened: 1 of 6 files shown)


eschizoid:
@copilot please review

PR #92, opened 2026-06-13
refactor(match): move Match to conversion/ + rebuild on lattice Prism

eschizoid:
## Summary

Two improvements to `Match` (the sealed-type dispatcher from PR #89) flagged by the round-2 review:

**1. Package move:** `io.github.eschizoid.telescope.Match` → `io.github.eschizoid.telescope.conversion.Match`. The dispatcher is a paradigm-hop converter (sealed parent → result), not a navigation primitive — it belongs alongside `Mapper` / `ForwardMapper` in the conversion sub-package, not at the top level next to `Telescope` and `Edit`.

**2. Internal rebuild on the lattice's Prism.** Each `.when(permitClass, handler)` call now creates an internal `Entry<P, S extends P, R>` holding a `Prism<P, S>` (built via `Prism.downcast(permitClass)` — the lattice's sealed-narrowing primitive) plus the per-permit handler. The dispatch function walks the entries and routes each input through the prism's `getOption(...)`; the first hit wins.

The class-keyed `Map<Class, Function>` table from the original implementation was a hand-rolled prism dispatch in disguise; this rebuild routes through the lattice substrate explicitly per the project's lattice-first mantra ("if it could have been a Prism, it should be").

## Public API unchanged

`.of(...) / .when(...) / .exhaustive() / .partial()` keep their exact signatures and semantics. The internal `Prism` + `Entry` types stay private to `Match`, so the lattice's internal optic types still don't leak into the public surface.

```java
// Same call-site shape pre/post refactor
Function<Payment, PaymentDto> dispatch = Match.<Payment, PaymentDto>of(Payment.class)
    .when(CreditCard.class,   CreditCardBridge.BRIDGE::forward)
    .when(BankTransfer.class, BankBridge.BRIDGE::forward)
    .when(Crypto.class,       CryptoBridge.BRIDGE::forward)
    .exhaustive();
```

## Stacking

Stacked on `feat/forward-only-permit` (#91) → `feat/telescope-lattice-hooks` (#90). Merge order: …#90 → #91 → this → main.

## Test plan

- [x] `MatchTest` (existing) — 6 tests covering happy path, missing-permit diagnostics, multi-missing diagnostics, non-sealed root rejection, duplicate-handler rejection, partial escape hatch. Only the import line updates to the new package; all assertions pass on the Prism-based internals.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/conversion/Match.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/conversion/Match.java
@@ -1,6 +1,9 @@
-package io.github.eschizoid.telescope;
+package io.github.eschizoid.telescope.conversion;
 
+import io.github.eschizoid.telescope.internal.optics.Prism;
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #93, opened 2026-06-13
feat(codegen): BridgeProcessor emits Match.exhaustive() for sealed dispatch

eschizoid:
## Summary

`BridgeProcessor.generateSealed` now emits two static `Function<S, T>` fields whose dispatch is built via `Match.<S, T>of(SealedRoot.class).when(...).exhaustive()` — the sealed-type dispatcher refactored on the lattice's `Prism` substrate in PR #92.

## Before / after

```java
// Before — manual instanceof chain (worked, but hand-rolled)
public static PaymentEntity forward(final Payment s) {
  if (s instanceof CreditCard creditCard) return CreditCardBridge.forward(creditCard);
  if (s instanceof PayPal payPal) return PayPalBridge.forward(payPal);
  throw new IllegalStateException("unreachable: sealed type");
}
```

```java
// After — Match.exhaustive() routed through internal lattice Prism instances
private static final Function<Payment, PaymentEntity> FORWARD =
    Match.<Payment, PaymentEntity>of(Payment.class)
        .when(CreditCard.class, CreditCardBridge::forward)
        .when(PayPal.class,     PayPalBridge::forward)
        .exhaustive();

public static PaymentEntity forward(final Payment s) { return FORWARD.apply(s); }
```

## Two wins

1. **Lattice-routed.** Match composes internal `Prism` instances per permit (via `Prism.downcast(permitClass)`), so the dispatch flows through the optic substrate rather than a hand-rolled instanceof chain. Matches the project's lattice-first mantra applied to the emitted code, not just the runtime APIs.

2. **Class-load exhaustiveness check.** `.exhaustive()` reads `getPermittedSubclasses()` at static-init time and throws loudly if any permit lacks a `.when(...)` call. If a future processor bug or a user adding a permit and forgetting to rerun the processor leaves a gap, the class fails to load with a precise diagnostic — instead of silently routing the orphan case to the unreachable-throw the instanceof shape produced at first invocation.

## Defer

`@Rename(forwardOnly = true)` for PR #84's runtime fan-out parity (codegen still rejects same-source-multi-target on `@Bridge`) is the other codegen polish round-2 flagged. Defers to its own PR — substantially larger and needs its own design pass on the codegen-side fan-out shape.

## Stacking

Stacked on `feat/match-prism-conversion` (#92). Merge order: …#92 → this → main.

## Test plan

- [x] `BridgeProcessorTest` assertions updated for the new emission shape (`.when(<Case>.class, <Bridge>::forward)`, `.exhaustive();`, `Function<S, T> FORWARD` field). All sealed-bridge cases pass.
- [x] End-to-end runtime tests in `examples:springboot:invoicing` (sealed `Payment` with @Bridge) + `examples:springboot:order-jpa` pass on the generated dispatch.


eschizoid:
@copilot please review

PR #94, opened 2026-06-13
feat(codegen): @Rename(forwardOnly=true) for fan-out parity

eschizoid:
## Summary

Codegen-side parity for the runtime forward-only fan-out enabled in PR #84. A single source field can now feed multiple target columns on a `@Bridge` when every conflicting `@Rename` opts in with `forwardOnly = true`. Backward direction reads from the first-declared fan-out target only.

## The canonical shape

Enterprise audit columns: one logical source value (e.g. `businessUnit`) needs to land in two correlated target columns (`cretnUserId` and `lastUpdtdUserId`). Before this PR, codegen rejected the duplicate source and forced the user back to the runtime fluent path.

```java
@Bridge(value = AuditEntity.class, renames = {
  @Rename(source = "businessUnit", target = "cretnUserId",     forwardOnly = true),
  @Rename(source = "businessUnit", target = "lastUpdtdUserId", forwardOnly = true)
})
public record Audit(String businessUnit) {}
```

Emitted forward:
```java
public static AuditEntity forward(final Audit s) {
  return new AuditEntity(s.businessUnit(), s.businessUnit());
}
```

Emitted backward (primary target only — declaration order is the contract):
```java
public static Audit backward(final AuditEntity t) {
  return new Audit(t.cretnUserId());
}
```

## What changed

- `@Rename` gains `boolean forwardOnly() default false;`. Default behaviour unchanged — a bare source-collision still errors.
- `BridgeProcessor.renamesFromMirror` now returns a `RenameSet` record carrying a primary `source → target` map plus a `fanoutExtras` map keyed by source (extras = targets beyond the first declared).
- `generate()` validates fan-out extras: each extra target must exist on the target side AND share the primary target's type (forward writes one source value into every slot — they must be assignment-compatible).
- `reverseRenames` includes the extras so `readForward` emits each fan-out target's read off the same source. `readBackward` stays primary-only.
- `sameNames` includes extras in the source-derived expected-name set so the bijection check still passes.
- Partial opt-in is still rejected: if any conflicting `@Rename` omits `forwardOnly`, the diagnostic asks the user to set it on every conflicting entry.

## Stacking

Stacked on `feat/codegen-match-emission` (#93). Merge order: …#93 → this → main.

## Test plan

- [x] `BridgeProcessorTest` — four new cases: happy fan-out (forward writes to both targets, backward reads primary only), partial opt-in rejected, type mismatch rejected, missing extra target rejected. All pass.
- [x] Existing `BridgeProcessorTest` cases — strict duplicate-source rejection without `forwardOnly` still fires identically.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -82,11 +82,19 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
   // read in generate().
   private final Map<TypePair, Set<String>> dropsByPair = new HashMap<>();
 
-  // Per-pair renames: source field name -> target field name for fields whose two sides have
-  // different names. The bijection check applies renames on the source side before comparing
-  // to target names; planFields and the field-read expressions follow the mapping. Populated in
+  // Per-pair renames: source field name -> primary target field name for fields whose two sides
+  // have different names. The bijection check applies renames on the source side before comparing
+  // to target names; planFields and the field-read expressions follow the mapping. For forward-only
+  // fan-out (one source feeding multiple targets), this map holds the FIRST-declared target — that's
+  // the one backward reads from. Extra fan-out targets live in renameFanoutsByPair. Populated in
   // process(), read in generate().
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #96, opened 2026-06-13
test(deepmap): close 5 uncovered branches + annotate 2 defensive guards

eschizoid:
## Summary

Round-2 DeepMap coverage audit identified five live branches that no existing test pinned and two dead-branch defensive guards that aren't reachable through any public API.

## Live branches covered (5 new tests)

| # | Branch | DeepMap.java |
|---|---|---|
| **D1** | zip(...) backward cardinality mismatch | line 609 |
| **D2** | autoIso Map key-type mismatch | line 688 |
| **D3** | via(...) auto-lifting through Set + Optional | lines 823-825 |
| **D4** | via(...) Map key-type mismatch | line 812 |
| **D5** | primitiveDefault for long/byte/short/char/float | lines 1128-1134 |

## Dead branches annotated (NOT testable)

D6 — fieldIsoOf exhaustiveness throws (lines 767-787). Only fires if a future Mapping permit is added and populateIso forgets to short-circuit. Marked DEAD-BRANCH-DEFENSIVE.

D7 — ForwardOnlyTransformTo throwingBackward lambda (line 752). Unreachable via the public API since PR #91 added rejectForwardOnlyRows. Marked DEAD-BRANCH-DEFENSIVE.

## Stacking

Stacked on `feat/gradle-config-cache` (#95). Top of the post-1.x cascade.

## Test plan

- [x] 6 new DeepMapCoverageTest cases pass
- [x] Full ./gradlew test green across all modules
- [x] D6/D7 sites carry DEAD-BRANCH-DEFENSIVE annotations for future coverage report readers

PR #97, opened 2026-06-13
feat(codegen): @Transform(forwardOnly=true) closes runtime parity gap

eschizoid:
## Summary

**P1 — runtime/codegen parity gap.** Closes the asymmetry where runtime has `Mapping.forward(srcAcc, tgtAcc, fn)` for forward-only typed transforms but codegen's `@Transform` was always bidirectional. Users who needed a one-way field conversion had to write a `BridgeFn` whose backward stubbed or threw, and the generated bridge would faithfully call it on every backward direction.

## What changed

`@Transform` gains `boolean forwardOnly() default false`. When true:

- **Forward emission** unchanged — routes through `BridgeFn.forward` as before.
- **Backward emission** skips the `__tx_<field>.backward(...)` call and emits the type's zero value instead (mirrors how `@Bridge(drops = …)` fills dropped sources on backward).

```java
@Bridge(value = AuditEntity.class, transforms = {
  @Transform(field = "createdAt", using = AuditTimestampFn.class, forwardOnly = true)
})
public record Audit(String id, Instant createdAt) {}

// Generated AuditBridge:
public static AuditEntity forward(final Audit s) {
  return new AuditEntity(s.id(), __tx_createdAt.forward(s.createdAt()));  // forward via BridgeFn
}
public static Audit backward(final AuditEntity t) {
  return new Audit(t.id(), null);  // zero-fill — BridgeFn.backward NOT invoked
}
```

The user's `BridgeFn.backward` can stub or throw; the generated bridge never invokes it. End-to-end mirrors the runtime `Mapping.forward(...)` semantics.

## Stacking

Stacked on `feat/deepmap-coverage` (#96). First of five codegen parity PRs (P1-P5 from the round-2 parity audit).

## Test plan

- [x] New BridgeProcessorTest case asserts forward routing, backward zero-fill, AND absence of `__tx_<field>.backward` call site.
- [x] All 20 existing BridgeProcessorTest cases pass.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -105,6 +105,12 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
   // the same-type bijection check. Populated in process(), read in generate().
   private final Map<TypePair, Map<String, String>> transformsByPair = new HashMap<>();
 
+  // Per-pair set of source field names whose @Transform was declared with forwardOnly = true.
+  // Backward direction emits a zero-value fill for these slots (the same shape `drops` uses) and
+  // never invokes the user's BridgeFn.backward. Mirrors the runtime Mapping.forward(...) semantics.
+  // Populated in process(), read in generate().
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #98, opened 2026-06-13
feat(codegen): @Default annotation for null-coalescing source values

eschizoid:
## Summary

**P2 — runtime/codegen parity gap.** Closes the asymmetry where runtime has `Mapping.toOrElse(srcAcc, tgtAcc, defaultValue)` for null-coalescing defaults but codegen had no equivalent — users had to write `@Transform` with a `BridgeFn` that checked null inline.

## Shape

```java
@Bridge(value = UserEntity.class, defaults = {
  @Default(field = "region", value = "EMEA"),
  @Default(field = "tier",   value = "FREE")
})
public record User(String id, String region, String tier) {}

// Generated UserBridge:
public static UserEntity forward(final User s) {
  return new UserEntity(
    s.id(),
    (s.region() == null ? "EMEA" : s.region()),
    (s.tier() == null ? "FREE" : s.tier())
  );
}
public static User backward(final UserEntity t) {
  return new User(t.id(), t.region(), t.tier());  // backward identity
}
```

Strict-null only — analogous to the 3-arg runtime `toOrElse`. The predicate-gated 4-arg variant stays runtime-only because annotation attributes can't hold predicates.

## Validation

- Source field must exist.
- Source field must be reference-typed (primitives can't be null — clear error message points the user at the wrapper type).
- Cannot overlap with `drops` on the same pair (incompatible semantics).
- Value parses via `parseConstantLiteral` against the source field type — same parser as `@Constant`: String, numeric primitives via wrappers, boolean, char, the literal `"null"` for reference types.

## Stacking

Stacked on `feat/codegen-transform-forward-only` (#97). Second of five codegen parity PRs.

## Test plan

- [x] New BridgeProcessorTest case asserts happy-path null-coalesce emission.
- [x] Primitive-rejection diagnostic asserted.
- [x] All 22 existing BridgeProcessorTest cases pass.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -707,7 +707,7 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
         return;
       }
       if (drops.contains(t)) {
-        error(source, "@Bridge field \"" + t + "\" appears in both transforms and drops — it must be one or the other");
+        error(source, "@Bridge field \"" + t + "\" appears in both transforms and drops — pick one.");
         return;
       }
     }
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #99, opened 2026-06-13
feat(codegen): @ViaMapper escape hatch for nested-bridge overrides

eschizoid:
## Summary

**P3 — runtime/codegen parity gap.** Closes the asymmetry where runtime has `Mapping.via(srcAcc, tgtAcc, mapper)` for explicit nested-bridge delegation but codegen had no equivalent — when auto-recursion couldn't reach a nested type, the user was stuck.

## Shape

```java
@Bridge(value = OrderDto.class, viaMappers = {
  @ViaMapper(field = "address", using = AddressBridge.class)
})
public record Order(String id, Address address) {}

// Generated OrderBridge:
public static OrderDto forward(final Order s) {
  return new OrderDto(s.id(), AddressBridge.forward(s.address()));
}
public static Order backward(final OrderDto t) {
  return new Order(t.id(), AddressBridge.backward(t.address()));
}
```

No auto-sub-bridge is emitted for the delegated pair — the user-named class takes over.

## When to use

- **Opaque nested type** — third-party value class without `@Bridge` / `@Focus`.
- **Shared bridge constant** — two parents want to reference one cached bridge instead of emitting their own sub-bridges.
- **Hand-rolled non-default bridge** — nested pair has a custom bridge under a non-default name.

## Implementation

Reuses the existing `FieldPlan.RECURSE` arm — `applyForward`/`applyBackward` already emit `<class>.forward(...)` / `<class>.backward(...)` for RECURSE plans, so no new dispatch infrastructure was needed.

## Validation

- Field must exist on source.
- Cannot overlap with `drops` or `transforms` on the same pair.
- The using-class's forward/backward signature is validated at the generated code level — `javac` surfaces signature mismatches at the generated bridge body.

## Stacking

Stacked on `feat/codegen-default-annotation` (#98). Third of five codegen parity PRs.

## Test plan

- [x] New BridgeProcessorTest case asserts forward + backward route through the user-named bridge.
- [x] No auto-sub-bridge emitted for the delegated pair.
- [x] All 24 existing BridgeProcessorTest cases pass.


eschizoid:
@copilot please review

PR #100, opened 2026-06-13
feat(codegen): @Bridge(writeStrategy = …) forces a POJO construction shape

eschizoid:
## Summary

**P4 — runtime/codegen parity gap.** Closes the asymmetry where runtime has `WriteHint.writeBean(cls, strategy)` to force a specific POJO construction path but codegen had no override — `buildExpr` always ran the priority ladder (name-matched ctor → static `builder()` → no-arg + setters).

## Shape

```java
@Bridge(value = HibernateEntity.class, writeStrategy = WriteStrategy.SETTERS)
public record Dto(String id, String name) {}
```

`WriteStrategy` enum: `AUTO` (default — preserve today's ladder), `CONSTRUCTOR`, `BUILDER`, `SETTERS`. Records ignore the setting (canonical ctor always). When forced, the requested strategy must apply; otherwise the processor surfaces a precise diagnostic naming the strategy and the alternative paths.

## What's NOT exposed

FIELDS strategy. Codegen never falls back to field reflection — exposing FIELDS would defeat the codegen path's no-runtime-reflection guarantee. Use the runtime path (`Telescope.mapper(...)` + `writeBean(..., FIELDS)`) if field reflection is required.

## Stacking

Stacked on `feat/codegen-via-mapper` (#99). Fourth of five codegen parity PRs.

## Test plan

- [x] SETTERS forces no-arg+setters even when builder is present (verified by asserting absence of `PB.builder()` in the emitted forward).
- [x] CONSTRUCTOR without a matching ctor produces a precise diagnostic mentioning `writeStrategy = CONSTRUCTOR`.
- [x] All 25 existing BridgeProcessorTest cases pass.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -1143,9 +1143,11 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
       );
     };
 
-    final var forwardBody = buildExpr(target, readForward, targetFields, writeStrategy);
+    // Pass `source` as the annotation site so write-strategy errors land at the user's @Bridge
+    // declaration rather than at `target` (which may be a third-party POJO with no annotation).
+    final var forwardBody = buildExpr(target, readForward, targetFields, writeStrategy, source);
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #101, opened 2026-06-13
feat(codegen): emit patch(base, partial) on generated *Bridge classes

eschizoid:
## Summary

**P5 — runtime/codegen parity gap. The last one of Phase 4.** Runtime has `Mapper.patch(base, partial)` for sparse-overlay semantics (HTTP PATCH on a resource: only non-null target components affect the source). Codegen bridges were full-conversion only.

## Shape

The generated `*Bridge` class now exposes:

```java
public static Source patch(final Source base, final Target partial)
```

For each source component:

- **Reference-typed target components null-gate to base** — the classic sparse-PATCH shape (only non-null partial fields propagate).
- **Primitive target components are always overlaid from partial** — primitives autobox to non-null wrappers on read, so they can't signal "unset". Matches runtime `Mapper#patch` semantics exactly.
- **Drop / forward-only-transform slots** read from `base` (the partial has nothing meaningful for that slot).

Example for `record User(String id, String email, int age)`:

```java
public static User patch(final User base, final UserDto partial) {
  return new User(
    (partial.id()    != null ? partial.id()    : base.id()),
    (partial.email() != null ? partial.email() : base.email()),
    partial.age()  // primitive int — always overlaid
  );
}
```

## Implementation

Reuses `buildExpr` with a custom `readPatch` closure, so records get canonical-ctor rebuild and POJOs get the same writeStrategy ladder (CONSTRUCTOR → BUILDER → SETTERS) the forward / backward paths use. The user's `writeStrategy` attribute applies to all three emit paths uniformly.

## Stacking

Stacked on `feat/codegen-write-strategy` (#100). **Last of five codegen parity PRs.**

## Phase 4 complete

| PR | Closes |
|---|---|
| #97 | P1 `@Transform(forwardOnly=true)` |
| #98 | P2 `@Default(field, value)` |
| #99 | P3 `@ViaMapper(field, using)` |
| #100 | P4 `@Bridge(writeStrategy = …)` |
| **#101** | **P5 `patch(base, partial)` emission** |

After Phase 4 merges, the codegen surface reaches feature parity with the runtime surface modulo the few items that are legitimately runtime-only (hooks, ForwardMapper.then, nested-path `Mapping.to(Telescope...)`, Telescope.before/after).

## Test plan

- [x] New BridgeProcessorTest case: patch signature, reference fields null-gate, primitive int field doesn't null-gate.
- [x] All 26 existing BridgeProcessorTest cases pass.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -1149,6 +1149,29 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
     if (forwardBody == null) return;
     final var backwardBody = buildExpr(source, readBackward, sourceFields, writeStrategy, source);
     if (backwardBody == null) return;
+    // Patch emits a sparse overlay: for each source field, read from `partial` (the user's
+    // partially-populated target) when the target's component is non-null, else fall back to the
+    // corresponding read on `base`. Primitive target components are always treated as
+    // patch-present (matching runtime Mapper#patch semantics: a boxed primitive is never null).
+    final Function<String, String> readPatch = sourceName -> {
+      final var sf = fieldByName(sourceFields, sourceName);
+      final var baseRead = readExpr(source, "base", sf);
+      if (drops.contains(sourceName)) return baseRead;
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

PR #105, opened 2026-06-14
feat(mapping): MS-2 Mapping.enumTo + MS-3 cycle handling coverage

eschizoid:
## Summary

Two Tier 4 items pulled forward to pre-1.0:

### MS-2 — `Mapping.enumTo(srcAcc, tgtAcc, SrcEnum.class, TgtEnum.class)`

Convenience factory closing MapStruct's `@ValueMapping` gap for by-name enum correspondence. Eager exhaustiveness check at factory time names any missing constants on either side — the value-add over a hand-written `Mapping.to(...)` with two `Enum.valueOf` lambdas.

```java
Telescope.mapper(UserEntity.class, UserDto.class,
    enumTo(UserEntity::status, UserDto::status, EntityStatus.class, DtoStatus.class));
```

Asymmetric enums get a clear diagnostic pointing at `Mapping.to(src, tgt, fwd, bwd)` for the explicit-fallback escape hatch.

### MS-3 — Cycle handling: already implemented, now pinned

Investigation showed `DeepMap.populateIso` already handles cycles via the cache reservation at `DeepMap.java:271-272` (the `cache.put(key, null)` placeholder + `containsKey` short-circuit). The org-chart example demonstrates this in production for Optional/List self-references.

The earlier review framing — "bidirectional graphs currently StackOverflow" — was incorrect. No code change needed; just coverage tests pinning the additional shapes the cache must handle:

- Self-reference through `Optional<Self>`
- Self-reference through `List<Self>`
- Mutual recursion A↔B through Optional containers

## Test plan

- [x] `./gradlew test` green (46 tasks)
- [x] 5 new `MappingEnumToTest` cases (round-trip, composition, 3 exhaustiveness diagnostics)
- [x] 3 new `CycleHandlingTest` cases (Optional self-ref, List self-ref, mutual A↔B)
- [x] All existing tests unchanged

## Future work

Codegen path for `enumTo` (emit a switch expression at the `@Bridge` processor instead of routing through the captured `Function`) is mentioned in the javadoc as a future revision — runtime-first lands the API, codegen optimization can follow once usage is real.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/mapping/Mapping.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/mapping/Mapping.java
@@ -144,6 +144,83 @@ public sealed interface Mapping<A, B>
   }
 
   /**
+   * Enum correspondence by constant name. Maps each {@code SE} constant to the {@code TE} constant
+   * of the same {@link Enum#name() name}; backward direction is symmetric. Closes MapStruct's
+   * {@code @ValueMapping} gap for the common "status enums that line up by name" case without
+   * forcing the user to hand-write a 2-arg typed transform with {@code Enum.valueOf} on both sides
+   * — the exhaustiveness check that comes free here is the value-add.
+   *
+   * <pre>{@code
+   * enumTo(UserEntity::status, UserDto::status, EntityStatus.class, DtoStatus.class)
+   * }</pre>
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #106, opened 2026-06-14
docs(readme): close MapStruct-gap claims + trim duplication

eschizoid:
## Summary

Six minor versions of feature work landed since the README was last touched (0.4.1 → 0.10.0). Several MapStruct comparison claims are flat-out wrong now, and a couple of sections duplicate each other.

### What was wrong / stale

- **Version: 6 minors stale** (0.4.1 → 0.10.0). Bumped in 8 places across Gradle + Maven snippets.
- **Comparison table row "`@BeforeMapping` / `@AfterMapping` hooks"**: said "Not supported". Shipped in #87 (`Mapper.before/afterForward/Backward`) and #90 (`Telescope.before/after`).
- **Comparison table row "Spring / Quarkus / CDI integration"**: said "None today — bring-your-own wiring". `telescope-spring-boot-starter` and `telescope-quarkus` are first-class published artifacts.
- **"When MapStruct is the right pick"**: listed `@BeforeMapping`/`@AfterMapping` hooks and Spring/Quarkus as MapStruct-only reasons. Both shipped on the telescope side. Replaced with the actual remaining gaps (`@InheritConfiguration`, `@SubclassMapping`, `@MappingTarget`, qualifier dispatch, embedded expression-language bodies).
- **Code example used `Mapping.auto()`** — that factory doesn't exist (recursion is auto by default, no row needed).
- **Code example used `Telescope.mapper(...)`** without the source/target `Class` args its actual signature requires.
- **Code example used `Telescope.map(A.class).to(B.class).field(...).build()`** — that fluent chain doesn't exist (only `Telescope.map(Class, Class, MapStep...)` does).

### What was duplicated

- **"What it is not" MapStruct paragraph** repeated the full "How it compares" section in narrative form. Trimmed to a brief framing pointing at the comparison table as canonical.
- **"Working with POJOs → Convert"** re-explained what "Type conversion → Deep recursive mapping" already covered. Trimmed to point back at Type conversion for the general factory and focus this section on the POJO-specific `writeBean` / `writeBeans` levers.

### What was added

- Comparison table gained rows for sealed-root dispatch, multi-source mappers, forward-only mappers, enum value mapping, null-coalescing defaults, and conditional/drop — all shipped post-0.4.1.
- "When telescope is the right pick" gained multi-source merge and sealed-root dispatch as differentiators MapStruct can't express.
- Override-forms cookbook in Type conversion now lists all 9 factories (`to` / `to`-transform / `forward` / `toOrElse` / `toOrElseGet` / `enumTo` / `via` / `constant` / `compute` / `drop`) with one-line annotations — was 3 rows.

## Test plan

- [x] Markdown renders correctly (verified anchors locally)
- [x] No remaining "Not supported" / "None today" / "bring-your-own" stale claims (greps clean)
- [x] All version references at 0.10.0
- [x] Code examples match actual API signatures

Net: 71 insertions, 97 deletions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,11 +207,11 @@ capability lists, vs-MapStruct callouts, and benchmark cross-links.
   On deeper workloads (nested records with list-of-records inside) the two engines match. See
   [Performance honesty](#performance-honesty).
 
-  Telescope covers the common `@Mapping(...)` shapes — same-name auto, renames, typed transforms, nested mappers,
-  flat → nested-path correspondences, eager literals, per-call computed values, forward-only mappers, multi-source
-  merge, by-name enum mapping, null-coalescing defaults, lifecycle hooks, and Spring/Quarkus autoconfig — plus the
-  shapes MapStruct can't compose to: deep navigation, effectful update, sealed-root dispatch, JPA cycles and Hibernate
-  `LAZY` proxy unwrap, bidirectional from one declaration. See [How it compares to MapStruct](#how-it-compares-to-mapstruct)
+  Telescope covers the common `@Mapping(...)` shapes — same-name auto, renames, typed transforms, nested mappers, flat →
+  nested-path correspondences, eager literals, per-call computed values, forward-only mappers, multi-source merge,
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #107, opened 2026-06-14
docs(readme): restructure for new-user clarity

eschizoid:
## Summary

Reorganized the README around a linear new-user journey instead of treating every reader as a MapStruct migrant.

### New top-of-file flow

1. **Tagline** (1 strong sentence; was 5 dense sentences)
2. **Install** (terse runtime dep; was at line 367)
3. **First 5 minutes** (NEW — linear `Address`/`User` tutorial with 4 next-step branches)
4. **Picking your entry point** (records/POJOs × navigate/convert picker; was buried inside "The DSL surface")
5. **More 30-second vignettes** (the existing Records/Mapping/Beans triple, demoted from primary intro)

### Key rewrites

- **Pitch**: "Build a typed path through your nested data — then read, update, or convert through it. Bidirectionally. One line." Drops "Optics-based DSL" from the headline (academic) and the catch-all opening paragraph.
- **Effects intro**: leads with the wow ("validate every email in a Batch and report all the bad ones in one call? Two lines.") instead of a dry list of method names.
- **Architecture**: collapsed ~35 lines of implementer-facing detail to a short paragraph + the same composition-rules sentence. (Inlined the prior detail rather than linking to `CLAUDE.md`, since `.claude/` is gitignored.)

### Consolidations

- **Two JPMS sections** (one about `requires`, one about `opens`) merged into a single section under "Additional artifacts". Was confusing to have JPMS appear twice.
- **"## Installation" → "## Additional artifacts"** since the basic install moved to the top and this section now serves the 6-artifact reference role.

### Net

Same line count (~1500), but time-to-first-useful-code drops from ~5 minutes of skimming to ~90 seconds. The deep reference (cookbook, DSL surface tables, MapStruct comparison, JPMS, codegen) all preserved.

## Test plan

- [x] Prettier markdown reflow applied locally
- [x] No `CLAUDE.md` or `.claude/` references (gitignored)
- [x] All section anchors valid
- [x] Section structure verified via `awk '/^## /'` map

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,10 @@
 
 # telescope
 
-**Optics-based DSL for Java records and POJOs.**
+**Build a typed path through your nested data — then read, update, or convert through it. Bidirectionally. One line.**
 
-One type drives deep navigation, immutable update, bidirectional mapping, and effectful update across records, plain
-POJOs, and Lombok `@Data` classes — no category-theory jargon, no hand-written copy constructors. Reach fields nested
-through lists, sets, maps, optionals, and sealed-type variants in one chain. Skip codegen for the runtime path, or add
-`@Focus` / `@BeanFocus` / `@Bridge` annotations for compile-time-bound navigators with deep recursion through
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #111, opened 2026-06-14
feat(mapping): Mapping.when(predicate, inner) — predicate-gated telescope rows (#1.3)

eschizoid:
## Summary

Closes Tier 2 #1.3 — MapStruct's `@Condition` equivalent for whole-source predicate gating on telescope-based rows.

```java
    to(Order::id, OrderDto::id),
    when(o -> o.shipping() != null,
        to(Telescope.of(Order.class).field(Order::shipping).field(Shipping::country),
           OrderDto::shipCountry)),
    when(o -> o.priority() == Priority.HIGH,
        constant(OrderDto::expediteFlag, true)));
```

## Design points

- **Scope:** wraps the five telescope-based row kinds (`TelescopeTo`, `FromTelescopeTo`, `TelescopeToTelescope`, `Constant`, `Compute`). Field-iso rows (`SameTypedTo`, `TypedTransformTo`, `Via`, `Drop`) reject at construction with a precise IAE pointing at `Mapping.toOrElse(src, tgt, default, predicate)` which already covers the field-level conditional case.
- **Forward-only by design:** same retraction semantics as `Constant` / `Compute` — `applyBackward` skips Conditional rows entirely. Predicate isn't evaluable from a target, and asymmetric round-tripping is the inherent shape of conditional logic.
- **Top-level pinning:** `Conditional.sourceClass()/targetClass()` return `null` unconditionally. Telescope-based inner rows already do this (erased root class on at least one side); making it explicit on the wrapper documents the contract and prevents a future inner permit with a non-null class from silently breaking the routing.
- **Codegen 1:1 via composition:** the codegen-emitted `<X>Telescope` constants ARE `Telescope` instances, so `when(p, constant(OrderDtoTelescope.of().audit().tenant(), "prod"))` uses the compile-checked codegen path inside the predicate gate without any processor-side changes.
- **Apply-time error decoration:** user-predicate exceptions wrapped in `IllegalStateException` naming the inner kind + source field so failures point at the `when(...)` call site, not at an opaque `applyForward` stack frame.

## Sealed-permit + DeepMap routing

- Added `Conditional<A, B>(Predicate<? super A>, Mapping<A, B>)` to the `Mapping` sealed permit list.
- `populateIso` unwraps Conditional to the inner row for soft-claim routing decisions, but registers the Conditional itself in `telescopeFixups` so `applyForward` can evaluate the predicate before dispatching the inner's effect.
- `applyForward` evaluates the predicate at the top of the row-dispatch loop; on rejection, `continue` skips the inner.
- `applyBackward` skips Conditional rows in both loops (field-overrides collection + telescope-source overlays).

## Test coverage

19 test cases across 9 nested classes covering:
- All 5 supported inner kinds, forward direction
- All inner kinds backward (skip semantics)
- Self-recursive types (cycle pin)
- Top-level pinning contract (asserts the null-class invariant)
- Predicate composition (`Predicate#and`)
- Predicate exception decoration
- All construction-time rejection paths (field-iso, forward-only, nested, null predicate, null inner)

## Test plan

- [x] `:core:test --tests MappingWhenTest` (19/19)
- [x] `:core:test` (full regression)

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -558,13 +558,20 @@ public final class DeepMap {
           // propagate raw — the breadcrumb is even more valuable for those.
           final var inner = cond.inner();
           final var innerField = inner.sourceField() == null ? "<telescope>" : inner.sourceField();
+          // Include the failure CLASS name — predicate.test() commonly throws NPE on null
+          // navigation, where getMessage() returns null and "Predicate failure: null" tells the
+          // user nothing. The class name (NullPointerException, ClassCastException, etc.) carries
+          // the actionable signal even when the message is null.
+          final var failureType = predicateFailure.getClass().getSimpleName();
+          final var failureMsg = predicateFailure.getMessage();
           throw new IllegalStateException(
             "Mapping.when(...) predicate threw — inner=" +
               inner.getClass().getSimpleName() +
               " (sourceField=" +
               innerField +
               "). Predicate failure: " +
-              predicateFailure.getMessage(),
+              failureType +
+              (failureMsg == null ? "" : ": " + failureMsg),
             predicateFailure
           );
         }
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #112, opened 2026-06-14
feat(mapping): NullHint.nullSourceValues(DEFAULT) — null source → type defaults (#1.4)

eschizoid:
## Summary

Closes Tier 2 #1.4 — MapStruct's `nullValuePropertyMappingStrategy=SET_TO_DEFAULT` equivalent for per-mapper null-source-value handling.

```java
import static io.github.eschizoid.telescope.mapping.NullHint.NullStrategy.DEFAULT;
import static io.github.eschizoid.telescope.mapping.NullHint.nullSourceValues;

final Mapper<UserEntity, UserDto> userMapper = Telescope.mapper(
    UserEntity.class, UserDto.class,
    nullSourceValues(DEFAULT),                  // any null source field → type default
    to(UserEntity::id, UserDto::userId));

// Null name → "", null Integer → 0, null List → List.of(), null BigDecimal → ZERO, etc.
```

## Design points

- **Lattice-honest**: new `Iso.coalesceForward(inner, defaultValue)` factory in `:internal/optics/`. Named explicitly to signal it is NOT a proper Iso — the round-trip law cannot hold at the null/default boundary because `SET_TO_DEFAULT` is deliberately non-bijective. Differentiating the factory from `Iso.of(...)` keeps `OpticLawsTest`'s contract honest.
- **Per-row precedence**: `Mapping.toOrElse(...)` wins over the global hint without any extra marker. SameTypedTo rows are wrapped; TypedTransformTo (which `toOrElse` constructs) is NOT — the user-supplied forward function handles null however they wrote it.
- **Forward-only**: backward direction skips substitution entirely. Same retraction semantics as the existing `Constant` / `Compute` rows — javadoc names the precedent explicitly.
- **No cascade**: substitution consults only the per-leaf `NullDefaults` table. Record/bean/enum/custom-type leaves stay at null; the engine's recursive-default allocation runs as it always does.
- **Codegen 1:1 via composition**: hint is a runtime assembly concern. Codegen-emitted `@Bridge` mappers used inside hinted outer mappers inherit the strategy automatically. No processor change needed.

## NullDefaults coverage

| Leaf type | Default |
|---|---|
| `String` | `""` |
| `Integer`/`Long`/`Double`/`Float`/`Short`/`Byte` | JLS zero (boxed) |
| `Boolean` | `false` |
| `BigDecimal` / `BigInteger` | `ZERO` |
| `List` / `Set` / `Map` | empty singleton |
| `Optional` | `Optional.empty()` |
| Records / beans / enums / arrays / UUID / Instant / custom | `null` (forces explicit `toOrElse`) |

## Touched files

- NEW: `core/.../mapping/NullHint.java` — sealed `MapStep` permit + factory + enum
- NEW: `internal/.../NullDefaults.java` — type → default substitution table
- NEW: `internal/.../optics/Iso.java` — `coalesceForward` factory + lattice-posture javadoc
- NEW: `core/.../NullStrategyTest.java` — 13 tests across 10 nested classes
- MODIFIED: `core/.../DeepMap.java` — `extractNullStrategy`, `wrapDefaultOnNull`, threading
- MODIFIED: `core/.../mapping/MapStep.java` — added `NullHint` permit

## Test plan

- [x] `:core:test --tests NullStrategyTest`
- [x] `:core:test` regression


eschizoid:
@copilot please review

PR #113, opened 2026-06-14
feat(mapper): Mapper.into(target, source) — in-place bean updates / @MappingTarget (#1.2)

eschizoid:
## Summary

Closes Tier 2 #1.2 — MapStruct's `@MappingTarget` equivalent for in-place bean target mutation.

```java
final UserEntity managed = repository.findById(id).orElseThrow();
userMapper.into(managed, dto);    // writes mapped fields onto `managed` in place
repository.save(managed);
```

Common JPA pattern: load entity, mutate with DTO fields, save — preserving the entity's identity so the persistence-context tracking survives.

## Design points

- **Identity preservation**: returns `target` (same reference passed in) for fluent chaining
- **Hook composition**: before/after hooks run on `into()` identically to `forward()` — same semantics, just final write step targets an existing instance rather than a fresh allocation
- **Records rejected** at apply time with `UnsupportedOperationException` naming the constraint and pointing at `Mapper.forward(source)` as the alternative
- **Null guards**: both target and source rejected up front with `NullPointerException`
- **No no-arg ctor required**: `Beans.writeBeanProperty(pojo, name, value)` helper takes a user-supplied instance — unlike `settersWriter(cls)` which allocates via no-arg ctor
- **Atomicity (review pass)**: two-phase apply — every (name, value) read into a staging map BEFORE any setter runs on target. If a setter throws midway, target stays untouched by the staging phase
- **Scoping (review pass)**: iterate `patchByTargetField.keySet()` (the mapper's actual produced fields) not every getter-derived property — preserves unmapped fields on the managed entity and avoids IAE on read-only computed getters

## Implementation

- NEW: `Beans.writeBeanProperty(pojo, name, value)` — mirrors `Beans.readProperty(pojo, name)` shape. `ClassValue<ConcurrentHashMap>` setter cache, LMF-built `BiConsumer<Object, Object>` per (class, property), same hot-path posture as readProperty
- NEW: `Mapper.into(target, source)` — null guards, record-target rejection, `forward(source)` to fresh, two-phase staged copy

## Codegen 1:1

`into(...)` operates on any Mapper produced by `Telescope.mapper(...)`, including those built from `@Bridge`-generated bridges via `Mapping.via(...)` composition. The codegen-emitted Iso participates in the same `forward()` that `into()` drives — no processor change needed.

## Test coverage

11 tests across 8 nested classes:
- Bean target identity preservation (`assertSame`)
- Repeatable mutation
- `afterForward` hook composition through `into()`
- `beforeForward` hook composition through `into()`
- Record target rejection (UnsupportedOperationException with clear message)
- Null target / null source rejection
- Missing setter error path
- Bean → bean variant
- Hook semantics symmetry between `into()` and `forward()`
- Two-phase staged-write atomicity pin

## Test plan

- [x] `:core:test --tests MapperIntoTest`
- [x] `:core:test` full regression

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/conversion/Mapper.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/conversion/Mapper.java
@@ -1,6 +1,7 @@
 package io.github.eschizoid.telescope.conversion;
 
 import io.github.eschizoid.telescope.Telescope;
+import io.github.eschizoid.telescope.internal.Beans;
 import io.github.eschizoid.telescope.internal.Reflective;
 import io.github.eschizoid.telescope.internal.optics.Iso;
 import io.github.eschizoid.telescope.mapping.MapStep;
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #115, opened 2026-06-14
feat(codegen): @Transform(method = ...) — @Named qualifier dispatch (#1.1)

eschizoid:
## Summary

Closes Tier 2 #1.1 — MapStruct's `@Named` qualifier dispatch for the codegen path. The runtime path already supports it via `Mapping.forward(srcAcc, tgtAcc, methodRef)`.

```java
public final class DateHelpers {
  public static String expiry(Instant i)    { return DateTimeFormatter.ISO_INSTANT.format(i); }
  public static String createdAt(Instant i) { return i.atZone(ZoneOffset.UTC).toString(); }
}

@Bridge(value = UserDto.class, transforms = {
  @Transform(field = "expiresAt",    using = DateHelpers.class, method = "expiry"),
  @Transform(field = "registeredAt", using = DateHelpers.class, method = "createdAt")
})
public record UserEntity(String id, Instant expiresAt, Instant registeredAt) {}
```

The "one BridgeFn class per transform" pattern doesn't scale when a codebase has 50+ field transforms sharing infrastructure (date formatting, currency, ID encoding). This unlocks the natural shape: one helper class hosting multiple static methods, different fields pointing at different methods.

## What changes

- New `String method() default ""` element on `@Transform`
- `using` relaxed from `Class<? extends BridgeFn>` to `Class<?>` — source-compatible (existing BridgeFn classes still satisfy)
- Generated bridge emits direct `UsingClass.methodName(value)` calls inline — NO `__tx_<field>` singleton, NO instantiation, NO BridgeFn requirement
- Qualifier rows implicitly forward-only: backward direction zero-fills (the inverse method name isn't recoverable from a forward method name)
- BridgeFn-shape and qualifier-shape rows coexist on the same `@Bridge` — pinned by test

## Backwards compatibility

Existing `@Transform(using = BridgeFnClass.class)` rows compile and behave identically — no migration required. The two shapes are routed at codegen-emission time based on whether `method` is non-empty.

## Implementation

- `TransformSet` extended with `methodsByField` alongside `byField`
- `BridgeConfig` extended with `transformMethods`
- `FieldPlan` extended with `qualifierMethod` field + new `ofTransformQualified` factory
- `planFields` routes to `ofTransformQualified` when `method` is set
- `applyForward` emits `UsingClass.methodName(...)` for qualifier rows
- Singleton declaration loop skips qualifier-dispatch fields
- Qualifier rows land in `forwardOnlyFields`, reusing the existing zero-fill backward path
- Processor-time validation (review pass): blank method name rejected, non-existent method rejected, non-static method rejected, method name must be a valid Java identifier

## Test coverage

Five BridgeProcessorTest cases:
1. **Pure qualifier dispatch** — no `__tx_` instance emitted, `DateHelpers.method()` calls inline, backward zero-fill
2. **Coexistence** — same `@Bridge` with one BridgeFn-shape row + one qualifier-shape row; both work independently
3. **Blank method name rejected** at processor time
4. **Non-existent method rejected** at processor time
5. **Non-static (instance) method rejected** at processor time

## Test plan

- [x] `:codegen:test` (regression + 5 new tests)
- [x] `:core:test`
- [x] `:lombok:test :spring-boot-starter:test :quarkus:test`
- [x] `:core:spotlessJavaCheck :codegen:spotlessJavaCheck`

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -387,11 +387,69 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
           error(element, "@Transform field=\"" + field + "\" is declared more than once");
           return null;
         }
+        // Validate `method` is either empty (legacy BridgeFn shape) or a syntactically valid Java
+        // identifier — whitespace, punctuation, or reserved-word inputs would otherwise produce
+        // emit-time syntactic garbage at the generated bridge body that javac surfaces with a
+        // cryptic "identifier expected" pointing at synthetic source the user never wrote.
+        final boolean hasMethod = method != null && !method.isBlank();
+        if (method != null && !method.isEmpty() && !hasMethod) {
+          error(element, "@Transform `method` must not be blank (was \"" + method + "\")");
+          return null;
+        }
+        if (hasMethod && !javax.lang.model.SourceVersion.isIdentifier(method)) {
+          error(element, "@Transform `method` must be a valid Java identifier (was \"" + method + "\")");
+          return null;
+        }
+        if (hasMethod) {
+          // Validate the named method exists, is static, and takes exactly one argument. Defers the
+          // parameter / return type compatibility check to javac at the generated bridge body —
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #117, opened 2026-06-14
feat(perf): Tier-A runtime fast-path — record→record + bean→bean

eschizoid:
## Summary

First slice of the runtime throughput closure plan ([Discussion #116](https://github.com/eschizoid/telescope/discussions/116)). Adds a runtime fast-path that bypasses structural `Iso` composition for record↔record + bean↔bean same-shape mappings.

## Measured before / after

JDK 25, Apple Silicon, 2-fork × 5-warmup × 10-iter × 1 s. `-Dtelescope.disableFastPath=true` forces the slow path for the before measurement.

| Fixture (5 flat scalar fields) | Slow path (ns/op) | Fast path (ns/op) | Δ |
|---|---:|---:|---:|
| record → record | 50.4 ± 6.6 | 44.0 ± 1.5 | ~12% |
| bean → record (mixed shape) | 375.9 ± 7.6 | — ¹ | — |
| record → bean (mixed shape) | 507.9 ± 7.6 | — ¹ | — |

¹ Mixed shape — fast-path tightly requires both sides be the same kind. Records use canonical constructor; beans use no-arg ctor + setters. Mixed-shape fast-path is a separate follow-up.

**Honest framing.** Earlier framing on this PR implied an 8× speedup, but that comparison was apples-to-oranges (bean→record slow path vs record→record fast path). The actual record↔record improvement is **~12%** because records were already cheap on the slow path — LMF-built readers + canonical-ctor spreader. The fast path eliminates the `LinkedHashMap` intermediate the structural `Iso` threads `name → value` through.

The **90× gap to MapStruct lives in the bean↔record / record↔bean rows** (376 ns / 508 ns), not the record↔record path. Bean getter/setter reflection through the structural Iso dominates. Closing that gap is the next research target (likely `LinkedHashMap → Object[]` structural intermediate).

## Scope

**Eligible (uses fast-path):**
- Records: both classes records, every component has same name + EXACTLY same type, all flat scalars
- Beans: both classes have no-arg ctor + same-named setters/getters with EXACTLY same type, all flat scalars

**Falls through to slow path:**
- Mixed shape (record↔bean)
- Container components (List/Set/Map/Optional) — preserves deep-copy + lift semantics
- Nested record/bean components — preserves cycle detection + composition
- Cross-type conversions (Integer ↔ Long)
- Any `Mapping` rows / `WriteHint` / `NullHint` — fast-path only fires when `steps.length == 0`

## Implementation

- `Records.fastPathForward(srcClass, tgtClass)` — cached `readers[]` (positional, LMF-built) + `ctorFn` (canonical-ctor spreader, LMF-built). Per-call: `Object[arity]` + N reader applies + ctor call
- `Beans.fastPathForward(srcClass, tgtClass)` — cached `GETTER_INVOKERS` + setter invokers (LMF-built). Per-call: no-arg ctor + iterate (getter, setter) pair array
- Per-`(srcClass, tgtClass)` cache; non-eligible entries cache a sentinel so subsequent lookups short-circuit
- `DeepMap.resolveMapper` integration: when `steps.length == 0` (and fast-path not disabled via system property), try records → beans → slow path
- `-Dtelescope.disableFastPath=true` escape hatch for benchmarking + diagnostics

## Lattice + reflection posture

- **Lattice-honest:** fast-path REPLACES the structural Iso for eligible pairs; for ineligible pairs, the full lattice still runs unchanged. No `Function<Object, Object>` plumbing — the Function returned IS the contract, built once at `resolveMapper` time
- **Zero per-call reflection:** all dispatch is LMF-built (same primitives as ADR-0005). One-time discovery uses `RecordComponent` / `Method.getMethods()` at cache-population time only

## Test plan

- [x] `:core:test` regression — CycleHandlingTest, DeepMappingTest, SetMappingTest all pass; fast-path correctly skips nested/container/cycle cases
- [x] `:internal:test` regression
- [x] New unit tests: `RecordsFastPathTest`, `BeansFastPathTest` — 16 cases covering eligibility (records, beans, identity, null source, cache reuse) + ineligibility (different name, different type, mixed shape, no no-arg ctor, container, nested)
- [x] `:benchmarks:jmh` long-fork measurement with `-Dtelescope.disableFastPath=true` for honest before/after

## Follow-ups

- Tier A v2: mixed-shape (record↔bean) fast-path
- Structural intermediate optimization: `LinkedHashMap` → `Object[]` indexed by component position — that's where the bean↔record 376 ns slow path actually pays the cost

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,9 +432,18 @@ the runtime path is fine for one-shot conversions in tests or non-hot service co
 both beans), have the same name + same type per component, and every component is a flat scalar (primitive / primitive
 wrapper / `String` / `enum`), the engine bypasses structural-`Iso` composition entirely. A per-pair `Function<S, T>` is
 built once at `Telescope.mapper(...)` time via cached `LambdaMetafactory` readers + canonical-constructor spreader (or
-no-arg ctor + setter dispatch for beans). Measured: flat record → record drops from **~360 ns/op to ~44 ns/op** — from
-**~91× to ~11× MapStruct**. Mixed shape (record ↔ bean), containers, nested records still go through the slow path so
-deep-copy / lift / cycle semantics survive unchanged.
+no-arg ctor + setter dispatch for beans). Measured at 2-fork × 5-warmup × 10-iter × 1 s (force the slow path with
+`-Dtelescope.disableFastPath=true`):
+
+| Fixture (5 flat scalar fields) | Slow path (ns/op) | Fast path (ns/op) |    Δ |
+| ------------------------------ | ----------------: | ----------------: | ---: |
+| record → record                |      50.4 ± 6.6   |      44.0 ± 1.5   | ~12 % |
+
+Record-to-record was already cheap on the slow path because records have LMF-built readers + canonical-ctor spreader;
+the fast path saves the `LinkedHashMap` intermediate the structural `Iso` threads `name → value` through. **The 90× gap
+to MapStruct lives in the bean ↔ record / record ↔ bean rows (376 ns, 508 ns)** — bean getter/setter reflection +
+`LinkedHashMap` combined. Closing that gap is the next research target. Mixed shape, containers, nested records still
+take the slow path so deep-copy / lift / cycle semantics survive unchanged.
 
 If you're in a tight inner loop where 1 ns matters, pick MapStruct. For realistic deep workloads — nested records with
 list-of-records inside — the codegen rows are a tie and you're picking on capability. Sealed-narrow paradigm hop,
--- a/README.md
+++ b/README.md
@@ -428,6 +428,14 @@ even with the LMF substrate, so it runs 30–80× slower than MapStruct's genera
 `Reflection.invoke`-based mappers. Sub-microsecond on flat and nested, single-microsecond on deep. Codegen on hot paths;
 the runtime path is fine for one-shot conversions in tests or non-hot service code.
 
+**Runtime fast-path — record ↔ record + bean ↔ bean.** When the source and target share the same kind (both records or
+both beans), have the same name + same type per component, and every component is a flat scalar (primitive / primitive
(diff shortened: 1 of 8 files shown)


eschizoid:
Closing without merge. Honest retrospective:

**Measured outcome:**
- Tier A (record↔record + bean↔bean fast-path): ~12% on record→record (50 → 44 ns/op). No effect on bean↔record — the canonical MapStruct use case
- Tier B (Iso composition fusion): within noise
- Tier C (LinkedHashMap → Object[] intermediate): not implemented; predicted 30-50% on bean↔X, would still leave 40-50× gap to MapStruct

**Why the predictions were optimistic.** I assumed the record→record slow path was as slow as the bean→record slow path. It wasn't — records were already cheap on the slow path because of LMF readers + canonical-ctor spreader. The 90× gap to MapStruct lives entirely in bean↔X (376 ns / 508 ns), which is bean reflection dispatch + LinkedHashMap intermediate.

**The substrate floor.** A 5-field bean read+write cycle through cached LMF `Function`/`BiConsumer` + canonical-ctor has a floor around 40-50 ns/op on JDK 25. We can't go below that without runtime bytecode generation (ASM/ByteBuddy), which the research doc explicitly ruled out (dep size, GraalVM, `--release 17`).

**Strategic conclusion.** Closing the runtime path to ≤2× MapStruct on bean↔X is unrealistic. The codegen path (`@Bridge` at 5.5 ns, ~1.4× MapStruct) is already competitive; that's where adopters should land for hot paths. Telescope's wins over MapStruct are capability (sealed-narrow paradigm hops, effectful update, JPA cycles, deep navigation, bidirectional from one declaration), not raw runtime perf.

**Code disposition.** Closing without merge. The `-Dtelescope.disableFastPath` escape hatch + the eligibility-detection scaffolding could be revived if future profiling reveals a specific bottleneck this PR's approach addresses cleanly. Not worth carrying 250 LOC + per-pair cache complexity for a 12% gain on a path nobody uses.

Research doc + discussion #116 updated with the conclusion.

PR #118, opened 2026-06-15
perf(deepmap): Object[] structural intermediate (LinkedHashMap → array)

eschizoid:
## Why

The runtime mapper's structural intermediate was a `LinkedHashMap<String, Object>` — measured as the dominant allocator (776 B/op on the flat bean→record benchmark) with hash bucket puts/gets in front of every field hop. Switching to a position-indexed `Object[arity]` drops allocation by 70-86% and lets the JIT scalar-replace the monomorphic call sites entirely. No public API change.

## What changed

- `Reflective.structuralIsoArr(...)` — new sibling to `structuralIso` returning `Iso<Object[], T>`. Records side rides the existing `Records.RecordInfo` positional `readers[]` + `ctorFn` (no new machinery). Bean side resolves a `Function<Object,Object>[]` once at composition time via `resolveReadersByPosition`; forward still uses the existing `construct` path (`Function<String,Object>`) wrapped over an array lookup.
- `Records.RecordInfo` / `Records.info(...)` made package-private so `Reflective.structuralIsoArr` can reuse the cached readers + ctor directly without going through the public reflective surface.
- `DeepMap.assembleIso` — now composes `Iso<S, Object[]> · Iso<Object[], Object[]> · Iso<Object[], T>` via the existing `Iso.then`. No hand-rolled `Function<Object,Object>` plumbing.
- `DeepMap.remapIsoArr` — precomputes `int[]` slot maps (forward and backward) + per-position `Iso[]` once per type pair, so the hot loop is one array index + one virtual `Iso#to`/`#from` per field. Sentinel `-1` for placeholder rows (where `step.sourceName` or `step.targetName` is null), matching the prior `Map.get(missingKey)` → `null` semantics.

## Before / after

JMH `avgt` (`Mode.AverageTime`), JDK 25, Apple Silicon, **rebooted machine, no competing work**. 18-row `MapStructComparisonBenchmark` at the standard config (3 warmup + 5 measurement × 1 fork, 10s/iter). Numbers are **measured side-by-side against the README baseline** captured on the same machine state pre-Tier-C.

| Tier   | Direction     | Before (Map)  | After (Object[]) | Δ ns | Δ %   |
|--------|---------------|--------------:|-----------------:|-----:|------:|
| flat   | bean → record | 340.1 ± 7.9   | **143.4 ± 2.3**  | -197 | **-58%** |
| flat   | record → bean | 464.3 ± 23.5  | **213.3 ± 2.5**  | -251 | **-54%** |
| nested | bean → record | 501.4 ± 15.4  | **241.2 ± 0.9**  | -260 | **-52%** |
| nested | record → bean | 709.9 ± 12.4  | **332.8 ± 5.1**  | -377 | **-53%** |
| deep   | bean → record | 2024.1 ± 380  | **926.8 ± 25.3** | -1097| **-54%** |
| deep   | record → bean | 2292.4 ± 28.2 | **1274.3 ± 120** | -1018| **-44%** |

Allocation reduction (GC profiler, focused-run state — structural change, unaffected by bench noise):

| Benchmark             | Before        | After          | Δ      |
|-----------------------|---------------|----------------|-------:|
| flat   bean → record  | 776 B/op      | 112 B/op       | **-86%** |
| flat   record → bean  | 856 B/op      | 256 B/op       | **-70%** |
| nested forward        | 1200 B/op     | 176 B/op       | **-85%** |
| nested backward       | 1296 B/op     | 384 B/op       | **-70%** |

Reference baselines on the same run (Tier C didn't touch these — sanity check passes):

| Engine                       | flat fwd | flat bwd | nested fwd | nested bwd | deep fwd | deep bwd |
|------------------------------|---------:|---------:|-----------:|-----------:|---------:|---------:|
| MapStruct                    |    3.48  |    3.47  |       5.19 |       5.33 |    47.43 |    60.68 |
| Telescope codegen (`@Bridge`)|    5.50  |    7.77¹ |      10.48 |      10.76 |    57.65 |    55.24 |

¹ flat backward codegen had an unusually high error (±12) on this run — outlier, not a regression.

## Mantras honoured

- **No inline FQNs.** All new code uses imported simple names.
- **Zero reflection.** Sits on the existing LMF-cached readers / ctor; no new `Method.invoke` or `Constructor.newInstance`.
- **Lattice-first.** The new structural Iso composes through `Iso.then` like every other path; no hand-rolled `Function<Object, Object>` plumbing.
- **Ergonomics preserved.** No public API touched.

## Tests

Full `:core` suite green (zero regressions). `OpticLawsTest` and `DeepMap` round-trip tests cover the new path.

## Notes

This branch also carries `c947021 build(benchmarks): wire -Pjmh.profilers through to JMH plugin` — required so `-Pjmh.profilers=gc` actually propagates to the JMH runner (used for the allocation table above).

## What's next (separate PRs)

Tier-D research identified five further levers (~80–135 ns headroom) and one big-swing alternative (MH-composed specialised flat Mapper, ~15–25 ns target). Tier C set up the structural Object[] intermediate they all build on.

### DIFF
--- a/benchmarks/build.gradle.kts
+++ b/benchmarks/build.gradle.kts
@@ -54,4 +54,5 @@ jmh {
     // Optional per-iteration time overrides — `-Pjmh.timeOnIteration=2s -Pjmh.warmupTime=2s` for quick smoke runs.
     (project.findProperty("jmh.timeOnIteration") as String?)?.let { timeOnIteration = it }
     (project.findProperty("jmh.warmupTime") as String?)?.let { warmup = it }
+    (project.findProperty("jmh.profilers") as String?)?.let { profilers = it.split(",") }
 }
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -1083,9 +1083,9 @@ public final class DeepMap {
    * Position-indexed variant of {@link #remapIso}. Precomputes {@code int[]} slot maps and a
    * per-position {@code Iso[]} once at type-pair build time so the hot loop is one array-index +
    * one virtual {@code Iso#to}/{@code Iso#from} per field. Sentinel slot {@code -1} for "no
-   * corresponding source/target field" (placeholder rows where {@code step.sourceName} or
-   * {@code step.targetName} is null) — the value defaults to {@code null}, matching the prior
-   * {@code Map.get(missingKey)} semantics.
+   * corresponding source/target field" (placeholder rows where {@code step.sourceName} or {@code
+   * step.targetName} is null) — the value defaults to {@code null}, matching the prior {@code
+   * Map.get(missingKey)} semantics.
    */
   @SuppressWarnings({ "unchecked", "rawtypes" })
   private static Iso<Object[], Object[]> remapIsoArr(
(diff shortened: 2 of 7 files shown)


eschizoid:
@copilot please review

PR #108, opened 2026-06-14
chore: strip inline FQNs across the repo

eschizoid:
## Summary

Replaced `java.util.X` / `java.lang.reflect.X` / `java.util.stream.X` inline qualified names with simple class references + matching imports in 15 files across main + test sources.

## What was kept

- **Test-fixture FQNs** inside `"""..."""` text blocks (BeanFocusProcessorTest, BridgeProcessorTest etc.) — those strings are Java source code fed to annotation processors, and the FQNs avoid requiring the fixture to declare its own imports. Stripping them would force every fixture to grow an import list.
- **\"contains(\\\"import java.util.X;\\\")\"** assertions in codegen tests — those verify generated code, not test code.
- **Comments mentioning `java.lang.reflect.Array`** etc. — they're prose, not Java references.

## Files touched (15)

**Main:**
- `codegen/BridgeProcessor.java` — `LinkedHashMap`
- `core/Sources.java` — `ArrayList`, `Set`
- `core/conversion/ForwardMapper.java` — `Function`
- `core/mapping/Mapping.java` — `Set`, `TreeSet` (from MS-2 enumTo work)
- `internal/Beans.java` — `reflect.Type`
- `internal/MetadataHolderProbe.java` — `reflect.Method`
- `internal/Records.java` — `reflect.Type`
- `internal/optics/collections/Traversals.java` — `stream.StreamSupport`

**Test:**
- `core test/DeepMapCoverageTest.java` — `Mapping`, `List`
- `core test/DeepMappingTest.java` — `Set`
- `core test/TypedContainerPathTest.java` — `stream.Collectors`
- `core test/beans/PojoOpticsTest.java` — `Map`, `Optional`
- `internal test/ReflectiveStructuralIsoTest.java` — `ArrayList`, `List`
- `lombok test/LombokFocusProcessorTest.java` — `reflect.Modifier`, `function.Function`
- `spring-boot-starter test/TelescopeMapperRegistryTest.java` — `List`

## Test plan

- [x] `./gradlew test` green (46 tasks, all 200+ tests pass)
- [x] No new compile warnings
- [x] No FQN appearing in non-fixture, non-import, non-comment code on a final grep

PR #121, opened 2026-06-15
chore: 1.0 readiness sweep + perf table refresh (Tier C + Tier-D wins)

eschizoid:
Three small things wrapped in for the 1.0 cut.

## 1. Perf table refresh

Fresh 18-row JMH on a quiet rebooted machine, no other workloads. The previously-published tables only reflected the first round of the Object[] structural intermediate change. The Round 1+2 Tier-D wins that shipped in the same PR (Iso.identity() cached singleton with ref-equal short-circuit in `remapIsoArr`, O(N²)→O(N) bean forward name dispatch, and direct LMF-reader binding for holder-annotated records) compounded to **another 11–16% on top** of the structural drop.

Cumulative drop vs the pre-release baseline:

| Tier   | Direction     | Pre-release | Now      | Δ %  |
|--------|---------------|------------:|---------:|-----:|
| flat   | bean → record |      340.13 |   124.04 | -64% |
| flat   | record → bean |      464.27 |   188.08 | -59% |
| nested | bean → record |      501.43 |   204.30 | -59% |
| nested | record → bean |      709.86 |   277.92 | -61% |
| deep   | bean → record |     2024.06 |   802.04 | -60% |
| deep   | record → bean |     2292.43 |  1172.88 | -49% |

Relative gap to MapStruct narrowed from ~78× to **~37× on flat, ~40× on nested, ~15× on deep**.

Codegen also dropped slightly (5.50 → 4.79 ns flat fwd) because the identity-singleton change cascaded into the `@Bridge` composition path. The two prior noisy cells (`flat record→bean codegen ±12.4` and `deep record→bean MapStruct ±34.4`) both stabilised on this run; a new noisy cell (`deep record→bean codegen ±211`) appeared in their place — JMH single-fork variance. Footnoted in the tables.

## 2. 1.0-readiness sweep

Scanned for the things that would be embarrassing on day one:

- `@Deprecated` shims: ✅ none in main source
- JPMS exports: ✅ 5 public packages, 3 qualified-internal — correct
- Examples compile against current local code: ✅ all 5 modules
- TODO / FIXME / XXX markers: ✅ none in main src or user docs
- `UnsupportedOperationException` throws: ✅ all 3 are intentional (multi-source merge has no inverse, forward-only `Mapping.forward` rows have no backward, records have no setters for `Mapper.into`)
- Full test suite: ✅ 46 tasks green

## 3. Mantras enforcement

While there, fixed two mantra violations the sweep surfaced:

- `DeepMap.recursiveDefault` had an inline `io.github.eschizoid.telescope.internal.Records.construct(...)` (FQN regression that escaped the PR #108 sweep) — replaced with an import + simple name.
- `Mapper.java` had a comment referencing a PR number (`PR #90`) inside source — removed per the convention of keeping PR numbers out of source-level docs.

No public API change. No behaviour change.


eschizoid:
@copilot please review

PR #122, opened 2026-06-15
fix(internal): drop cross-module javadoc {@link}s breaking the release pipeline

eschizoid:
**Blocks the v1.0.0 release** — without this, `:internal:javadoc` fails with `reference not found` on five `{@link io.github.eschizoid.telescope.mapping.*}` references, breaking the javadoc jar that JReleaser signs and publishes.

`:internal` qualified-exports its packages only to `:core`. The reverse — `:internal` referencing `:core` types — doesn't compile at the javadoc level because `:internal:javadoc` runs without `:core` on its classpath.

Converts the 5 offending links to plain `{@code}` references with a parenthetical "in `:core`" where the reader benefits from the pointer:

- `NullDefaults.java` — 4 refs (`NullHint.NullStrategy`, `Mapping#toOrElse`, `NullHint`, `NullHint.NullStrategy`)
- `Iso.java` — 1 ref (`NullHint`)

Behaviour unchanged. Verified `:internal:javadoc` passes locally.

This was originally committed (`d5850b9`) on `chore/perf-table-post-tierd` but landed there AFTER PR #121 had already merged, so it never made it onto `main`. Cherry-picked here onto a fresh branch.


eschizoid:
@copilot please review

PR #126, opened 2026-06-15
experiment: all-identity short-circuit (INCONCLUSIVE — keep as research artifact)

eschizoid:
**Experiment 4 of overnight Tier-D autonomous run. Inconclusive result, pushing for transparency.**

## Hypothesis

For pure-auto same-name same-typed mappings, every per-slot `Iso` resolves to the cached `Iso.identity()` singleton. The current `assembleIso` hot loop still pays a per-slot reference-equality check + conditional ternary per field. Predicted ~1 ns per slot saved by detecting the all-identity case at build time and emitting a tighter loop body that just gathers from readers.

## Measured vs the stacked Exp1+Exp2 baseline

| Bench | Stacked | Exp4 | Δ ns | Δ % | Verdict |
|---|---:|---:|---:|---:|---|
| flat fwd | 106.08 ± 1.3 | 105.03 ± 2.0 | -1.05 | -1.0% | wash |
| flat bwd | 173.17 ± 3.0 | 170.63 ± 2.6 | -2.54 | -1.5% | tiny win |
| nested fwd | 145.59 ± 4.7 | 139.20 ± 9.2 | -6.39 | -4.4% | within error band |
| nested bwd | 224.95 ± 4.6 | 233.18 ± 4.5 | **+8.23** | +3.7% | regression |
| deep fwd | 548.73 ± 6.9 | 559.64 ± 12.9 | +10.91 | +2.0% | within error band |
| deep bwd | 873.97 ± 11.9 | 917.22 ± 84.4 | +43.25 | +4.9% | high noise |

## Why it didn't land

Most likely cause: the change introduces a SECOND `Iso.of(...)` call site in `assembleIso` (one for the all-identity branch, one for the general branch). The JIT inline cache at `assembleIso`'s return statement now sees two anonymous-Iso shape variants instead of one. This likely turns a previously monomorphic site into a bimorphic one, costing more inlining headroom than the per-slot ternary check saves.

Lesson: optimizations that branch the assembly path need to factor out into separate emit functions OR keep a single call site. Worth a follow-up if anyone wants to try the variant with a static method factory instead of inline `Iso.of`.

## Status

**Not for merge.** Pushed as a research artifact so the design + measurement is preserved. The conservative recommendation is to land Exp1 (PR #123) and Exp2 (PR #124) separately to main and skip this branch.

## Drafted because

Per the overnight directive — push every experiment for transparency, win or lose.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -1066,19 +1066,100 @@ public final class DeepMap {
     final var tgtNames = tgtRefl.names(target);
     final var tgtHolderReaders = Telescope.holderReadersFor(target, tgtNames);
     final var tgtHolderConstructor = Telescope.holderConstructorFor(target);
-    // Position-indexed structural intermediate: Object[arity] instead of LinkedHashMap. The hash
-    // bucket puts/gets dominated the runtime mapper's allocation profile (~776 B/op on the flat
-    // 5-field bean→record benchmark); arrays drop that to ~80 B/op and let the JIT scalar-replace
-    // monomorphic call sites entirely. Position semantics: slot i of `srcArr` corresponds to
-    // `srcRefl.names(source)[i]`; same for the target side. The remap step bridges the two layouts
-    // via precomputed int[] slot maps, one allocation per type-pair-load.
-    final Iso<S, Object[]> srcReader = srcRefl
-      .structuralIsoArr(source, srcHolderReaders, srcHolderConstructor)
-      .reverse();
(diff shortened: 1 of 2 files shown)


eschizoid:
Closing as a preserved research artifact — never intended for merge. The hypothesis (build-time all-identity short-circuit in assembleIso) measured as a wash-to-regression, and the lesson it surfaced is captured in the PR body: branching the assembly path turns the assembleIso call site bimorphic and costs more inlining headroom than the per-slot ternary saves. Any future attempt should use a single call site / static factory rather than a second inline Iso.of(...). Branch deleted; this PR record preserves the design + measurements.

PR #128, opened 2026-06-16
fix(deepmap): Bug 2 — peel nested-telescope rows before normalising (P0 blocker)

eschizoid:
Bug 2 from `docs/migration-feedback.md`. P0 BLOCKER.

## The migration's diagnosis was wrong

The adopter's feedback attributed the NPE to "boolean accessor" / `Telescope.BeanFieldOptics.lensFor() → methodNameOf()` returning null. Investigation showed that's a red herring — `LambdaIntrospection.methodNameOf` cannot return null (it either returns the method name or throws).

The actual trigger is at `DeepMap.populateIso`: `srcRefl.normalize(row.sourceField())` runs unconditionally **before** the `instanceof` peel that handles nested-telescope sub-shapes (`FromTelescopeTo`, `TelescopeToTelescope`) whose `sourceField()` returns `null` by design. On a bean source side that null flows through `Beans.normalize` → `Beans.propertyOf(null)` and NPEs.

## Fix

**Two complementary changes**:

1. **Structural fix** (`DeepMap.populateIso`): guard the normalise call with a null check on `row.sourceField()`. The downstream branches that consume `srcField` (Drop, TelescopeTo's source side, the default row path) either don't fire on null-source rows or re-read the first-hop name from the telescope itself, so null here is safe.

2. **Defensive guard** (`Beans.propertyOf`): kept as belt-and-suspenders — protects the public surface from any future caller that forgets to peel.

## TDD evidence

| Test | What it pins | Without fix |
|---|---|---|
| `MigrationRegressionTest.nestedSourceTelescopeRowConstructsWithoutNpe` | The actual adopter scenario: `Mapping.to(srcTelescope, tgtAccessor)` with a 2-hop bean source path | **FAILS** ← real reproduction |
| `MigrationRegressionTest.mapperConstructionDoesNotNpeOnBooleanAccessors` | Plain `Telescope.mapper(Order, OrderDto)` on bean fixtures with `boolean` fields | passes (adjacent contract pin) |
| `MigrationRegressionTest.mapperRoundTripsBooleanField` | End-to-end forward through a boolean field | passes |
| `BeansTest.propertyOfNullIsNullSafe` | Defensive guard at the unit level | FAILS without the Beans guard |

The "boolean field" tests stay because they pin the adjacent contract — even though they're not the reproduction of the actual bug, they verify that the migration's other claim ("any class with boolean fields is unusable") is now safe.

## Docs

`docs/migration-feedback.md` updated with the corrected root cause so future Bug-2-style triage doesn't re-inherit the "is*" red herring.

## Mantras

- No public API change.
- No new reflection.
- Lattice-honest — the fix is in the row-loop peel, not in the lattice.

First of a stacked PR series targeting the 8 bugs in `docs/migration-feedback.md` for a `v1.0.1` patch release.

## Review history

Round 1: applied defensive guard only. Reviewer (independent agent) caught that the end-to-end tests passed without the fix — meaning the defensive guard alone was symptom relief, not root cause. Round 2 added the structural fix at the actual call site + a test that genuinely reproduces.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -343,8 +343,11 @@ public final class DeepMap {
       // applyForward can evaluate the predicate before dispatching the inner's effect.
       final Mapping<?, ?> row = (rawRow instanceof Conditional<?, ?> c) ? c.inner() : rawRow;
       // Normalize raw method names per side — record::name stays "name", bean::getName becomes
-      // "name".
-      final var srcField = srcRefl.normalize(row.sourceField());
+      // "name". `sourceField()` returns null on rows whose source is a nested telescope rather
+      // than a flat accessor (FromTelescopeTo, TelescopeToTelescope); those branches re-read the
+      // first hop name from the telescope below and don't consume srcField, so null here is safe.
+      final var rawSrcField = row.sourceField();
+      final var srcField = rawSrcField == null ? null : srcRefl.normalize(rawSrcField);
       // TelescopeTo: flat source accessor → nested target telescope.
       // Soft claim on the source field (from srcAcc) AND the target telescope's first hop name.
       // Multiple rows sharing the same source field or top-level target field all compose.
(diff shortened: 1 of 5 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #129, opened 2026-06-16
fix(beans+telescope): Bug 4 — null intermediate short-circuit

eschizoid:
Bug 4 from `docs/migration-feedback.md`. P1. Stacked on #128.

## The bug

Multi-hop nested bean paths like `Telescope.ofBean(Order).field(::getCustomer).field(::getEmail)` NPE at runtime when `order.getCustomer()` returns null. The migration feedback reported one NPE at `Beans.readProperty` — investigation showed there's actually a second one at `Telescope.read` (the `findFirst()` route).

## Fix

**Two complementary changes, both required**:

1. **`Beans.readProperty(null, name)` returns null** instead of NPEing on `ClassValue.get(null)`. Lets the optic pipeline propagate a null intermediate gracefully — same shape as MapStruct's generated null guard.

2. **`Telescope.read` materialises `getAll` via `.toList()`** instead of `findFirst()`. `Stream.findFirst()` routes through `Optional.of(element)` which NPEs on null. The Traversal can now legitimately surface a null element once the readProperty short-circuit above lands. Materialise the list and decide emptiness explicitly.

## TDD evidence

| Test | Pins | Without fix |
|---|---|---|
| `BeansTest.readPropertyNullPojoIsNullSafe` | Unit-level short-circuit | FAILS (NPE in ClassValue.get) |
| `MigrationRegressionTest.forwardWithNullIntermediateShortCircuitsToNull` | Adopter's exact 2-hop scenario with null intermediate | FAILS (NPE in Optional.of) |
| `MigrationRegressionTest.forwardWithPopulatedIntermediateReadsThrough` | Happy-path regression — short-circuit doesn't suppress non-null values | passes (regression guard) |

Round 1 of review on Bug 2 caught that my initial commit was symptom relief without root cause. Applied the same discipline here — wrote the end-to-end test FIRST, watched it fail in two places, fixed both, watched it pass.

## Mantras

- No public API change.
- Lattice-honest — `Telescope.read` swap is materialise-then-index, semantically identical to the prior `findFirst().orElseThrow()` for non-null elements.

Second in the stacked PR series for the migration-feedback bugs.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -987,7 +987,14 @@ public sealed class Telescope<
     if (optic instanceof final Affine<S, A> affine) {
       return affine.getOption(source).orElseThrow(Telescope::noValue);
     }
-    return optic.getAll(source).findFirst().orElseThrow(Telescope::noValue);
+    // Stream.findFirst() routes through Optional.of(element), which NPEs on null. A Traversal
+    // can legitimately surface null elements when an intermediate hop of a multi-hop bean path
+    // is null (see Bug 4: Beans.readProperty short-circuits to null, the traversal still
+    // produces a one-element [null] stream). Materialise to a List to preserve nulls and decide
+    // emptiness explicitly.
+    final var values = optic.getAll(source).toList();
+    if (values.isEmpty()) throw noValue();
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #130, opened 2026-06-16
fix(beans): Bug 5 — SettersWriter no-op for missing setters

eschizoid:
Bug 5 from `docs/migration-feedback.md`. P1. Stacked on #129.

## Problem

`SettersWriter.buildSetterInvoker` threw `IllegalArgumentException` ("no setter setX") on getter-only properties — read-only computed fields, immutable accessors, derived state. Made `writeBean(SETTERS)` unusable on any class with a single read-only property. MapStruct silently ignores unwritable target fields; telescope should match.

## Fix

Return a no-op `BiConsumer` for the missing-setter case. The construct loop iterates names and calls `setterFor(name).accept(pojo, value)` — with the no-op installed in the per-class `setterInvokers` cache, the property is skipped at write time. The underlying field stays at its JLS default (null/0/false).

## Investigation note

The migration feedback claims the throw fires "at mapper construction time"; investigation showed it actually fires lazily at first `forward()` call (per-name setter resolution is lazy via `ConcurrentHashMap.computeIfAbsent`). The test pins the end-to-end contract from the public `mapper.forward(source)` surface either way.

## TDD evidence

| Test | Pins | Without fix |
|---|---|---|
| `MigrationRegressionTest.Bug5GetterOnlyProperties.getterOnlyTargetPropertiesAreSilentlySkipped` | Adopter scenario: target with `isProcessed()` getter and no `setProcessed()` setter | FAILS (IllegalArgumentException) |
| `BeansTest.settersMissingSetterIsSilentlySkipped` | Unit-level no-op contract | FAILS (was pinning old throw) |

Verified the migration test genuinely reproduces by reverting the fix temporarily — test failed; reverted the revert; test passed.

## Stacking caveat

Bug 5 test must use a source class that ALSO has the `processed` property so DeepMap's same-name strict bijection (Bug 6, not yet fixed) doesn't fire first. Bug 6 is the next PR in the stack and will relax that — at which point we can extend Bug 5's coverage to the asymmetric source-doesn't-have-the-field case too.

## Mantras

- No public API change.
- No new reflection.
- Behaviour now matches MapStruct's "silently skip unwritable target fields" default.

Third in the stacked PR series for the migration-feedback bugs.

### DIFF
--- a/core/src/test/java/io/github/eschizoid/telescope/MapperIntoTest.java
+++ b/core/src/test/java/io/github/eschizoid/telescope/MapperIntoTest.java
@@ -1,5 +1,6 @@
 package io.github.eschizoid.telescope;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
@@ -228,7 +229,7 @@ class MapperIntoTest {
   }
 
   @Nested
-  @DisplayName("Missing setter — clear error names the property")
+  @DisplayName("Missing setter — silently skipped to match SettersWriter / MapStruct @MappingTarget")
   class MissingSetter {
 
     record DtoOnly(String id, String missingOnEntity) {}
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:

PR #131, opened 2026-06-16
fix(deepmap): Bug 6 — nested auto-recursed pairs are lenient

eschizoid:
Bug 6 from `docs/migration-feedback.md`. P1. Stacked on #130. Closes most-impactful adopter pain.

## Problem

`DeepMap.populateIso` enforced strict bijection on EVERY pair encountered during recursive descent — including nested types the user never explicitly configured. A simple `Parent → ParentDto` mapper where the nested `ScoreResponseDto` carries an extra field would throw at construction time. Forces the caller to declare a `constant()` for every unmatched target and a `drop()` for every unmatched source on every nested pair — defeats the point of recursive auto-mapping. The migration's real-world example: `DocumentT` (39 fields) → `IdentificationResponse` (29 fields) required **13 drops + 2 constants** just to satisfy the check.

## Fix

The user only explicitly asked for the TOP-LEVEL pair. Nested pairs follow the principle of least surprise: unmatched target fields stay at JLS defaults, unmatched source fields are silently dropped. Same shape MapStruct uses.

Detection: `inProgress.size() > 1` at the unmatched-field check means we're recursed from `computeAutoIso`, not at the user-facing top level. The existing telescope-row branch (`!telescopeFixups.isEmpty()`) already installed the right placeholder — we just extend its gate to fire on nested calls too. Strict top-level behaviour is preserved.

Both the target-side and source-side throws are guarded; both reuse the existing placeholder paths (`placeholderIsoFor` for target, `NULLING_ISO` for source).

## TDD evidence

| Test | Pins | Without fix |
|---|---|---|
| `MigrationRegressionTest.Bug6...nestedAsymmetricPairConstructsLeniently` | Adopter's exact `Parent → ParentDto` scenario; nested `ScoreResponseDto` has extra `matchingStatus` | FAILS (IllegalStateException) |
| `MigrationRegressionTest.Bug6...topLevelAsymmetricPairStillThrows` | Preserved strictness at the user-facing top level | passes (unchanged) |

## Adjacent test update

`DeepMappingTest.oneArgDropOnNestedSourceDoesNotReachNestedPair` was pinning the OLD strict-on-nested behaviour. Updated to pin the new lenient contract — the mapper constructs, the nested pair silently drops the unmapped field.

## Mantras

- No public API change.
- Lattice-honest — reuses the existing placeholder paths.
- Strictness preserved at the top level (where missing fields ARE a user configuration mistake).

Fourth in the stacked PR series for the migration-feedback bugs.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -315,183 +315,242 @@ public final class DeepMap {
     }
     cache.put(key, null); // reserve slot so cycle-re-entry short-circuits
     inProgress.push(key);
+    // The pop lives in finally (at the end of this method) so a thrown IllegalStateException —
+    // the strict-bijection guard at the top-level pair — doesn't corrupt the `inProgress` stack.
+    // `inProgress.size() > 1` is the lenient-nested gate; a leaked pre-pop frame would silently
+    // flip future top-level calls in the same cache into "nested" mode.
+    try {
+      final var srcRefl = pickReflective(source, beanRefl);
+      final var tgtRefl = pickReflective(target, beanRefl);
 
-    final var srcRefl = pickReflective(source, beanRefl);
-    final var tgtRefl = pickReflective(target, beanRefl);
+      final var byTargetName = new LinkedHashMap<String, FieldStep>();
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:

PR #132, opened 2026-06-16
fix(beans): Bug 8 — null-guard primitive setters (defence-in-depth)

eschizoid:
Bug 8 from `docs/migration-feedback.md`. P1. Stacked on #131.

## Status: preempted in practice, locked as contract

Bug 6's `placeholderIsoFor` short-circuits unmatched primitive target fields to their JLS default before any value reaches the SettersWriter — so the LMF-built primitive setter never sees a null on the current forward path. This PR lands defence-in-depth at the setter level to lock the contract for future code paths that may legitimately deliver null to a primitive setter (e.g. the forthcoming autoboxing relaxation in Bug 3).

## Fix

When `buildSetterInvoker` resolves a setter whose declared parameter is a primitive, wrap the LMF-built `BiConsumer` to skip on null. The field stays at its JLS default — same semantics as MapStruct's generated null-guard on primitive assignments. Non-primitive setters return the raw LMF dispatcher unchanged (no per-call cost).

## TDD evidence

Both regression tests pass with the current architecture even WITHOUT this guard (Bug 6 preempts). They stay as forward-compatible contract pins:

- `Bug8...primitiveSetterReceivesJlsDefaultOnNullValue` — nested target with `int count`; source doesn't have `count`.
- `Bug8...nullBoxedSourceReachesBoxedSetterWithoutNpe` — matched boxed→boxed pair carrying null.

The defence-in-depth wrap doesn't change observable behaviour on these tests — it just locks the contract.

## Mantras

- No public API change.
- No new reflection.
- Behaviour now matches MapStruct's null-guard on primitive assignments.

Fifth in the stacked PR series.

### DIFF
--- a/core/src/test/java/io/github/eschizoid/telescope/MigrationRegressionTest.java
+++ b/core/src/test/java/io/github/eschizoid/telescope/MigrationRegressionTest.java
@@ -352,6 +352,180 @@ class MigrationRegressionTest {
   }
 
   @Nested
+  @DisplayName("Bug 8 — SettersWriter NPE on null primitive value")
+  class Bug8PrimitiveSetterNullNpe {
+
+    public static class Source {
+
+      private String name;
+
+      // No `count` property — after Bug 6 lenient nested mode, valueByName returns null for the
+      // target's `count` field.
+
+      public String getName() {
+        return name;
+      }
+
+      public void setName(final String name) {
+        this.name = name;
+      }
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:

PR #133, opened 2026-06-16
fix(deepmap+beans): Bug 7 — refuse to bean-decompose JDK collection subtypes

eschizoid:
Bug 7 from `docs/migration-feedback.md`. P1. Stacked on #132.

## Problem

A bean graph containing a class that extends `ArrayList` / `HashMap` / any other JDK collection caused DeepMap's auto-recursion to try bean-decomposing the wrapper. The scan surfaced inherited `isEmpty()` etc. as "getters", then LMF-bound them — fails with `LambdaConversionException: Invalid caller: java.util.ArrayList` because `java.base` modules don't grant private lookup to application code. Common in legacy enterprise codebases that have a typed wrapper around a collection.

## Fix

**Two-layer**:

1. **`DeepMap.isReflectable`**: refuses any `Collection` or `Map` subtype. Container values still flow through the parameterised-type detector (`ContainerShape.of` handles concrete `List<E>`, `Set<E>`, `Map<K,V>`, `Optional<E>`); subtypes that look like a bean type are now correctly treated as scalars (pass-through by reference, same shape MapStruct uses for collection wrappers).
2. **`Beans.scanGetters`**: defence-in-depth — skip inherited methods declared in platform modules (`java.*`). Catches any future caller that scans a JDK-derived class directly without going through `isReflectable` first.

## TDD evidence

| Test | Pins | Behaviour |
|---|---|---|
| `Bug7...collectionSubtypeIsPassThrough` | Same `ImageUrls` on both sides — happy path | Passes either way (autoIso same-type shortcut) |
| `Bug7...differentCollectionSubtypesRefuseBeanRecursion` | Adversarial — `ImageUrls` vs `ImageUrlsAlt` subtypes force the bean-recursion path | Drills the cause chain for any LMF binding failure against `java.util.ArrayList` |

The adversarial test surfaces the bug condition; if the recursion ever tries to LMF-bind a java.base method, the assertion fires with a clear message.

## Mantras

- No public API change.
- No new reflection — uses the existing `Class.getModule()` lookup.
- Lattice-honest — Collection/Map subtypes are now correctly scalar values.

Sixth in the stacked PR series.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -829,6 +829,24 @@ public final class DeepMap {
     // (a) Same generic type → identity Iso.
     if (srcType.equals(tgtType)) return Iso.identity();
 
+    // (a.2) Collection / Map subtype pair (Bug 7). Common in legacy bean codebases: `class
+    //       ImageUrls extends ArrayList<ImageUrl>` and `class ImageUrlsBO extends
+    // ArrayList<ImageUrl>`
+    //       on opposite sides. Neither is identity (different concrete classes), neither is a
+    //       parameterised raw `List` / `Map` (so ContainerShape skips them), and trying to bean-
+    //       decompose would hit `MethodHandles.privateLookupIn(ArrayList.class)` rejection. Copy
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:

PR #134, opened 2026-06-16
fix(deepmap): Bug 3 — primitive ↔ wrapper auto-iso with JLS-default null guard

eschizoid:
Bug 3 from `docs/migration-feedback.md`. P1. Stacked on #133.

## Problem

DeepMap's shape-compatibility check rejected primitive ↔ wrapper pairs at populateIso time with `incompatible source/target shapes`. The JavaBeans spec treats `boolean isFoo()` and `Boolean getFoo()` as the same property semantically. MapStruct silently autoboxes both directions and substitutes the JLS default for null-wrapper-to-primitive assignments. Telescope didn't — making any class with the common `int`/`Integer` or `boolean`/`Boolean` mix unmappable.

## Fix

New branch (a.1) in `computeAutoIso` between the identity short-circuit and the bean-recursion gate. When the pair is primitive ↔ wrapper (in either direction), build a null-safe `Iso` that substitutes the primitive default for null. Three helpers added: `isPrimitiveWrapperPair`, `wrap`, `primitiveWrapperIso`. Reuses the existing `primitiveDefault` lookup.

## TDD evidence

| Test | Pins | Without fix |
|---|---|---|
| `Bug3...primitiveWrapperPairsAreAutoBoxed` | Adopter scenario — `Source(boolean, Integer)` ↔ `Target(Boolean, int)` round-trips correctly | FAILS at populateIso (shape mismatch) |
| `Bug3...nullBoxedToPrimitiveUsesJlsDefault` | Null `Integer` → primitive int substitutes JLS default 0 | FAILS (shape mismatch, never reaches setter) |

The Bug 8 setter-level null-guard and this null-safe Iso are belt-and-suspenders: either alone prevents the unboxing NPE; together the contract is locked at both layers.

## Mantras

- No public API change.
- Reuses existing `primitiveDefault`.
- Matches MapStruct behaviour for primitive ↔ wrapper pairs.

Seventh in the stacked PR series.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -830,6 +830,18 @@ public final class DeepMap {
     // (a) Same generic type → identity Iso.
     if (srcType.equals(tgtType)) return Iso.identity();
 
+    // (a.1) Primitive ↔ wrapper pair → autobox / unbox via JLS-default-safe Iso. Forward
+    // unboxes (null-safe — substitutes the primitive default to avoid NPE on the wrapper-to-
+    // primitive setter); backward boxes via the wrapper's static valueOf. Matches MapStruct's
+    // behaviour for these pairs (it silently autoboxes and uses 0/false/etc. for null).
+    if (
+      srcType instanceof Class<?> srcClsAB &&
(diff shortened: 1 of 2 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:

PR #135, opened 2026-06-16
fix(bridgeprocessor): Bug 1 — handle unresolvable @Bridge target

eschizoid:
Bug 1 from `docs/migration-feedback.md`. P1. Stacked on #134.

**Last in the migration-feedback PR stack — all 8 bugs fixed.**

## Problem

When `@Bridge(value = SomeClass.class)` referenced a class that the annotated compilation unit couldn't see (different Maven module with no dependency edge), javac's `AnnotationValue.getValue()` returned the target FQN as a `String` instead of a `TypeMirror`. The processor cast unconditionally to `TypeMirror`, which crashed the build with:

```
ClassCastException: class java.lang.String cannot be cast to class javax.lang.model.type.TypeMirror
```

No file, no line, no hint at what was wrong.

## Fix

Rename `targetTypeFromMirror` → `rawTargetValueFromMirror` and have it return `Object` (the documented union of `TypeMirror` / `String` / `null`). Callers dispatch on `instanceof`. The primary caller surfaces a precise diagnostic:

```
@Bridge target 'some.x.Foo' is not resolvable from this compilation unit — annotate the other side, or add the dependency.
```

The sealed-roots caller (which iterates per-case bridges) silently skips non-`TypeMirror` values — matches the prior "continue on non-DECLARED" semantics; subsequent code emits a more specific diagnostic for that case anyway.

## TDD evidence

`BridgeProcessorTest.unresolvableTargetEmitsPreciseDiagnostic` — new test in the existing `Rejections` nest. Compiles a `@Bridge` with an unresolvable target FQN. Without the fix the processor throws `ClassCastException`; with the fix the compilation reports the precise diagnostic.

## Mantras

- No public API change.
- No new reflection.
- Diagnostic message names the missing FQN so the user knows exactly what to add to their module dependencies.

## Stack summary

| # | PR | Bug | Severity |
|---|---|---|---|
| 1 | [#128](https://github.com/eschizoid/telescope/pull/128) | Bug 2: boolean accessor NPE (real root cause: nested-telescope row `sourceField()` null) | **P0** |
| 2 | [#129](https://github.com/eschizoid/telescope/pull/129) | Bug 4: null intermediate NPE + `Telescope.read` toList swap | P1 |
| 3 | [#130](https://github.com/eschizoid/telescope/pull/130) | Bug 5: SettersWriter no-op for missing setters | P1 |
| 4 | [#131](https://github.com/eschizoid/telescope/pull/131) | Bug 6: lenient nested auto-recursed pairs | P1 |
| 5 | [#132](https://github.com/eschizoid/telescope/pull/132) | Bug 8: primitive setter null-guard (defence-in-depth) | P1 |
| 6 | [#133](https://github.com/eschizoid/telescope/pull/133) | Bug 7: refuse to bean-decompose JDK collection subtypes | P1 |
| 7 | [#134](https://github.com/eschizoid/telescope/pull/134) | Bug 3: primitive ↔ wrapper auto-iso | P1 |
| 8 | **THIS** | Bug 1: BridgeProcessor unresolvable target | P1 |

Eighth and final PR in the migration-feedback stack.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -162,7 +162,22 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
       if (bridges.size() > 1) multiTargetSources.add(sourceFq);
 
       for (final var bridgeAm : bridges) {
-        final var target = targetTypeFromMirror(bridgeAm);
+        final var rawValue = rawTargetValueFromMirror(bridgeAm);
+        // When the @Bridge target class lives in a module that the annotated element's
+        // compilation unit can't see, javac's AnnotationValue.getValue() returns the target FQN
+        // as a String instead of a TypeMirror. Surface a precise diagnostic instead of letting
+        // a downstream cast crash with an unhelpful ClassCastException.
+        if (rawValue instanceof String fqn) {
+          error(
+            element,
+            "@Bridge target '" +
+              fqn +
+              "' is not resolvable from this compilation unit — annotate the other side, " +
+              "or add the dependency."
+          );
+          continue;
+        }
+        final var target = rawValue instanceof TypeMirror tm ? tm : null;
         if (target == null || target.getKind() != TypeKind.DECLARED) {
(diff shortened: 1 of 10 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #153, opened 2026-06-17
docs(migration-feedback): record Bugs 12 + 13 — codegen path lacks null parity

eschizoid:
## Summary

Adopter-driven feedback (Round 2). Two new P1 bugs recorded against the `@BeanFocus` codegen path; **doc-only** — no code change yet.

Both bugs are the **same pattern**: the codegen-emitted `*FieldOptics` class lacks null-handling parity with the runtime `Beans` path that was fixed in v1.0.1.

- **Bug 12** — `@BeanFocus` generated `construct()` doesn't null-guard `Integer → int` unbox. Sibling of Bug 3; Bug 3's fix landed only for the runtime `DeepMap` path, not the codegen template in `AbstractTelescopeProcessor#emitBeanConstruct` (codegen/.../AbstractTelescopeProcessor.java:952).
- **Bug 13** — `@BeanFocus` generated `FieldOptics.get()` NPEs on null intermediates. Sibling of Bug 4; Bug 4 added `if (pojo == null) return null;` to `Beans.readProperty`, but the generated `*FieldOptics` dispatch bypasses that path entirely.

Both fixes are template-level: single file (`AbstractTelescopeProcessor.java`), two emitter methods. `BeanFocusProcessor` and `LombokFocusProcessor` both route through these templates, so one fix covers both.

## Test plan

- [x] `./gradlew spotlessCheck` — clean
- [ ] Follow-up PR will land the actual fixes + reproducer tests in `:codegen` and `:lombok` testFixtures

### DIFF
--- a/docs/migration-feedback.md
+++ b/docs/migration-feedback.md
@@ -579,6 +579,148 @@ EnumMap with its key class. Same applies to any user-defined Collection/Map subc
 
 ---
 
+### 12. `@BeanFocus` codegen: no null-guard on `Integer → int` in generated `construct()`
+
+**Severity:** P1 — NPE at runtime when source has a nullable `Integer` field and target has a primitive `int`.
+
+**Version:** Found in 1.0.1 — sibling of Bug 3 (primitive ↔ wrapper). Bug 3 was fixed for the runtime `DeepMap` path,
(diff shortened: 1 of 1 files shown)


eschizoid:
Folded into #154 — that PR now ships the bug entries and the actual fixes in one go, so this standalone doc PR is redundant. Closing.

PR #154, opened 2026-06-17
fix(codegen,internal): null-safety parity for @BeanFocus codegen path

eschizoid:
## Summary

Two adopter-reported regressions where the `@BeanFocus` codegen path diverged from the runtime path's null-handling. Both are tracked in `docs/migration-feedback.md` and pinned by new sibling slots in `MigrationRegressionTest`.

### Reader path — null intermediate in a multi-hop bean path

Runtime `Beans.readProperty` short-circuits on a null pojo. The codegen-emitted `*FieldOptics` constants are `Telescope.lens(MethodRef, …)` values whose `Lens.get(null)` dispatches the captured method reference on a null receiver and NPEs. When a composed Telescope is projected as a Traversal (via `Lens.getAll` / `Traversal.then` flatMap), that NPE surfaces inside `DeepMap.overrideTargetField`'s read.

**Fix at the Traversal projection:** `Lens#getAll` returns `Stream.empty()` when source is null instead of dispatching `get(null)`. `Lens#getOption` gains the same null guard with `Optional.ofNullable` on the result. Atomic `Lens#get(null)` still NPEs when the user's getter does — strict-lens semantics for direct gets are unchanged.

**Companion fix in `DeepMap.overrideTargetField`:** switch from `read()` to `find().orElse(null)` so the now-empty stream resolves to a null target value instead of `NoSuchElementException`.

### Rebuild path — null boxed value into a primitive setter

Runtime `SettersWriter` substitutes the JLS default when binding a null boxed value to a primitive setter; the codegen-emitted `construct()` did not. `AbstractTelescopeProcessor#emitBeanConstruct` emitted

```java
c.setX((Integer) values.apply("x"))
```

which NPEs on the implicit `Integer → int` unbox when the value is null. Refactored to extract a single `valueExprForProp` helper that emits the `instanceof`-pattern null-guard form for primitive-typed properties:

```java
c.setX(values.apply("x") instanceof Integer __v ? __v : 0)
```

…and keeps the plain cast for reference-typed properties (null is a legal argument to a reference-typed setter). A small `primitiveDefaultLiteral` table covers all 8 primitive JLS defaults.

## Test plan

- [x] `./gradlew check` — green across all modules (122 tests across `:core`, `:codegen`, `:lombok`, `:internal`)
- [x] Two new nested classes in `MigrationRegressionTest` — sibling slots to the existing runtime-path coverage; both initially RED, then GREEN after the corresponding fix
- [x] Existing `BeanFocusProcessorTest` snapshot for the int-typed property setter updated to assert the new `instanceof`-guarded form
- [x] `./gradlew spotlessCheck` clean


eschizoid:
@copilot please review

eschizoid:
@copilot please review — branch has 4 commit rounds of review-driven follow-ups since the initial mention; current head is 30793a3

PR #155, opened 2026-06-18
test(internal): meaningful coverage hardening for Beans / Records / Reflective

eschizoid:
## Summary

Coverage hardening for the three `:internal` files that sat below 60% on the codecov dashboard. Each new test pins a real contract or edge-case behaviour, not a line-touch — one deep test per gap as the project's meaningful-tests rule prescribes.

## Coverage movement

| File | Before | After | Delta |
|---|---|---|---|
| `Beans.java` | 59.96% | **73.09%** | +13 pp |
| `Records.java` | 56.84% | **84.21%** | +27 pp — now green |
| `Reflective.java` | 57.33% | **93.33%** | +36 pp — now green |

### What's still missed in `Beans.java`

The 127 residual missed lines fall into four buckets:

- **Hibernate lazy-initializer path** (`buildHibernateLazyInitializerFn` / `buildHibernatePersistentClassFn`) — only fires when Hibernate is on the classpath. Out of scope without adding Hibernate as a test dep.
- **JPMS-opens diagnostic branches** in `privateLookupOrThrow` — the `IllegalAccessException` catch can't be reached from inside the same module without a deliberately-broken module-info.
- **LMF writer error / recovery paths** in `buildSetterInvoker` / `buildGetterInvokerMap` — `LambdaMetafactory.metafactory` failure modes that don't surface for well-formed JavaBeans.
- **Cache hot path** (`SETTER_INVOKERS.computeIfAbsent` at line 412) — exercised transitively by every `writeBeanProperty` test, but no dedicated test pins the "second call hits the cache" contract because none of the cached-invoker fields are package-accessible.

### Beans (additions to `BeansTest`)

- `writeBeanProperty` round-trips reference- and primitive-typed setters, silently no-ops getter-only properties (the MapStruct-`@MappingTarget` contract), routes through the declaring class for inherited setters.
- `fieldLens(String)` class-deferred lens covers get / set / modify, null-source guards on set + modify, subtype rebuild via the call-time-resolved writer, and the documented primitive null substitution (`set(s, null)` → JLS default for primitive components).

### Records (new `RecordsTest`)

- `read` / `with` / `construct` / `componentNames` / `componentType` contract surface, including the unknown-name / non-record diagnostics and the null-source guards.
- `fieldLens(String)` class-deferred + `fieldLens(Class, String)` class-aware variants; the latter fails fast at lens build time when the component name is unknown.
- `info()` per-class memoisation via direct `assertSame` identity check plus cross-class `assertNotSame` cache-bleed pin.

### Reflective (new `ReflectiveSurfaceTest`)

- `of(Class)` record-vs-bean dispatch through the `RECORDS` / `BEANS` singletons.
- Delegate wrappers (`names` / `genericType` / `read` / `construct` / `normalize`) routed end-to-end, including the two-leading-caps `normalize` rule for `getURL` → `URL`.
- `beansWithHints` hint-map consultation pinned via a tracking `HashMap` subclass that records `get()` probes — proves the hint path is taken before fallthrough to `autoWriter`.
- `defaultFactory` invocation behaviour pinned at per-call (no per-instance memoisation at this layer — see source comment below).
- `positionalBuilder` fast paths for both record (canonical-ctor cache) and bean (autoWriter) shapes; `positionalReaders` exposes per-slot LMF readers.
- Holder-aware path: a non-null `holderConstructor` takes priority over the reflective build, with an `AtomicInteger` counter confirming exactly-one invocation per `builder.apply`.

## Source change

- **`Reflective.beansWithHints` javadoc fix.** The doc previously claimed the default factory was "consulted lazily on first encounter with each not-explicitly-hinted target, then cached". The implementation never cached at this layer — per-class LMF reuse happens one tier down in `Beans.autoWriter`. Test investigation surfaced the mismatch; the doc now matches the actual stateless-resolution behaviour and points readers at the right cache.

## Test plan

- [x] `./gradlew :internal:test` — all tests pass
- [x] `./gradlew :internal:jacocoTestReport` — confirms the coverage movement above
- [x] Spotless clean

### DIFF
--- a/internal/src/main/java/io/github/eschizoid/telescope/internal/Reflective.java
+++ b/internal/src/main/java/io/github/eschizoid/telescope/internal/Reflective.java
@@ -73,9 +73,12 @@ public record Reflective(
    * (when non-null) and ultimately to {@link Beans#autoWriter}. Used by {@code DeepMap} when the
    * user supplies {@code writeBean(targetClass, strategy)} rows and/or a single {@code
    * writeBeans(strategy)} default — the per-class hint map is keyed on target class and provides a
-   * pre-instantiated {@link Beans.BeanWriter}; the default factory is consulted lazily on first
-   * encounter with each not-explicitly-hinted target, then cached, so a default-strategy
-   * incompatible with a particular target only throws when that target is actually constructed.
+   * pre-instantiated {@link Beans.BeanWriter}; the default factory is consulted on every
+   * not-explicitly-hinted call, so a default-strategy incompatible with a particular target only
+   * throws when that target is actually constructed. Per-class LMF reuse happens one layer down in
+   * {@link Beans#autoWriter} (and the underlying {@code SETTER_INVOKERS} / {@code GETTER_INVOKERS}
+   * caches) — this layer is intentionally stateless so a factory swap is observable on the next
+   * call.
    */
   public static Reflective beansWithHints(
     final Map<Class<?>, Beans.BeanWriter<?>> hints,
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #156, opened 2026-06-18
test(lombok): drive LombokFocusProcessor through ProcessorHarness — 0% → 91%

eschizoid:
## Summary

Lifts `LombokFocusProcessor.java` from **0% → 90.91%** coverage. The codecov dashboard reported the file as fully uncovered (22 tracked lines, all missed) despite a working `LombokFocusProcessorTest` — the existing test only verified the *generated code* (which runs in the JUnit JVM where JaCoCo can see it). The processor itself ran inside javac's `compileTestJava` pipeline, opaque to coverage instrumentation.

## Fix

Drive `LombokFocusProcessor#process` through `:codegen`'s in-memory `ProcessorHarness` from inside a JUnit test. `ToolProvider.getSystemJavaCompiler()` runs the processor in the same VM the test runner started, so JaCoCo instruments it normally.

The previous file-based-fixture approach can't be ported to the harness because Lombok's own AST-patching processor doesn't install in the in-process `JavaCompiler.CompilationTask` flow — the round-deferred gotcha already documented on `LombokFocusProcessor`. Workaround: each fixture in the new test supplies explicit getters / setters / `builder()` — Lombok's synthesised members aren't needed because they're written by hand. The processor only needs `lombok.Data` / `@Value` / `@Builder` resolvable as a type element (Lombok jar on the test classpath) and a readable bean shape from `Elements#getAllMembers` — both satisfied without Lombok firing.

## Build wiring

- `lombok.jar` moves from `testCompileOnly` to `testImplementation` so its annotation classes land on the test runtime classpath the harness reads. Lombok's processor still only fires when listed under `testAnnotationProcessor` (unchanged).
- Added `testImplementation(testFixtures(project(":codegen")))` so the harness types are reachable.

## Coverage movement

| File | Before | After |
|---|---|---|
| `LombokFocusProcessor.java` | **0.00%** | **90.91%** (20/22 lines) |

The 2 residual missed lines are the rare last-resort emit at `processingOver()` when a Lombok-annotated target never became readable — exercised only when the test harness deliberately races Lombok's AST patches, which the existing `LombokFocusProcessorTest` (file-based, runs in the Gradle pipeline) already covers.

## What's pinned

Seven tests across two `@Nested` classes:

- **`@Data` / `@Value + @Builder` / `@Builder` triggers** each emit a `<Pojo>Telescope` navigator with the expected property name in the generated source.
- **Bare `@Value`** (no setters, no builder) is rejected with a no-write-strategy diagnostic that names the offending class — pins the negative-case contract on `AbstractTelescopeProcessor`'s emit-rejection branch.
- **Element-kind filter** — `@Data` on an interface compiles cleanly with no navigator emitted (gatekeeper at `element.getKind() != ElementKind.CLASS`).
- **Multi-property `@Data`** — navigator carries one method per property.
- **Unannotated source** — processor stays a no-op (no spurious navigators).

## Test plan

- [x] `./gradlew :lombok:test` — 7 new harness tests + 16 existing file-based tests all pass
- [x] `./gradlew :lombok:jacocoTestReport` — confirms 0% → 90.91%
- [x] Spotless clean

### DIFF
--- a/lombok/build.gradle.kts
+++ b/lombok/build.gradle.kts
@@ -57,10 +57,16 @@ dependencies {
 
     testImplementation(project(":core"))
     testImplementation(project(":codegen"))
+    testImplementation(testFixtures(project(":codegen")))
     testImplementation(platform(libs.junitBom))
     testImplementation(libs.junitJupiter)
     testRuntimeOnly(libs.junitPlatformLauncher)
-    testCompileOnly(libs.lombok)
+    // Lombok jar carries the @Data/@Value/@Builder annotation classes alongside the AST-patching
+    // processor. testImplementation puts the annotation classes on both the compile classpath (for
+    // the file-based fixtures in fixtures/) and the test runtime classpath (so the in-memory
+    // ProcessorHarness can resolve `lombok.Data` via `Elements#getTypeElement`). The Lombok
+    // processor itself only fires when listed under testAnnotationProcessor below.
+    testImplementation(libs.lombok)
     testAnnotationProcessor(libs.lombok)
     testAnnotationProcessor(project(":core"))
     testAnnotationProcessor(project(":codegen"))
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

PR #158, opened 2026-06-18
fix(core): Bug #14 null-safety parity + Enh #10/#11 — bundled v1.0.5 follow-ups

eschizoid:
## Summary

Bundled v1.0.5 follow-up addressing **one P1 regression + three Round-3 adopter enhancements** in a single PR. Replaces the original Bug `#14` hotfix scope with the comprehensive set requested for v1.0.5.

### 🐛 Bug `#14` — `DeepMap.applyForward` null-safety parity on `TelescopeToTelescope` path

PR #154 fixed the null-intermediate `NoSuchElementException` at `DeepMap.overrideTargetField:835` (the `FromTelescopeTo` path) but missed the sibling site at `DeepMap.applyForward:745` (the `TelescopeToTelescope` BROADCAST path). One-line patch: `srcT.read(s)` → `srcT.find(s).orElse(null)`. Codebase audit (see commit log) confirmed only the one missed site — lines 779 and 807 are backward overlays (read from target, immune).

Regression tests added in `MigrationRegressionTest.TelescopeToTelescopeForwardLeniency`:
- Null-intermediate path over `to(srcTelescope, tgtTelescope)`
- `.whenPresent` / `Optional.empty` Affine-miss path (symmetric with the existing `OverrideTargetFieldLeniency` sibling)

### ✅ Enh `#11` — verified `@SafeVarargs` already in place on `Telescope.merge`

Round-3 feedback proposed adding `Telescope.mergeBuilder(...)` to dodge the `@SuppressWarnings("unchecked")` on `Telescope.merge(...)` call sites. **Declined the parallel-API shape, verified the underlying fix is already present.** `Telescope.merge:761` carries both `@SafeVarargs` and `@SuppressWarnings("varargs")`; the parallel `Telescope.all(Edit<S>...)` entry point is similarly annotated. Adopters who still see the warning should look one layer down in their own chain.

No code change in this PR — purely doc clarification. Marked **Verified (v1.0.5)** in `docs/migration-feedback.md`.

### 🆕 Enh `#10` — auto-discover `@Bridge`-generated bridges in `Telescope.mapperForward()`

The highest-impact Round-3 enhancement. When `Telescope.mapperForward(A.class, B.class)` is invoked with **no per-field rows** AND a sibling `<A>Bridge.BRIDGE` (or `<A>To<B>Bridge.BRIDGE`) constant is on the classpath, the forward direction routes through it directly. Closes the "mapping defined twice" pain: adopters declaring `@Bridge(renames = {…})` no longer have to restate those renames in `to()` rows on the mapper bean.

**Implementation:**
- New `internal/.../BridgeHolderProbe.java` — mirrors the existing `MetadataHolderProbe` pattern: `ClassValue`-cached, presence-and-absence both memoised, per-`(sourceClass, targetClass)` keyed via a nested `ConcurrentHashMap`. Probes both emission shapes the `BridgeProcessor` produces (long form `<A>To<B>Bridge` first for determinism; short form `<A>Bridge` second with parameterised-type verification on the `BRIDGE` field's generic args).
- `Telescope.mapperForward(source, target, steps)` gains a `steps.length == 0` short-circuit that consults the probe; when found, `ForwardMapper.create(bridge::read, …)`. Per-field rows (`steps.length > 0`) opt out of auto-discovery as the explicit-overrides escape hatch.

**Scope decisions:**
- MVP: auto-discovery only kicks in when no rows are supplied. Compose-with-overrides (bridge + per-field `to()` rows on top) is a future enhancement; adopters needing overrides fall back to explicit rows today.
- `BridgeRef.bridge` is typed as `Object` inside `:internal` so the probe stays compile-time-oblivious to `:core`'s `Telescope` type.

**Tests in `MigrationRegressionTest.MapperForwardAutoDiscoversBridge`:**
- `record→record`: produces the same value as direct `BridgeRecABridge.BRIDGE.read(src)`
- `pojo→pojo`: same shape on a Lombok-style POJO pair
- Plain POJO pair without `@Bridge`: falls through to `DeepMap.resolveForward`

### 📋 Enh `#12` / Enh `#13` (deferred to v1.0.6)

Recorded in `docs/migration-feedback.md` with full analysis:
- **Enh `#12`** — `Telescope.fromMap()` nested map → sub-POJO composition. Real use case (DynamoDB AttributeValue, JSON payloads), but investigate-first whether existing API already composes with a single cast at the lambda boundary.
- **Enh `#13`** — note: in the table this maps to the Bridge auto-discovery itself; tracked as Enh `#10` in this PR's body for clarity with the adopter's feedback numbering.

## Reviewer convergence

Two prior review rounds on the Bug `#14` + Enh `#11` scope have converged TP-quality:

- **Round 1** — 4 reviewers (mantra / code / comment-analyzer / silent-failure-hunter). Comment-analyzer P0 (Enh `#11` doc factual error) + 3 P1 + 4 P2 all addressed in commit `6438752`.
- **Round 2** — all 4 reviewers say clean. 3 cosmetic P2s addressed in commit `d6d6e37`.

The Enh `#10` implementation lands fresh and needs its own review round.

## Test plan

- [x] `./gradlew :core:test --tests "*MigrationRegressionTest*"` — all 80+ adopter-feedback regressions pass
- [x] Spotless clean
- [x] 4-reviewer round on the bundled scope (Enh `#10` specifically)

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -742,11 +742,10 @@ public final class DeepMap {
           );
           t = tgtT.updateIndexed(t, (i, _ignored) -> values.get(i));
         } else {
-          // Lenient: mirrors the overrideTargetField shape — when the source path resolves to an
-          // empty focus (null intermediate in a chained bean read, or an Affine miss further down
-          // the path), write null to the target field rather than aborting the mapper with
-          // NoSuchElementException. Downstream type-default handling — where configured — takes
-          // over from there.
+          // Lenient: when the source path resolves to an empty focus (null intermediate in a
+          // chained bean read, or an Affine miss further down the path), write null to the target
+          // field rather than throwing. Downstream type-default handling — where configured —
+          // takes over from there.
           t = tgtT.set(t, srcT.find(s).orElse(null));
         }
       } else if (fx instanceof Constant<?, ?, ?> r) {
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -742,7 +742,12 @@ public final class DeepMap {
           );
           t = tgtT.updateIndexed(t, (i, _ignored) -> values.get(i));
         } else {
-          t = tgtT.set(t, srcT.read(s));
+          // Lenient: mirrors the overrideTargetField shape — when the source path resolves to an
+          // empty focus (null intermediate in a chained bean read, or an Affine miss further down
+          // the path), write null to the target field rather than aborting the mapper with
+          // NoSuchElementException. Downstream type-default handling — where configured — takes
+          // over from there.
+          t = tgtT.set(t, srcT.find(s).orElse(null));
         }
       } else if (fx instanceof Constant<?, ?, ?> r) {
         final var tgtT = (Telescope<T, Object>) r.targetTelescope();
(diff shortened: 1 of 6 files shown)


eschizoid:
@copilot please review

PR #180, opened 2026-06-21
perf: flatten effectful traversal accumulator to O(n)

eschizoid:
## What

`Traversal#modifyF` — the engine behind every `update{Async,Optional,Either,Validated}` — folded each focused element's effectful result into a **fresh `ArrayList` copy per step**. Over an n-element traversal that's `1+2+…+n` copies → **O(n²)** total. Invisible at tiny fan-outs, but it explodes on list-heavy traversals and generates enough garbage to stall on GC.

## Fix

Replace the growing-list fold with a **pre-sized positional `Object[]`**: each element writes its own slot exactly once, then a single `modify` pass walks a cursor in order. O(n²) → **O(n)**.

The `map2` fold still threads the effect (sequencing, short-circuit, error accumulation, empty-propagation) — only the accumulator shape changed. The fold's dependency chain (`acc[i]` derives from `acc[i-1]`) serializes the cheap slot writes in index order **even for `CompletableFuture`**, so the array fills under a happens-before chain while the futures themselves still run concurrently. Single-focus optics (`Affine`/`Lens`/`Prism`/`Iso`) already override `modifyF` with an O(1) path and are untouched.

## Evidence

New `EffectfulTraversalBenchmark` sweeps traversal size 16 → 1024 across `updateOptional` / `updateValidated`, with `updatePure` as the O(n) control. Per-element cost (ns/op ÷ size), **before → after**:

| size | `updateOptional` before | after | `updateValidated` before | after |
|-----:|------------:|-----------:|------------:|-----------:|
| 16   | 256 | 446 | — | 447 |
| 64   | 530 | 272 | 545 | 545 |
| 256  | 802 | 111 | 810 | 346 |
| 1024 | **10,400** | **100** | **2,959** | **277** |

Before, per-element cost *climbs* with size (the O(n²) signature). After, it *converges to a flat plateau* — linear with fixed setup amortizing. Total-time scaling for a 64× size increase (16→1024): `updateOptional` went from **2601×** (quadratic) to **14.3×** (linear). At size 1024, `updateOptional` dropped ~10.6 ms → ~0.10 ms (**~105×**), `updateValidated` ~3.0 ms → ~0.28 ms (**~10.7×**).

> Numbers are a laptop smoke run (2 warmup + 3 iterations × 1s) — wide error bars, but the *scaling shape* is unambiguous. The benchmark stays in-tree as the regression guard; run on CI for tight bands.

## Verification

- `./gradlew check` green across all modules under `-Werror`
- Effect/law tests pass unchanged: `EffectfulUpdateTest`, `ApplicativeLawsTest`, `BoundedAsyncTest`, `HttpEffectsIntegrationTest`, `OpticLawsTest`, `IndexedTraversalTest` — short-circuit / accumulate / empty / async semantics preserved
- No public API change

### DIFF
--- a/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/EffectfulTraversalBenchmark.java
+++ b/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/EffectfulTraversalBenchmark.java
@@ -20,12 +20,13 @@ import org.openjdk.jmh.infra.Blackhole;
  * Scaling benchmark for effectful traversal updates. Each benchmark traverses a list of {@code N}
  * elements and rebuilds it through one effect, where {@code N} is swept across orders of magnitude.
  *
- * <p>The purpose is to expose the complexity class of {@code Traversal#modifyF}'s accumulator, not
- * to produce an absolute ns/op headline. The default {@code modifyF} folds results into a fresh
- * {@code ArrayList} copy per element ({@code O(n^2)} total copying); the pure {@code update} path
- * walks {@code modify} once ({@code O(n)}). Reading the result: divide each row's ns/op by its
- * {@code size} to get per-element cost. If that per-element number is flat across sizes the path is
- * linear; if it climbs roughly proportionally with {@code size} the path is quadratic.
+ * <p>The purpose is to guard the complexity class of {@code Traversal#modifyF}'s accumulator, not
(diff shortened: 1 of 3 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review — added substrate test for many-focus modifyF + reframed the benchmark javadoc since the last pass.

PR #183, opened 2026-06-22
fix(codegen): null-tolerant N-hop null intermediates for @BeanFocus beans and @Focus records

eschizoid:
## Problem

A **multi-property/multi-component** generated lens used as a nullable write-intermediate **beyond the first hop** crashed at runtime — for **both** `@BeanFocus` beans and `@Focus` records:

```
java.lang.NullPointerException: Cannot invoke "Address.getCountryName()" because "p" is null
   at AddressFieldOptics.lambda$static$0   (@BeanFocus)
   ... or "s" is null                       (@Focus record)
```

The generated rebuild reads every **off-path** member off the previous instance to carry it forward:

```java
// @BeanFocus:  (p, v) -> { var c = new Address(); c.setCityName(v); c.setCountryName(p.getCountryName()); ... }
// @Focus:      (s, v) -> new Address(v, s.countryName(), s.zipCode())
```

When the lens writes into a never-constructed nested intermediate, the previous instance is `null` and each off-path read NPEs. At hop 1 the prior null-intermediate seeding builds a fresh non-null instance before the leaf lens runs, so the crash needs a multi-member type reached **through** a null intermediate (hop 2+).

## Why it slipped through

Every prior null-intermediate fixture was **single-member** (`Leaf { String value }` / `record Leaf(String value)`). A one-member lens has no off-path read, so it tolerates a null previous instance regardless. No fixture exercised the multi-member shape.

## Fix

Null-guard each off-path read in the generated lens, in **both** codegen generators:

```java
// @BeanFocus — AbstractTelescopeProcessor.beanRebuild / new offPathRead(Prop)
c.setCountryName(p == null ? null : p.getCountryName());   c.setZipCode(p == null ? 0 : p.getZipCode());
// @Focus     — FocusProcessor.canonicalSetter
new Address(v, (s == null ? null : s.countryName()), (s == null ? 0 : s.zipCode()))
```

- Shared `primitiveDefaultLiteral` helper (widened to `protected`): primitives default to their JLS literal (`0`/`false`/`'\0'`), references to `null` — matching the reflective `SettersWriter` end-state.
- The focused member always takes the incoming value, so **single-member types emit no guard and are unchanged**.
- Per-lens, gated only on `null` → **N-hop safe by construction**, no depth limit.

### Records: reversing a deliberate stance

The first pass deliberately left `@Focus` records strict ("an immutable record cannot be partially constructed"). That reasoning is wrong — a canonical constructor accepts defaults for the unbuilt components (`new R(v, null, 0)`), exactly as the bean setter rebuild does. The consistency across both codegen paths is worth more than the strictness, so records now behave identically to beans.

## Coverage

Both codegen entry points are covered — the `<X>FieldOptics` holder constants (driven by `ofBean()/of().field()` + `mapperForward`) **and** the fluent `<X>Telescope` navigator (pure codegen, composed-lens `Telescope.set`, not `DeepMap`):

| Path | beans | records |
|------|-------|---------|
| navigator `of().a().b().c().set(...)` | ✅ | ✅ |
| `mapperForward(...)` / `ofBean().field()` | ✅ | ✅ (via navigator/holder) |
| hop 2 | ✅ | ✅ |
| hop 3 (N-hop) | ✅ | ✅ |
| populated intermediate → carry off-path forward | ✅ | ✅ |
| off-path defaults: reference→null, int→0, boolean→false, char→'\0' | ✅ | ✅ (source-pinned) |
| both rebuild strategies (no-arg ctor + setters, static `builder()`) | ✅ | n/a (canonical ctor) |

Source-shape assertions in `BeanFocusProcessorTest` / `FocusProcessorTest` pin the guarded emitted form for every primitive kind. Full suite green across all modules.

## Docs / comments

- Corrected the obsolete `Lens.modify` javadoc bullets for both the `@BeanFocus` and `@Focus` lenses (they previously declared the multi-member null case "out of scope" / "crash loudly by design").


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #186, opened 2026-06-22
fix(codegen): three more runtime-parity fixes (primitive setter, null element, container subtypes)

eschizoid:
## Codegen ↔ runtime parity — three more gaps from the proactive audit

Continues the parity sweep (#184, #185). Each fix closes a place where the `@Focus`/`@BeanFocus`/`@Bridge` codegen path shipped a narrower behaviour than the reflective runtime, found by reading the runtime capabilities against the generated output rather than waiting for a report. Every fix is TDD'd (RED proving the gap, GREEN locking the parity).

### Focused primitive setter null-coalesces (was: NPE)

The generated `@BeanFocus` rebuild wrote the focused value straight into its setter, so setting a primitive lens to `null` (`ofBean(T).field(T::getCount).set(bean, null)`) NPE'd on the implicit unbox. A `focusedArg` helper now coalesces a null focus to the primitive's JLS default; reference setters still pass `null` through. Matches the runtime `SettersWriter`, which skips null on primitive setters.

### Null container element passes through (was: NPE)

A `null` element inside a bridged container called `subBridge.forward(null)`, and the generated `forward`/`backward` had no null-source guard, so it NPE'd on the first field read. Both now start with `if (s == null) return null` — null in, null out, matching the runtime structural `Iso` and hardening the directly-callable `BRIDGE_FN` against a null argument too.

### Container subtypes accepted and rebuilt as the declared concrete class (was: rejected)

`@Bridge` container detection exact-string-matched `java.util.List`/`Set`/`Map`, rejecting a field declared as `ArrayList<X>`/`LinkedList<Y>`/`TreeSet<Z>` even though the runtime `ContainerShape` accepts any subtype via `isAssignableFrom`. Detection now uses erasure assignability (Optional stays exact — it's final). Accepting a concrete subtype forces the rebuild to allocate that exact class (an `ArrayList<Y>` helper can't satisfy a `LinkedList<Y>` constructor parameter), so the element-bridging helpers and the inline identity-element copy allocate the target's declared concrete class — the class itself when instantiable, else the interface's default impl (`List`→`ArrayList`, `Set`→`LinkedHashSet`, `Map`→`LinkedHashMap`) — mirroring the runtime `DeepMap` allocation table, presizing only when the impl has a capacity constructor.

## Tests

- `MigrationRegressionTest`: primitive `@BeanFocus` lens set to null → JLS default.
- `BridgeCodegenTest`: null container element passes through; `LinkedList<Y>` target rebuilt as a `LinkedList` (both the element-bridging and identity-element-copy paths) and round-trips.
- `BridgeProcessorTest`: a `LinkedList<Y>` field auto-lifts and the helper allocates the concrete class.
- Full `:codegen` / `:core` / `:lombok` suites green.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
@@ -1201,14 +1201,14 @@ public abstract class AbstractTelescopeProcessor extends AbstractProcessor {
     if (useBuilder) {
       final var sb = new StringBuilder("(p, v) -> " + pojoName + ".builder()");
       for (var i = 0; i < all.size(); i++) {
-        final var arg = all.get(i).name().equals(target.name()) ? "v" : offPathRead(all.get(i));
+        final var arg = all.get(i).name().equals(target.name()) ? focusedArg(target) : offPathRead(all.get(i));
         sb.append(".").append(setters[i]).append("(").append(arg).append(")");
       }
       return sb.append(".build()").toString();
     }
     final var sb = new StringBuilder("(p, v) -> { final var c = new " + pojoName + "(); ");
     for (var i = 0; i < all.size(); i++) {
-      final var arg = all.get(i).name().equals(target.name()) ? "v" : offPathRead(all.get(i));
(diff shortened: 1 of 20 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #188, opened 2026-06-23
bench(lmf): captured monomorphic/megamorphic dispatch rows

eschizoid:
## Captured monomorphic/megamorphic dispatch rows for the LMF tier

Benchmark-only. No production code changes.

The existing `LmfBenchmark` `*_lmf` rows call the public `Records.read(obj, name)` / `Beans.readProperty` API, which pays a per-call string→`Function` `ClassValue`/map lookup — so they measured *lookup + dispatch* against the pre-resolved `*_methodInvoke` rows' *dispatch only*. That made LMF look slower than reflection, which is misleading.

These new rows resolve the accessor once (as a built lens holds it) and isolate **pure dispatch**, at a monomorphic site (index 0 fixed) vs a megamorphic one (8 distinct LMF synthetic classes cycled through one `apply()` site — the shape a deep-copy traversal over many lens types actually produces), with matching pre-resolved `Method.invoke` baselines.

Measured (CI, 3 forks × 30 iters, JDK 25):

| call site | LMF (captured) | `Method.invoke` |
|---|---|---|
| monomorphic | 1.44 | 7.12 |
| megamorphic (8-way) | 4.37 | 7.71 |

So LMF dispatch is ~5× faster than reflection monomorphic, ~1.6× megamorphic — ADR-0005's premise holds. The ~11 ns in the API rows is ~80% lookup, not dispatch; the megamorphic-site penalty (~3×) is what generated bytecode removes.

These rows are worth keeping: they pin the dispatch-vs-lookup split that the recorded numbers obscured, and they're the baseline any future dispatch-table or runtime-codegen work has to clear.

### DIFF
--- a/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/LmfBenchmark.java
+++ b/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/LmfBenchmark.java
@@ -2,6 +2,9 @@ package io.github.eschizoid.telescope.benchmarks;
 
 import io.github.eschizoid.telescope.internal.Beans;
 import io.github.eschizoid.telescope.internal.Records;
+import java.lang.invoke.LambdaMetafactory;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.lang.reflect.Method;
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

Issue #187, opened 2026-06-23
Explore ClasFile API for Runtime Dispatch


eschizoid:
> Populating a POJO with Reflection vs. with the ClassFile API - small benchmark
I wanted to learn about the ClassFile API, and as a small pet project, I created a benchmark to compare how faster a generated class is than a reflection-based approach when we want to populate a POJO from a Map.
TL;DR: it turned out, the generated bytecode was 5x faster than using reflection.
Details here, enjoy: [https://github.com/erosb/learn-classfile-api](https://github.com/erosb/learn-classfile-api)
And please prove me wrong if you find any mistakes in the benchmark :)

eschizoid:
Neat result, and the ClassFile API is a good fit version-wise — JDK 25 ships it final (JEP 484), so no preview-flag friction. A few thoughts on how this would land *for telescope specifically*, since our starting point isn't the one the benchmark measures against.

### The baseline that matters here is LMF, not reflection

The "5x faster than reflection" is generated bytecode vs `Method.invoke`. But the runtime path already moved off naïve reflection (ADR-0005): field access dispatches through cached `LambdaMetafactory`-built `Function`/`BiConsumer`/`Supplier` (`Beans`/`Records`), which are JIT-inlinable. So we've already eaten most of that 5x. The question for us is **ClassFile-gen vs our LMF tier**, which is a much smaller gap — it has to be measured, not assumed.

### Where it would fit, and the real prize

It'd be a **third dispatch strategy**, between the two we have:

- compile-time codegen (`@Focus`/`@BeanFocus`/`@Bridge`) — fastest, zero runtime cost, needs annotations + the processor
- **runtime ClassFile gen** — codegen-like speed, *no annotations*, generated on first use
- LMF-reflective (`Telescope.of(...)`) — today's default

The interesting prize isn't raw throughput — it's the **residual wrapper cost** our own JMH rows already show: `recordComponentRead_lmf ~7.9ns` / `beanSetterDispatch_lmf ~22ns` vs hand-rolled `~0.6 / ~2.4ns`. That delta is LMF's `Function.apply` indirection, and it gets worse when a *shared* optic call site (we hold these as `Function<Object,Object>` in `ClassValue`/maps) goes **megamorphic**. A bespoke generated class doing direct `getfield`/`putfield` against a monomorphic call site could collapse that toward hand-rolled — i.e. make the **un-annotated** `Telescope.of(...)` path nearly codegen-fast. That's the thing worth proving.

[...]

PR #189, opened 2026-06-23
bench(fieldByName): dispatch baseline (measured, parked)

eschizoid:
## `fieldByName` dispatch baseline — measured, parked

Benchmark-only. No production change. This records the baseline that decided *not* to optimize the string-keyed lens path.

The runtime-checked `fieldByName(String)` lens is class-erased at build, so `Beans.fieldLens` re-resolves on every traversal — `get` routes through `readProperty` (a `ClassValue.get` + map lookup per call); `set` re-walks `propertyNames` + `autoWriter` and re-reads every sibling via `readProperty`. The typed `.field(Accessor)` lens knows the concrete class at build (`implClassOf(getter)`) and captures the accessor/writer once. This benchmark builds each lens once and traverses it hot, to size the re-resolution tax.

Measured (CI, 3 forks × 30 iters, JDK 25):

| Op | `fieldByName` (re-resolves) | `.field` (captured) | tax |
|---|---|---|---|
| `get` | 70.94 | 52.72 | +18.2 ns (+35%) |
| `set` | 232.10 | 173.21 | +58.9 ns (+34%) |

**Decision: parked.** A per-lens last-class memo would close that ~34% gap, but:
- it lands only on the escape hatch (`fieldByName`/`fromMap`), not the common typed path,
- even the fast typed path is ~53ns/get and ~173ns/set — the cost is dominated by the DSL terminal + immutable bean copy, which the memo can't touch, so the user's experience barely moves, and
- it turns a stateless lens stateful for a secondary-surface win, with no evidence of a hot-loop consumer.

The wider conclusion of the dispatch investigation (#187): the LMF substrate is already excellent (~1.4 ns captured dispatch), so there's no large runtime-dispatch win — codegen remains the hot-path answer. Kept here as the recorded baseline + the harness for a future revisit, whose trigger is a real adopter reporting `fieldByName`/`fromMap` in a hot loop.

### DIFF
--- a/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/FieldByNameBenchmark.java
+++ b/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/FieldByNameBenchmark.java
@@ -1,0 +1,135 @@
+package io.github.eschizoid.telescope.benchmarks;
+
+import io.github.eschizoid.telescope.Telescope;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #19, opened 2026-06-07
chore(benchmarks): JMH benchmarks for LambdaMetafactory vs reflection paths

eschizoid:
## Summary

- Add `LmfBenchmark` — six new JMH micro-benchmarks pinning the three single-step `LambdaMetafactory` dispatch hot paths swapped on `main` (Phases 1-3 of ADR-0005) against directly-inlined Java baselines: `Records.read` (record component readers), `Beans.readProperty` (bean getters), and `Beans.settersWriter(...).construct` (bean setter dispatch). Each LMF row has a `_handRolled` sibling so the delta isolates the residual LMF wrapper overhead on top of an inlined accessor / setter.
- Wire `--add-exports=io.github.eschizoid.telescope/io.github.eschizoid.telescope.internal=io.github.eschizoid.telescope.benchmarks` onto `compileJmhJava` only — JMH bundles everything on the classpath at runtime so no runtime flag is needed. Existing module-info / fat-JAR setup is otherwise untouched.
- Update `benchmarks/README.md` with both the descriptive table for the new benchmarks and a captured-numbers section (JDK 25 / Apple Silicon, full 3 warmup + 5 measurement iterations / 1 fork). The LMF wrapper adds ~15-35 ns per dispatch on top of an inlined Java accessor: record reader ~15 ns, bean getter ~26 ns, bean setter ~36 ns — versus ~100-260 ns for the `Method.invoke` path the LMF substrate replaces.

The existing `TelescopeBenchmark` is unchanged. The new `LmfBenchmark` lives alongside it in the same package and uses the same `@State` / `@Benchmark` / `@OutputTimeUnit` / `@BenchmarkMode` conventions; it adds `BenchRecord` and `BenchPojo` fixtures local to the new file. Phases 4-5 (still in flight in PRs #13 / #14) are intentionally out of scope.

## Test plan

- [x] `./gradlew :benchmarks:compileJmhJava` builds clean with the new `--add-exports` flag.
- [x] `./gradlew :core:test` passes (no behavioral changes to the core).
- [x] `./gradlew spotlessCheck` passes (formatting applied via `spotlessApply`).
[...]

PR #190, opened 2026-06-23
fix(codegen): @Bridge element-bridges user-defined Collection/Map subtype fields

eschizoid:
## `@Bridge` element-bridges user-defined Collection/Map subtype fields

Adopter Round 7 (their #19). A nested field typed as a custom collection wrapper — `class ImageUrls extends ArrayList<ImageUrl>`, common in legacy codebases — used to be bean-introspected by the `@Bridge` processor: ArrayList's synthesized `isEmpty()` reads as a property `empty` with no `setEmpty`, so the build failed with the misleading **`...ImageUrlsBO has a no-arg constructor but no setter for 'empty'`**.

Codegen now bridges these fields properly instead of crashing.

### How

A `(2-raw)` branch in the field planner detects a same-kind Collection/Map subtype on either side — `raw↔raw` or a generic interface paired with a subtype (the mixed `List<X>` ↔ `Wrap` case) — walks the supertype graph to extract the element/key type (the subtype's own type-argument list is empty; the element lives in `extends ArrayList<Elem>`), and recurses on the element pair via the existing `planElementSubBridge`. Self-contained raw helpers allocate a fresh target wrapper through its **no-arg constructor** (a subclass doesn't inherit `ArrayList(Collection)`) and fill it with `addAll`/`putAll` for an identity element or an element-bridging loop otherwise. List, Set, and Map are all covered.

This goes **beyond the runtime**, whose `collectionCopyIso` does a verbatim `addAll` with no per-element bridging — a footgun for distinct element types (`ImageUrlDBO` → `ImageUrlBO`) rather than a working conversion. Codegen actually converts the elements.

The `isReflectableDeclared` Collection/Map exclusion remains as the **backstop** for the pairs `(2-raw)` can't claim — a kind mismatch (List-subtype vs Set-subtype) or a subtype opposite a non-container — turning the old `no setter for 'empty'` crash into the accurate `no auto-bridge could be derived` diagnostic.

### Isolation

The generic container path shipped in #186 stays **byte-identical**: `FieldPlan` gains a `rawContainer` flag (existing factories pass `false`), and `applyForward`/`applyBackward`/`emitContainerHelpers`/`importsFor` short-circuit raw plans before the generic logic. Detection reuses the `assignableToRaw` helper from #186.

### Tests

- Harness (generated-source): List/Set/Map distinct-element bridging, List identity-element (`addAll`, no sub-bridge), mixed generic↔raw (forward subtype + backward interface default), `lenient = true` propagation into the element pair (the adopter's exact flag), and two kind-mismatch backstops.
- Runtime end-to-end (`:core`, loads + runs the generated bridge): List element-bridging both directions, Map values-bridged/keys-preserved both directions, identity-element `addAll`.
- Reviewer fleet (correctness / regression / test-coverage) converged clean; the flagged Map-runtime / identity-runtime / lenient gaps are closed here.

### Notes for reviewers

- **Intentional FQN rendering in the raw helpers.** The raw-container helpers render every type fully-qualified (and `importsFor` skips raw plans), unlike the generic #186 helpers which use simple names + imports. This is deliberate: a raw subtype's element type can live in a package the import-collection block wasn't built to discover, so FQN output is always-correct and self-contained. Please don't "unify" it back onto the simpleName+import path — that reintroduces the import-discovery gap.
- **Ctor-less subtype.** A Collection/Map subtype that hides its no-arg constructor is rejected up front with a telescope-authored diagnostic ("no accessible no-arg constructor"), rather than emitting `new Wrap()` that fails later in the consumer's build.
- **`FieldPlan` shape.** The 8th component (`rawContainer` flag) is the pragmatic call for this isolated extension; a `// next variant → sealed refactor first` line-in-the-sand is recorded on the record so the boolean-vs-sealed decision isn't re-litigated.

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -1992,6 +1992,16 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
    *       {@code Y}; bridge the element and unwrap on forward / wrap on backward.
    *   <li>{@code NULLABLE_TO_OPTIONAL} — mirror direction.
    * </ul>
+   *
+   * <p>This flat record now carries several components that are live only for a subset of {@link
+   * Kind}s (the container-impl and raw-container fields for LIST/SET/MAP_VALUES, the null-default
+   * fields for PRIM_WRAPPER, the qualifier for TRANSFORM) — the per-variant validity matrix lives
+   * in the field comments, not the type. That is acceptable for this private, single-file plan
+   * object, but it is the binding constraint now: do NOT add a further variant-specific component
+   * to this flat shape. The next container/variant feature should first convert this to a sealed
+   * hierarchy (a Container variant carrying kind/sub/impls/raw, separate PrimWrapper / Recurse /
+   * Transform / Identity variants) so the compiler enforces the matrix the comments currently
+   * describe.
    */
   private record FieldPlan(
     Kind kind,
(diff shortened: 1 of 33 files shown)


eschizoid:
@copilot please review

PR #194, opened 2026-06-24
docs(examples): graphql-java Map→POJO demo (runtime vs generated @FromMap)

eschizoid:
## graphql-java Map→POJO demo — runtime vs generated `@FromMap`

> Stacked on #193 — depends on the `@FromMap` processor.

A real [graphql-java](https://www.graphql-java.com/) server over the JDK `HttpServer` whose `createUser(input:{…})` resolver turns the argument `Map<String,Object>` into a typed `User`. Answers: **how do you do Map→POJO in a GraalVM native image without a pile of reflection config?**

Two entry points share the `GraphQlServer` harness, differing only in the converter:

| Entry point | Converter | Native? |
|---|---|---|
| `RuntimeFromMapServer` | runtime `Telescope.fromMap` | ❌ JVM only |
| `GeneratedFromMapServer` | generated `@FromMap` `UserFromMap` | ✅ the native-image entry |

### The native A/B (the case for `@FromMap`)
| Path | Native build | Native runtime | Reflection registered |
|---|---|---|---|
| generated `UserFromMap` | builds | **works** | ~0 |
| runtime `Telescope.fromMap` | builds | **`NoSuchMethodException: …writeReplace()`** | 1,097 types |

GraalVM doesn't synthesize `writeReplace` on non-serializable lambdas, so `SerializedLambda` can't recover accessor names — the runtime path dies at runtime. The generated converter has none of that machinery and runs `--no-fallback` with no hand-written reachability config.

### Layout & tests
`model/` (the `@FromMap` domain) · `server/` (harness + entry points) · `src/test` `GraphQlServerTest` asserts both tiers through `GraphQlServer.serveOnce`. Entry points just run and print; verification lives in the test. `./gradlew :examples:graphql:test`


eschizoid:
Folded into #192 (the demos PR).

PR #195, opened 2026-06-24
fix(bridge): auto-discover carrier-form @Bridge via a ServiceLoader SPI

eschizoid:
## Problem

A carrier-form `@Bridge` (the third-class form from ADR-0007) emits its converter in the **carrier's** package. But zero-row `mapperForward(A, B)` auto-discovery only probed for a sibling bridge in the **source's** package, by a name derived from the source class. The two never matched on either name or package, so a declared carrier bridge was **silently skipped** — `mapperForward` (lenient by default) fell through to same-name mapping and dropped the carrier's `@Rename`s. A renamed field came back `null` with no error. Adopter-reported, P2.

## Why the obvious fix doesn't work

Renaming the carrier bridge to the probe's convention closes the *name* gap but not the *package* gap — the probe does `Class.forName` in the source's package, and ADR-0007 explicitly forbids emitting the carrier bridge there (a JSR-269 processor can't write into a package its module doesn't own). A source-package-keyed probe fundamentally can't locate a class in the carrier's package by name. Discovery has to be package-agnostic.

## Fix

A `ServiceLoader`-based discovery path, keyed by `(source, target)` — emission stays in the carrier's package (ADR-0007 unchanged), the registry finds it by pair, not by name.

- **New `BridgeProvider` SPI** (`conversion`) + **`BridgeRegistry`** lookup over `ServiceLoader`.
- **`BridgeProcessor`** carrier path emits a sibling `<Carrier>BridgeProvider` and registers it in `META-INF/services` (accumulated, written once in `processingOver()`). `core`'s `module-info` declares `uses BridgeProvider`.
- **`mapperForward`** consults the registry after the name probe misses, before the lenient default. The model-anchored form is untouched — already name-discoverable, emits no provider.
- **Loud, not silent**: a null bridge constant or two providers for one pair throw; a pair with *no* provider keeps the intended lenient default, so the lenient-by-default contract is preserved (the registry makes a *declared* carrier impossible to silently skip; an *absent* bridge still maps leniently).
- **`ProcessorHarness`** now captures resource outputs in memory — the generated services file is assertable and never escapes to disk under `-proc:only`.

## Tests (TDD)

- `BridgeRegistryTest` — registry lookup + loud-fail on null/ambiguous providers.
- `MapperForwardRegistryTest` — zero-row `mapperForward` routes through a registered bridge.
- `CarrierBridgeAutodiscoverTest` — end-to-end, a **cross-package** carrier `@Bridge` is auto-discovered and its rename applied.
- `BridgeProcessorTest` — carrier emits provider + services entry; model-anchored emits neither.

Full `:core` / `:codegen` / `:lombok` suites + javadoc + spotless green.

## Notes

- New public API (`BridgeProvider`, `BridgeRegistry`) — flagged for review, **not** auto-merged.
- ADR-0011 records the decision and supersedes ADR-0007's "use the BRIDGE constant directly" limitation. Numbered 0011 to avoid colliding with the 0010 on the in-flight from-map branch.
- Recorded as Bug 20 in `docs/migration-feedback.md`.


eschizoid:
@copilot please review

PR #196, opened 2026-06-25
fix(codegen): import cross-package @Bridge sub-bridges instead of bare simple names

eschizoid:
## Problem

A nested `@Bridge` field bridged to a **same-simple-name type in a different package** — the classic DB-entity → BO split (`…dbm.DocumentData` → `…bom.DocumentData`) — makes the generated parent bridge fail to compile:

```
error: cannot find symbol  DocumentDataToDocumentDataBridge
  return new …bom.B(DocumentDataToDocumentDataBridge.forward(s.getDoc()), …)
```

## Corrected root cause

The report's diagnosis was "the processor fails to emit the sub-bridge classes." It doesn't — the sub-bridge **is** emitted, in the nested source's package (`…dbm`). The real defect: the parent bridge referenced it by **bare simple name and never imported it**. `BridgeProcessor.importsFor` imported container raw/impl types and `Optional`, but added nothing for a `RECURSE` / container-element / Optional-element sub-bridge. A bare simple name resolves for free only when the sub-bridge shares the parent bridge's package (the common single-package case); across packages it's unresolvable. Because `isSameType` is package-aware, two same-simple-name types in different packages are correctly treated as a real (non-identity) sub-pair, so the cross-package reference is reached exactly in this shape.

**Not Lombok-specific.** The report hit it with `@Data`, but the cause is in the import logic, not round-deferred emission — reproduced with plain records.

## Fix

`importsFor` now imports the sub-bridge's FQN (the nested source element's package + the sub-bridge's simple name) whenever it differs from the parent bridge's package. Same-package references stay bare simple names, so **single-package bridge output is byte-for-byte unchanged** — the 37 existing simple-name reference assertions are unaffected.

## Tests (TDD)

- `CrossPackageSubBridgeTest` (`:core`, plain records) — a carrier `@Bridge` over a DB→BO split with same-simple-name nested types compiles and round-trips. This whole test would fail to compile on the pre-fix processor. Proves the fix and that the bug is Lombok-independent.

## Notes

- No public API change — internal codegen + a regression test.
- Recorded as Bug 21 in `docs/migration-feedback.md`.


eschizoid:
@copilot please review

PR #197, opened 2026-06-25
feat(fromMap): extract overload taking a nested fromMap mapper (nested composition)

eschizoid:
## What

A nested `Telescope.fromMap(...)` can now be used directly as the converter of an `extract(...)` row, so a nested `Map<String, Object>` fills a nested POJO **declaratively** — no hand-rolled converter, no cast at the call site:

```java
Telescope.fromMap(CaseListRequest.class,
    extract("caseId", CaseListRequest::getCaseId, Object::toString),
    extract("pageDetails", CaseListRequest::getPageDetails,
        Telescope.fromMap(PageDetails.class,
            extract("pageSize", PageDetails::getPageSize, v -> (Integer) v),
            extract("exclusiveStartKey", PageDetails::getExclusiveStartKey, Object::toString))));
```

## Why an overload (not just a method ref)

A bare `nestedMapper::forward` does **not** satisfy the converter slot: `extract` takes `Function<Object, X>`, but `ForwardMapper.forward` is `Function<Map<String,Object>, X>` — a narrower input than `Object`, which javac rejects. The only way to compose with the existing overload is `v -> nestedMapper.forward((Map<String,Object>) v)` — an unchecked cast at every nested call site, the exact boilerplate the request set out to remove.

The new overload `extract(key, accessor, ForwardMapper<Map<String,Object>, X>)` takes the nested mapper directly and does the one unchecked cast internally, once. `X` binds the nested mapper's output to the target accessor's type (fully type-checked), and an absent nested key yields a `null` component (`forward` is null-in/null-out). It delegates to the existing `Function` overload — no duplication, and `Function` vs `ForwardMapper` argument types mean no overload ambiguity.

## Tests (TDD)

- `FromMapNestedExtractTest` (`:core`): a 2-level nested `fromMap` composes and round-trips; an absent nested key yields a null component (no throw). The composition test would not compile on the pre-overload API.

## Notes

- New public API (one `extract` overload). Resolves Enh 12 in `docs/migration-feedback.md` (the "investigate composition first" condition — composition is blocked without a cast, so the overload is justified).

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/mapping/MapExtractStep.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/mapping/MapExtractStep.java
@@ -2,6 +2,8 @@ package io.github.eschizoid.telescope.mapping;
 
 import io.github.eschizoid.telescope.Telescope;
 import io.github.eschizoid.telescope.Telescope.Accessor;
+import io.github.eschizoid.telescope.conversion.ForwardMapper;
+import java.util.Map;
 import java.util.function.Function;
 
 /**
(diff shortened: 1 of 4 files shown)


eschizoid:
@copilot please review

PR #198, opened 2026-06-25
fix(core): resolve same-typed to(src, tgt) renames through autoIso (P1 bidirectional Mapper CCE)

eschizoid:
## Problem (P1, runtime ClassCastException)

A bidirectional `Mapper` (`Telescope.mapper()` / `mapperBuilder().build()`) throws `ClassCastException` at runtime when a **2-arg `to(src, tgt)` rename** maps a source field to a differently-typed target field:

```java
mapperBuilder(DocSrc.class, DocTgt.class)
    .add(to(DocSrc::getDocUpdateAttempts /* Integer */, DocTgt::getNumberOfAttempts /* String */))
    .build()
    .forward(src);   // CCE: Integer cannot be cast to String, at SettersWriter.construct / assembleIso
```

`mapperForward()` is unaffected (lenient). It surfaces through **both** construct paths — reflective `Beans.SettersWriter.construct` *and* the `@BeanFocus`-generated `FieldOptics.construct` — but they share one root.

## Root cause

`Mapping.to(Accessor<A,X>, Accessor<B,X>)` is meant to be *same-typed*, but javac infers the shared `X` as the **least-upper-bound** of the two accessors' return types — so `to(Integer-getter, String-getter)` compiles with `X = Serializable & Comparable` and reaches `DeepMap` as a `SameTypedTo`. `fieldIsoOf` returned a **blind `Iso.identity()`** without checking the actual leaf types, and a renamed explicit row **bypasses** the `computeAutoIso` shape check (which only runs for auto-matched same-name fields). So an `Integer` is identity-passed into a `String` setter at construct time → CCE. Build succeeds; it explodes at runtime.

(The reporter attributed the CCE to a typed `to(..., forward, backward)` transform on `sorId`; that path actually applies the conversion correctly — the real trigger is the 2-arg rename of `docUpdateAttempts → numberOfAttempts`.)

## Fix

A `SameTypedTo` carries no conversion, so it resolves exactly like an auto-matched (possibly renamed) field — **route it through `autoIso`** instead of a blind `Iso.identity()`. That one tested path supplies identity / null-safe primitive-wrapper / container lifting / `DEFAULT` null-wrapping **and** the clear incompatible-shapes rejection. Net `+31 / -13`: deletes the `SameTypedTo`-specific `wrapDefaultOnNull` special-casing, no new validation code. The shared incompatible-shapes message gained a pointer to `to(src, tgt, forward, backward)`. `fieldIsoOf`'s now-unreachable `SameTypedTo` branch became a defensive routing guard.

**Bonus correctness:** a `SameTypedTo` over `Integer↔int` or `List<X>↔List<X>` now resolves properly (it was previously blind identity, which only worked when the types were literally equal).

## `@Bridge` codegen — checked, no analogous hole

A `@Bridge` `@Rename` between mismatched-type fields is rejected at **compile** time (`@Bridge … field 'x' has incompatible types (Integer vs String)`) — `@Rename` reads field types from the model by name, so there's no method-ref LUB inference to slip through.

## Tests

- `RenameLeafTypeMismatchTest` — the 2-arg rename mismatch now fails at build with a field-named message pointing to the 4-arg form, on **both** the reflective and `@BeanFocus` construct paths; and the 4-arg `to(src, tgt, forward, backward)` maps correctly.
- `BidirMapperTypedTransformTest` — pins that an explicit typed transform on a mismatched scalar still round-trips.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/DeepMap.java
@@ -490,18 +490,30 @@ public final class DeepMap {
         // sync (typically same-typed copies of the same column), and the test pin makes the
         // last-row-wins behaviour explicit so silent ambiguity is impossible.
         claimedSrc.add(srcField);
+        final var srcType = srcRefl.genericType(source, srcField);
         final var tgtType = tgtRefl.genericType(target, tgtField);
-        final var rawRowIso = fieldIsoOf(row, srcRefl.genericType(source, srcField), tgtType);
-        // SameTypedTo rows are pure identity at the leaf — wrap with default-on-null when the
-        // mapper's null strategy is DEFAULT so an unset source field lands as the type default
-        // instead of null. Other field-iso rows (TypedTransformTo, ForwardOnlyTransformTo, Via)
-        // carry user-supplied forward functions and are explicitly NOT wrapped — the user already
-        // decided how their lambda handles null. This precedence rule lets toOrElse(...) (a
-        // TypedTransformTo) and explicit transforms win over the global hint without any extra
-        // marker on the row, which is the simplest "per-row beats per-mapper" semantics.
-        final var rowIso = (nullStrategy == NullHint.NullStrategy.DEFAULT && row instanceof SameTypedTo<?, ?, ?>)
-          ? wrapDefaultOnNull(rawRowIso, tgtType)
-          : rawRowIso;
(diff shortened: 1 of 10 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #199, opened 2026-06-26
chore: fix IntelliJ inspection warnings (safe mechanical only)

eschizoid:
## What

A full IDE-inspection pass over all 477 source files, fixing only the genuinely-broken, zero-risk subset — matching the "safe mechanical only" posture of the prior inspection sweep.

## Why so few changes for 477 files

The sweep surfaced a lot of flagged items, but triage showed **~80% are intentional-by-design or IDE false-positives** that would break the library or its tests if "fixed":

- **Correct-but-IDE-flagged** `{@link Method#invoke}` / `Class#forName}` links — the real javadoc renders them fine; IntelliJ's module-graph resolution mislabels them "inaccessible." Left untouched (rewriting them would degrade correct docs).
- **JMH `@Benchmark`/`@Setup` & POJO accessors "never used"** — invoked reflectively; the IDE doesn't model JMH/codegen usage.
- **Generated-symbol "cannot resolve"** (`*Telescope`, `*Bridge`, `*Path`) and **JPMS "does not export/read"** — the build supplies these; the IDE index doesn't.
- **Phantom type params** (`Kind<F,A>`, `EitherK<L>`), **annotation elements**, **inference-only `Class` params** (`of`/`ofBean`/`from`/`fieldByName`) — all load-bearing.
- **Test landmines** — e.g. `LambdaIntrospectionTest`'s `u -> u.name()` is deliberately a lambda (proves lambda rejection); reflectively-used builders/`BRIDGE` fields flagged "unused."

## What changed (12 files, +32/-55)

**Javadoc `{@link}` repairs (library):**
- `DeepMap` — stale `resolution(...)` ref (varargs→array, add missing `boolean`)
- `Compute` / `Constant` — link param FQN splintered by a line-wrap space → simple name (also drops an inline FQN)
- `Telescope` / `MapExtractStep` — links to JPMS-inaccessible `internal` members → `{@code}`
- `AbstractTelescopeProcessor` — dead `writeClass` overload ref + inaccessible holder link

**Test cleanups:**
- redundant `String.toString()` in assertion message suppliers
- constant-message supplier lambdas → plain `String` overload
- broken varargs / cross-module `{@link}`s in fixture javadoc → `{@code}`

## Verification

`spotlessCheck` + `:core:test` + `:codegen:test` + `:lombok:test` + `:core:javadoc` + `:codegen:javadoc` all green (`--rerun-tasks`, no stale cache).

### DIFF
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/AbstractTelescopeProcessor.java
@@ -51,28 +51,10 @@ public abstract class AbstractTelescopeProcessor extends AbstractProcessor {
   protected record Prop(String name, String getter, TypeMirror type) {}
 
   /**
-   * Emit a generated {@code public final} class whose body references {@code Telescope}: the
-   * package declaration, the {@code io.github.eschizoid.telescope.Telescope} import, a one-line
-   * javadoc, and a private constructor, then {@code body} writes the members, then the closing
-   * brace. The package is derived from {@code qualifiedName}. An {@link IOException} is reported as
-   * a compile error on {@code origin}.
-   */
-  protected void writeClass(
-    final String qualifiedName,
-    final String simpleName,
-    final String javadoc,
-    final Element origin,
-    final Consumer<PrintWriter> body
-  ) {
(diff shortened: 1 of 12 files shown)


eschizoid:
@copilot please review

Issue #208, opened 2026-07-03
Compile-time mapper verification: MapperVerifierProcessor + shared PairingRules (ADR-0012)


eschizoid:
The last "MapStruct is safer by default" concession — our own `examples/mapstruct-vs-telescope` README concedes it. MapStruct's trust contract is *"if it compiles, every mapping is complete"*; telescope only delivers that on the opt-in `@Bridge` path, while the ergonomic API (`Telescope.mapper(...)` et al.) checks at construction time. This feature makes the ergonomic API compile-checked, with **no new user ceremony**. Decision record: ADR-0012 (design tree resolved in full: trigger, scope, severity, drift strategy, placement, phasing).

## What ships

`MapperVerifierProcessor` in `:codegen` — a JSR-269 processor that reaches into the Compiler Tree API (`Trees.instance(processingEnv)`, `jdk.compiler`, no new dependency) to verify **call sites** of the four conversion factories at compile time, powered by a shared pairing-rule spec extracted from `DeepMap`.

### Behavioral contract

- **Zero ceremony.** Processor present ⇒ every call site in the module is verified. No annotation to opt in.
- **Severity:** compile **ERROR** by default — the verifier only fires on shapes `DeepMap` would reject at construction, so it can only add errors for provably-runtime-broken code, never falsely pass broken code. Global `-Atelescope.verify=error|warn|off` (default `error`); per-site `@UncheckedMapping(reason)` (new annotation in `annotations/`).
- **Graceful degradation:** non-javac compilers (ECJ) ⇒ one NOTE, no-op; construction-time checking remains the backstop everywhere, so a skipped site is exactly as safe as today.

### Scope (v1 — all four entry points)

| Entry point | Check |
|---|---|
| `Telescope.mapper(A, B, rows…)` | full: completeness (strict bijection) + per-row shape |
| `Telescope.map(A, B, steps…)` | full: completeness + per-row shape |
| `Telescope.mapperForward(A, B, rows…)` | per-row shape only (lenient by contract — unmatched fields are legal) |
| `Telescope.fromMap(T, extract…)` | target-side: extract-row accessor vs `T`'s components, key-duplication, converter-output vs component type where statically visible |

[...]

PR #207, opened 2026-06-30
docs(examples): canonical telescope-vs-MapStruct head-to-head slice

eschizoid:
## What

A runnable, **reproducible** `examples/mapstruct-vs-telescope/` slice that doubles as the **canonical "telescope vs MapStruct" reference** — wired from the main README's "How it compares to MapStruct" section. Same `Order → OrderDto` mapping written both ways, in one module:

```bash
./gradlew :examples:mapstruct-vs-telescope:test
```

This is the adoption deliverable from a design session: *proof before tooling* (validate the migration story by hand before any OpenRewrite investment), and the artifact is itself the adoption content.

## Two acts, one domain

**Act 1 — refactor-safety.** The one cross-named field is `to(Customer::email, CustomerDto::contactEmail)` (method ref) in telescope vs `@Mapping(source="email", target="contactEmail")` (string) in MapStruct. Rename `Customer.email()`:
- telescope's reference follows the IDE refactor; nothing goes stale.
- MapStruct's string is opaque text — **Layer 1** (default `unmappedTargetPolicy=WARN`): silently `null` at runtime; **Layer 2** (strictest `ERROR`): compile error you hand-fix across every `@Mapping`.

**Act 2 — the structural gap.** The same typed path deep-updates the immutable graph — `of().each().field().update()` — which a mapper architecture can't express.

## Fair by construction

- Committed state is **green**: both frameworks map identically (test pins it); MapStruct is configured the normal way, not strawmanned.
- The default-policy silent drop is pinned **permanently in CI** (`SilentDropMapper` + a `null` assertion), so the footgun is demonstrated on every run without anyone applying a rename by hand.
- The rename → divergence itself is a documented **one-command reproduction** (a compile failure can't also be a passing test).
- The README credits where MapStruct still wins: mature ecosystem, all-compile-time generation, and that jakarta validation *also* accumulates (so the validation angle is honestly left out).

## Tests

4 tests, all green: Act 1 baseline equality, bidirectional round-trip (free with telescope), the Layer 1 silent-`null`, and Act 2 immutable deep-update. Build + spotless (Java + both READMEs) clean.

## Files

New `examples:mapstruct-vs-telescope` module (reuses the `benchmarks/` MapStruct 1.6.3 wiring); main `README.md` callout; `settings.gradle.kts` include.

PR #209, opened 2026-07-03
docs: ADR-0012 — compile-time mapper verification via shared pairing spec

eschizoid:
## What

The decision record and roadmap entry for **compile-time mapper verification** — the outcome of a full design-grill session. Implementation spec lives in the tracking issue #208; this PR carries only the docs.

## Why this feature

The last "MapStruct is safer by default" concession — the one `examples/mapstruct-vs-telescope` (#207) honestly concedes: MapStruct's *"if it compiles, every mapping is complete"* only holds for telescope on the opt-in `@Bridge` path, while the ergonomic `map/mapper/mapperForward/fromMap` factories check at construction time.

## The two coupled decisions (ADR-0012)

1. **`MapperVerifierProcessor` in `:codegen`** — JSR-269 + Compiler Tree API, verifies the four factories' call sites at compile time. Zero ceremony (on by default with the processor present), ERROR severity (it only fires on shapes the runtime would reject at construction), row-anchored diagnostics reusing the runtime's message text, graceful no-op off-javac, verify-what's-statically-visible with the construction-time check as the universal backstop.
2. **`PairingRules<P>` extracted from `DeepMap` into `:internal`** — one shared pairing-decision spec with reflection (`ReflectionProps`) and `TypeMirror` (`MirrorProps`) adapters, qualified-exported to `:codegen`. Compile-time and construction-time rules cannot drift, by construction. Refines ADR-0004: strategies stay separate, decision rules unify. Build-time only — never the per-`forward()` hot path.

## Files

- `docs/adr/0012-compile-time-mapper-verification.md` — full context/decision/consequences/alternatives
- `.claude/PLAN.md` — new top **Now** entry (delivery shape: one PR, two commit stages — extraction with zero behavior change gated on the existing suite, then the verifier + parity tests)

Spec: #208.

### DIFF
--- a/docs/adr/0012-compile-time-mapper-verification.md
+++ b/docs/adr/0012-compile-time-mapper-verification.md
@@ -1,0 +1,86 @@
+# ADR-0012: Compile-time mapper verification via a shared pairing spec
+
+**Status:** Accepted · **Date:** 2026-06-30
+
+## Context
+
+Telescope's runtime conversion factories — `Telescope.map`, `Telescope.mapper`, `Telescope.mapperForward`,
+`Telescope.fromMap` — validate their field pairing **at construction time**: the strict bijection guard, the
+incompatible-shapes rejection, and the eager `WriteHint` validation all throw a precise `IllegalStateException` /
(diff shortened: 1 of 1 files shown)


eschizoid:
@copilot please review

PR #210, opened 2026-07-03
feat: compile-time mapper verification — shared PairingRules + MapperVerifierProcessor (ADR-0012)

### LINKED ISSUE #208: Compile-time mapper verification: MapperVerifierProcessor + shared PairingRules (ADR-0012)

eschizoid:
## What

Compile-time mapper verification (ADR-0012, spec #208) — closes the last "MapStruct is safer by default" concession: *"if it compiles, every mapping is complete"* now holds on the ergonomic `Telescope.map` / `mapper` / `mapperForward` API, with zero new user ceremony.

Two commit stages, as spec'd:

### Stage 1 — `refactor(core)`: the shared pairing spec

`DeepMap`'s pairing **decisions** move to `:internal/pairing` — `PairingRules` (the compat lattice `decidePair`, container-view selection, kind discriminators, reflectable exclusions, primitive/wrapper rule, same-name `matchFields`), `PairingMessages` (every diagnostic string, single source of truth), `PropertySystem` (per-world type primitives), `ReflectionProps` (the reflection adapter). `DeepMap` now maps decisions to Isos; the dead private twins are deleted. **Zero behavior change — the full `:internal` + `:core` suites pass unchanged (the parity pin).** JPMS: `internal.pairing` qualified-exports to `:core` and `:codegen`.

### Stage 2 — `feat(codegen)`: `MapperVerifierProcessor`

Consumes the same spec through a `TypeMirror` adapter (`MirrorProps`), so compile-time and construction-time rules **cannot drift, by construction**. Reports the runtime's own diagnostic text as compile errors anchored on the offending row expression.

**Architecture note (learned the hard way):** scanning runs from a **post-ANALYZE `JavacTask` listener**, not in the processing rounds — attributing method bodies mid-round corrupts javac's anonymous-class symbol bookkeeping when the final compile re-attributes (`:internal`'s own test suite caught this). The listener is armored: any exception skips the unit with a NOTE; the verifier can never break a build except through its own diagnostics.

**Coverage:** completeness (strict bijection) + shape checks for `map`/`mapper`; rows-only for the lenient `mapperForward`; 2-arg `to()` renames replay the shared decision (the LUB-hole check at compile time); nested pairs recurse shape-checked but lenient (mirroring the runtime); rows keyed to **nested** pairs (grouped by their accessors' classes, exactly like `groupOverridesByPair`) claim nothing at the call's pair; `writeBean` record-target + duplicate eager checks. Anything dynamic defers silently to the construction backstop — the invariant is *the verifier only fires where construction would reject*.

**Controls:** on by default with the processor present; `-Atelescope.verify=error|warn|off`; per-site `@UncheckedMapping(reason)` (new `annotations/` type). `:core`'s own test compile sets `verify=off` — that suite deliberately constructs invalid mappers to pin the construction-time rejections.

## Deviations from the #208 spec (flagged, not silent)

- **`fromMap` verification deferred** — its runtime validation semantics (duplicate extract targets, key rules) need pinning before mirroring them; guessing risked false positives. Tracked in #208.
- **Write-strategy ladder feasibility** (BUILDER-on-no-builder etc.) deferred with it; `writeBean` record-target + duplicate checks shipped.
- **toOrElse/toOrElseGet are claims-only**: the runtime constructs them without a shape gate today, and the verifier never fires where construction succeeds. (Routing them through `autoIso` at runtime is the right follow-up; then the verifier inherits it for free.)

## Verification

- 17 harness tests — completeness, shapes (scalar/nested/container-lifted), degradation (dynamic rows, non-literal classes, `@UncheckedMapping`, `off`/`warn` flags), duplicate rows, `writeBean`, single-diagnostic-per-problem. Via the new `ProcessorHarness.compileFully` (full javac pipeline; post-analysis listeners never fire under `-proc:only`; class output sinks in memory).
- **Full repo build green with the verifier live in every module** — examples, benchmarks, starters, lombok. Real dogfood: the demos' nested-keyed rows exposed the pair-scoping bug during development; fixed and covered.

Closes #208.

### DIFF
--- a/codegen/build.gradle.kts
+++ b/codegen/build.gradle.kts
@@ -53,6 +53,10 @@ tasks.withType<Javadoc>().configureEach {
 
 dependencies {
     api(project(":core"))
+    // MapperVerifierProcessor executes PairingRules at annotation-processing runtime. The classes
+    // arrive transitively through :core's api scope today, but the verifier's dependency is direct
+    // — declared explicitly so a future :core scope change can't silently break the processor path.
+    implementation(project(":internal"))
 
     testImplementation(project(":core"))
     testImplementation(platform(libs.junitBom))
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/MapperVerifierProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/MapperVerifierProcessor.java
@@ -138,9 +138,8 @@ public final class MapperVerifierProcessor extends AbstractProcessor {
             } catch (final RuntimeException e) {
               // The verifier must never break a build except through its own diagnostics — an
               // uncaught exception in a task listener aborts the compile. Skip the unit; the
-              // construction-time validation still applies. If this ever fires it is a
-              // verifier
-              // bug: verbose mode appends the stack trace so the report is debuggable.
+              // construction-time validation still applies. If this ever fires it is a verifier
+              // bug; verbose mode appends the stack trace so the report is debuggable.
               final var detail = new StringWriter();
               if (verbose) e.printStackTrace(new PrintWriter(detail));
               processingEnv
@@ -452,8 +451,9 @@ public final class MapperVerifierProcessor extends AbstractProcessor {
         report(row, PairingMessages.writeBeanTargetsRecord(elements.getBinaryName(targetEl).toString()));
         return;
       }
-      if (!hintTargets.add(elements.getBinaryName(targetEl).toString())) {
-        report(row, PairingMessages.duplicateWriteBeanHint(elements.getBinaryName(targetEl).toString()));
+      final var binaryName = elements.getBinaryName(targetEl).toString();
+      if (!hintTargets.add(binaryName)) {
+        report(row, PairingMessages.duplicateWriteBeanHint(binaryName));
       }
     }
 
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/MapperVerifierProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/MapperVerifierProcessor.java
@@ -16,6 +16,8 @@ import io.github.eschizoid.telescope.annotations.UncheckedMapping;
 import io.github.eschizoid.telescope.internal.pairing.PairDecision;
 import io.github.eschizoid.telescope.internal.pairing.PairingMessages;
 import io.github.eschizoid.telescope.internal.pairing.PairingRules;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
@@ -63,7 +65,7 @@ import javax.tools.Diagnostic;
  *
  * <p>Scope notes, mirroring the runtime exactly: {@code mapperForward} is lenient by contract, so
  * only its explicit rows are checked (no completeness). Rows that carry user conversion functions
- * ({@code to(src, tgt, fwd, bwd)}, {@code forward}, {@code enumTo}, {@code via}) are claims only —
+ * ({@code to(src, tgt, fwd, bwd)}, {@code toOneWay}, {@code enumTo}, {@code via}) are claims only —
  * their leaf conversion is the user's. Presence of a target-telescope / {@code constant} / {@code
  * compute} row switches the runtime into its permissive mode, so completeness checking is disabled
  * for that site too. Nested auto-recursed pairs are lenient about unmatched fields (as at runtime)
@@ -95,6 +97,7 @@ public final class MapperVerifierProcessor extends AbstractProcessor {
   private boolean off;
   private boolean verbose;
   private boolean notedUnavailable;
+  private String unavailableReason;
 
   @Override
   public synchronized void init(final ProcessingEnvironment processingEnv) {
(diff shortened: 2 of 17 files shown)


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #211, opened 2026-07-03
test(internal): contract tests for the pairing spec + verifier README claims

eschizoid:
## What

Two follow-ups to the compile-time mapper verification work (ADR-0012):

### 1. Direct contract tests for `internal/pairing` (42 tests)

The shared pairing spec was exercised only **indirectly** — `:core`'s conversion suite drives it through `ReflectionProps`, `:codegen`'s verifier tests through `MirrorProps` — so per-module coverage showed 0% and nothing pinned the spec independently of its consumers. Since the spec is now the load-bearing contract for both worlds, it gets its own suite:

- **`PairingRulesTest`** — the `decidePair` decision lattice branch by branch (identity, primitive↔wrapper both directions, collection/map subtype copy *preceding* reflectable recursion, record/bean recursion, cross-`Optional` bridges both directions, same-kind container lift, map-key mismatch, terminal incompatibility), the `containerViewOf` selection rules (incl. the wildcard-keyed-Map exclusion), the `reflectable` exclusion families, the same-kind discriminator axes (List / Set / SortedSet / Queue / Deque / SortedMap), `matchFields` claim handling and target-order iteration, and the `ContainerView` key-type invariant.
- **`ReflectionPropsTest`** — the reflection adapter's boxing table, class-handle discrimination (primitives and arrays included), parameterized-type decomposition, and the allocability probe's tri-state answers (never `UNKNOWN` in the reflection world).

The `Incompatible` assertions compare against `PairingMessages` output, pinning the byte-identical-diagnostics contract between the runtime throw and the compile-time error.

### 2. README cleanup — stop conceding the gap the verifier closed

The user-facing docs still described the pre-verifier posture:

- `examples/mapstruct-vs-telescope/README.md` said unmapped fields are refused *at construction* only — now also states the compile-time refusal with `telescope-codegen` on the processor path.
- Root `README.md` gains an **Unmapped-target safety** capability-table row, an intro claim, and a pointer next to the eager construction-time validation paragraph.
- `codegen/README.md` — the verifier's shipping artifact — documents the feature for the first time: zero-ceremony default, what's verified vs deferred, the `-Atelescope.verify` / `@UncheckedMapping` knobs, javac-only depth. Also retires the stale "no MapStruct benchmark yet" paragraph (the CI-reproducible matrix has been published in `benchmarks/README.md`).

## Verification

- `./gradlew build` green across all modules.
- `./gradlew spotlessCheck --rerun-tasks` green (cold cache).
- Doc claims cross-checked against `MapperVerifierProcessor`'s actual `@SupportedOptions` and verified-factory set.


eschizoid:
@copilot please review

eschizoid:
@copilot please review

eschizoid:
@copilot please review

PR #212, opened 2026-07-04
fix(ci): unblock release verifyRelease under config cache + sweep stale Path→Telescope docs

eschizoid:
## Why

Two things, both surfaced by the recent `actions/upload-artifact` v4→v7 sweep and the fleet review of #211:

### 1. The Release action was failing

[Run 28679807746](https://github.com/eschizoid/telescope/actions/runs/28679807746) died at **Update Version and Create Tag** (`./gradlew release`). Root cause is not the action bump itself — it's `verifyRelease` under the configuration cache:

```
Configuration cache state could not be cached: field `__snapshotDependencies__` of task
`:verifyRelease` ... error writing value of type 'DefaultSetProperty'
> Resolution of the configuration ':examples:graphql:nativeImageTestClasspathInternal'
  was attempted without an exclusive lock. This is unsafe and not allowed.
```

axion-release's snapshot-dependency scan resolves **every** subproject's configurations from the root `verifyRelease` task. The GraalVM `native-buildtools` plugin (added to `:examples:graphql` in #204) forbids that cross-project resolution, so the scan throws at config-cache store time and aborts the release.

**Fix:** empty the task's `snapshotDependencies` provider. Only the SNAPSHOT scan is neutralized — the uncommitted-changes and ahead-of-remote guards still run. Safe here because every dependency is a catalog-pinned release version, so the SNAPSHOT guard has nothing to catch.

Verified locally: `./gradlew verifyRelease -Prelease.disableChecks` now passes **with the configuration cache on** (previously failed).

### 2. Stale `Path` → `Telescope` doc drift

The codegen navigator was renamed from `<X>Path<R>` / `start()` to `<X>Telescope<R>` / `of()` long ago, and the ADR-0006 metadata holder is `<X>FieldOptics` — but several docs and test strings never got swept. A reviewer fresh-eyes pass on #211 surfaced them. None of the README snippets compiled against the shipped generator.

Corrected across `README.md`, `codegen/README.md`, `lombok/README.md`, and the codegen/lombok processor test javadocs / `@DisplayName`s / stale `"...Path not generated"` failure messages (which named classes the processor never emits).

Also folded in two accuracy overhangs the #211 review flagged:
- **ADR-0012** said the verifier walks "the four factories" — it scans **three** (`map`/`mapper`/`mapperForward`); `fromMap` verification is descoped to a tracked follow-up (#208).
- The root README **writeBean** paragraph over-claimed — only the record-target and duplicate-hint rejections replay at compile time; builder-feasibility and unused-hint stay construction-time (both still eager and loud).

## Verification

- `./gradlew build` — green.
- `./gradlew :codegen:test :lombok:test spotlessCheck --rerun-tasks` — green (the test-string edits).
- `./gradlew verifyRelease -Prelease.disableChecks` — green with config cache on (the actual release-workflow gate).
- Every doc claim re-checked against the processor sources (`FocusProcessor` `pathName = ... + "Telescope"`, factory `of()`, holder `... + "FieldOptics"`, three-factory scan).

`.claude/CLAUDE.md` carried the same drift and was corrected locally, but it's gitignored so it's not in this diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1089,10 +1089,11 @@ final Telescope<OrderRecord, OrderPojo> conv = Telescope.map(
 
 Validation is eager: a misconfigured hint (`BUILDER` on a no-builder class, hint targeting a record, duplicate hint,
 unused hint) throws at `Telescope.map(...)` time — not on first `iso.to()` deep in production. And with
-`telescope-codegen` on the annotation-processor path, the same validation moves up to **compile time**:
-statically-visible `map(...)` / `mapper(...)` call sites are replayed by the verifier and violations surface as compile
-errors with the identical diagnostic text
-([details in `telescope-codegen`](codegen/README.md#compile-time-mapper-verification)).
+`telescope-codegen` on the annotation-processor path, the structural rejections that don't need the live classpath — a
+hint targeting a record, a duplicate hint — move up to **compile time**: statically-visible `map(...)` / `mapper(...)`
+call sites are replayed by the verifier and those violations surface as compile errors with the identical diagnostic
+text ([details in `telescope-codegen`](codegen/README.md#compile-time-mapper-verification)). The builder-feasibility and
+unused-hint checks stay at construction time (both still eager and loud).
 
 **`writeBeans(STRATEGY)` — one default for every bean target.** When every entity in the recursion shares the same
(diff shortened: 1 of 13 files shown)


eschizoid:
@copilot please review

PR #215, opened 2026-07-05
build(spotless): make cold provisioning parallel-safe

eschizoid:
## What

Makes Spotless's cold-cache provisioning parallel-safe. A cold `./gradlew clean spotlessApply build --no-build-cache` (with `org.gradle.parallel=true`) fails non-deterministically; a warm build and CI do not.

## Why it fails

`org.gradle.parallel=true` runs every module's Spotless tasks concurrently. On an empty cache they provision their shared caches at the same time and corrupt each other:

- the google-java-format worker classpath → shifting `NoClassDefFoundError` on `RegularImmutableAsList` / `Throwables` / `Newlines`
- the prettier `node_modules` → `NoSuchFileException` on `chevrotain/…/constants.ts` (an incomplete `npm install`)

The shifting class names are the tell that it's a classloader/provisioning race, not a missing dependency. **CI never hits it** because `gradle/actions/setup-gradle` restores the Gradle caches, so nothing provisions cold; warm local builds reuse the caches too.

## Fix (two complementary parts, both required)

- **`SpotlessProvisionLock`** — a single-permit shared `BuildService` (`maxParallelUsages = 1`) that every `SpotlessTask` uses, so provisioning is serialized and no two gjf workers build their classpath at once. Warm/incremental builds (caches already present) are unaffected.
- **`npmInstallCache()`** on both prettier steps — the prettier `node_modules` now comes from spotless's managed shared cache (installed once, correctly) instead of a flaky per-project fresh `npm install`.

Serialization alone still tripped the prettier install; `npmInstallCache()` alone still tripped the gjf worker — both are needed.

## Verification

`./gradlew clean spotlessApply build --no-build-cache` now passes cold across repeated runs. No formatting output changes (google-java-format and prettier versions unchanged).


eschizoid:
Closing — not needed. Investigation from a pristine cache showed the cold-build failures were a corrupted google-java-format/guava jar in the local Gradle cache (from interrupting a cold provision: killed daemons / rm -rf mid-install), NOT a cold-parallel provisioning race. Baseline main passes `clean spotlessApply build --no-build-cache` cold reliably (3/3) once the jar is refreshed. Resolution for anyone who hits it: `./gradlew --refresh-dependencies` (or delete the com.google.googlejavaformat / com.google.guava dirs under ~/.gradle/caches/modules-2). No build-config change warranted.

PR #245, opened 2026-07-16
refactor(core): unify the four effectful update* variants through one Applicative seam

eschizoid:
## What

The four effectful update variants — `updateAsync` / `updateOptional` / `updateEither` / `updateValidated` — were each the identical shape:

```java
XK.unbox(optic.modifyF(XK.applicative(), source, a -> XK.box(fn.apply(a))))
```

Extract that into one private seam, `updateVia(Applicative<F>, source, fn, box, unbox)`; the four public methods become thin delegators differing only in witness/applicative and the `box`/`unbox` pair.

## Honest framing (please judge the depth)

This is a **DRY** refactor of four one-liners into one shared mechanism + four delegators. It centralizes the "how to run `modifyF` with an effect's Applicative and box/unbox the carrier" knowledge — the Kind/Applicative HKT seam — so adding a future effect is one delegator + its `*K` witness.

But I want the review to explicitly judge: **is `updateVia` a genuine deepening, or a shallow extraction?** Its interface is a 3-type-param generic (`<F extends Kind.Witness, EffA, EffS>`) with two `Function` (box/unbox) params — arguably nearly as complex as the four self-contained one-liners it replaces. If the panel judges it net-negative (indirection > DRY win), I'll revert and leave the four one-liners as-is. The four public methods stay identical in signature and behaviour either way.

## Correctness

Behaviour-preserving — the delegators expand to byte-identical calls. Full `:core` (including the effectful-update tests) + `:internal` suites green. Type inference resolves without extra witnesses beyond the pre-existing `ValidatedK.<E>forError()` / `EitherK.<E>forLeft()`.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -17,7 +17,9 @@ import io.github.eschizoid.telescope.internal.NullDefaults;
 import io.github.eschizoid.telescope.internal.Records;
 import io.github.eschizoid.telescope.internal.Reflective;
 import io.github.eschizoid.telescope.internal.optics.Affine;
+import io.github.eschizoid.telescope.internal.optics.Applicative;
 import io.github.eschizoid.telescope.internal.optics.Iso;
+import io.github.eschizoid.telescope.internal.optics.Kind;
 import io.github.eschizoid.telescope.internal.optics.Lens;
 import io.github.eschizoid.telescope.internal.optics.Prism;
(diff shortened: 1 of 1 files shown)


eschizoid:
Closing without merging. The full 4-reviewer panel confirmed this correct but **shallow**: `updateVia`'s interface (3 type params + 5 params incl. two Function refs) is heavier than its one-line body, the deletion test shows only trivial self-contained one-liners reappear (the real mechanism already lives once in `Traversal.modifyF`), it costs locality, and the 'cheap future effect' benefit is speculative YAGNI. The four inline `update*` one-liners are already the right design. Keeping them.

PR #267, opened 2026-07-22
perf(core): loop-based read folds via Fold.visitWhile (exists ~4.4x, gated)

eschizoid:
## What

Read terminals rode `Fold.getAll`, which composes across hops as `self.getAll(source).flatMap(next::getAll)` — a deep path builds a tower of per-element single-element `Stream`s, so `count`/`exists`/`toList` over a 100-leaf tree pay microseconds of stream machinery regardless of terminal.

Add `Fold.visitWhile(source, Predicate)` — a short-circuitable visitor that loops instead of streaming. Every composing optic overrides it:
- **`Traversal.then`** walks nested loops (an inner short-circuit returns up through the outer visit) — the flatMap-tower killer.
- **`Traversal.filter`** skips non-matching focuses without invoking the visitor.
- The **four collection traversals** loop their container directly; `eachIterable`'s rebuild also now pre-sizes to `source.size()`, matching its eachList/eachSet siblings (P7, bundled).
- Each **single-focus leaf** (Getter/Lens/Iso/Prism/Affine) visits its one focus mirroring its own `getAll` (including Lens's null-source guard). The diamond (`Iso <: Lens, Prism`; `Lens <: Affine, Getter`) forces explicit overrides at Lens and Iso.

`Fold.toList/count/findFirst/any` reroute through `visitWhile`; `getAll` stays for the lazy head-pull `read()`/`find()` use. Telescope's `toList`/`count`/`exists`/`toListIndexed` ride the loop path. `exists()` also sheds a latent `Stream.findAny()` NPE on a null focus.

Review: fleet CONVERGED/CLEAN (per-optic visitWhile-vs-getAll equivalence traced, short-circuit propagation, diamond resolution, modifyF untouched). Added the direct `visitWhile`-vs-`getAll` equivalence + short-circuit + filter pins in `OpticLawsTest`, and a null-focus read-terminal test. Full `:internal` + `:core` suites green.

## Gate numbers (CI, isolated runner, fork=3, 5 warmup / 10 measurement, n=30)

BEFORE = main @ 52ac13d9 ([run 29907671190](https://github.com/eschizoid/telescope/actions/runs/29907671190)), AFTER = this branch @ d535504e ([run 29907672879](https://github.com/eschizoid/telescope/actions/runs/29907672879)). Floors are identical across runs (countHand 108.270 vs 108.287, countAbsoluteFloor 15.215 vs 15.187) — the runs are directly comparable.

| Row | BEFORE (ns/op) | AFTER (ns/op) | Δ |
|---|---|---|---|
| `existsMiss` (scan all 100, no match) | 4685.4 ± 685.5 | **513.5 ± 63.7** | **9.1×** |
| `existsEarlyHit` | 296.7 ± 1.6 | **39.1 ± 0.1** | **7.6×** |
| `countTelescope` | 1597.9 ± 12.9 | **477.5 ± 91.9** | **3.3×** |
| `toListTelescope` | 1963.5 ± 5.4 | **672.8 ± 8.6** | **2.9×** |
| `toListHand` (floor) | 685.6 | 690.7 | stable |
| `countHand` (floor) | 108.3 | 108.3 | stable |

`toList` through the telescope now sits **at the hand-written loop's floor** (672.8 vs 690.7). The short-circuit terminals win biggest — `exists` no longer builds any stream at all. Local runs were abandoned for gating: the agent's own CPU contaminated three consecutive attempts; CI's isolated runners are the honest measurement environment.

### DIFF
--- a/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
+++ b/core/src/main/java/io/github/eschizoid/telescope/Telescope.java
@@ -1521,7 +1521,7 @@ public sealed class Telescope<
     if (optic instanceof final Affine<S, A> affine) {
       return affine.getOption(source).map(List::of).orElseGet(List::of);
     }
-    return optic.getAll(source).toList();
+    return optic.toList(source);
   }
 
   /**
@@ -1538,7 +1538,7 @@ public sealed class Telescope<
   public List<Indexed<A>> toListIndexed(final S source) {
     final var out = new ArrayList<Indexed<A>>();
     final var i = new int[] { 0 };
-    optic.getAll(source).forEach(a -> out.add(new Indexed<>(i[0]++, a)));
+    optic.forEach(source, a -> out.add(new Indexed<>(i[0]++, a)));
     return List.copyOf(out);
   }
 
@@ -1560,7 +1560,7 @@ public sealed class Telescope<
     if (optic instanceof final Affine<S, A> affine) {
       return affine.getOption(source).isPresent() ? 1L : 0L;
     }
-    return optic.getAll(source).count();
+    return optic.count(source);
   }
 
   /**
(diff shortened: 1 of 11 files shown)


eschizoid:
Bundled **P7 (pre-size)** into this PR — `eachIterable.modify` now pre-sizes its List/Set rebuild to `source.size()`, matching the eachList/eachSet siblings (commit fbea3877). Same file, same traversal-hot-path theme; a self-evident consistency fix, no separate benchmark. Also added the `visitWhile`-vs-`getAll` equivalence + short-circuit pin in OpticLawsTest per review (commit a41b68f3).

PR #298, opened 2026-09-06
Runtime mapper correctness and allocation improvements

xchopin:
## Summary

This change hardens runtime structural mapping and delivers a **92× throughput improvement with 99.7% less allocation** for 4,096-element `CopyOnWriteArrayList` mappings. It fixes correctness gaps in cycle handling, inherited generic containers, sorted-map comparators, and class-loader lifetime; adds regression coverage; and introduces a cardinality-based JMH benchmark plus a comparison gate for future performance claims.

## What was the issue

- Cycle tracking treated every repeated object in a traversal as a cycle. A shared object reached through two sibling branches could map as a value in the first branch and `null` in the second.
- Parameterized container subclasses used their own declared type parameters instead of the parameters of their `List`, `Set`, or `Map` supertype. For example, `StringMap<V> extends HashMap<String, V>` could fail with an `IndexOutOfBoundsException` or infer the wrong key type.
- Sorted-map lifts rebuilt a `TreeMap` without the input comparator, silently changing iteration and ordering semantics.
- Lambda metadata was held in global strong-key maps, which could retain application classes and class loaders.
- Container lifts allocated without input cardinality. `CopyOnWriteArrayList` was especially expensive because every mapped element triggered a backing-array copy.

## The outcome

- Active-path cycle tracking now severs only real back-edges. Shared acyclic references map independently, and tracking is isolated for forward/backward, concurrent, reentrant, and exceptional calls.
- Container generic arguments resolve through inherited supertypes in both runtime mapping and the compile-time verifier.
- Supported sorted maps retain their comparator. Changed sorted-set element types with a custom comparator now produce an explicit `Mapping.via(...)` diagnostic.
- Method-reference metadata uses `ClassValue`, allowing disposable application class loaders to unload.
- Lists, maps, and sets allocate using the source size. Large copy-on-write containers map into a temporary pre-sized collection and perform one bulk copy into the final container; empty and singleton inputs retain the lightweight path.

Measured with JMH 1.36, OpenJDK 25.0.3, Apple M5 Pro, three forks, three warmups, five 500 ms measurements:

| Forward runtime mapping | Before | After |
| --- | ---: | ---: |
| `List`, 4,096 elements | 14,745 ns / 115,824 B | 12,728 ns / 82,008 B |
| `CopyOnWriteArrayList`, 4,096 elements | 1,170,474 ns / 33.7 MB | 12,655 ns / 98 KB |
| `Map`, 4,096 elements | 40,031 ns / 262,336 B | 31,404 ns / 229,488 B |
| `Set`, 4,096 elements | 37,894 ns / 295,136 B | 30,352 ns / 262,288 B |

The full benchmark method, limitations, and reproduction command are in `docs/perf-runtime-collections.md`.

## How to review

Review the commits in this order:

1. `fix: track only active mapper cycles` — cycle-scoped identity tracking in `DeepMap`.
2. `fix: resolve inherited container type arguments` — shared pairing rules plus reflection/compiler adapters.
3. `perf: reduce runtime container allocations` — input-aware allocators and bulk copy-on-write construction.
4. `fix: release cached method-reference classes` — `ClassValue` metadata cache.
5. `test: cover runtime mapper regressions` — cycle, inherited generics, comparators, class unloading, and strategy parity.
6. `test: add runtime allocation benchmark gate` — JMH cardinality benchmark and JSON comparison script.
7. `docs: document runtime mapper semantics and benchmarks` — public semantics and measurement guidance.

Validation completed:

```bash
./gradlew check --no-configuration-cache
```

The JMH comparison script passes for container cardinalities 0, 1, 16, 256, and 4,096, as well as flat, nested, deep runtime mapping and mapper construction:

```bash
python3 scripts/compare-runtime-benchmarks.py before.json after.json --check
```

Co-Authored: Codex + Claude

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -14,6 +14,12 @@ This module is not published. It depends on `:core` and `:codegen` and is only b
 Config (in `benchmarks/build.gradle.kts`): average time in nanoseconds, 3 warmup iterations, 5 measurement iterations, 1
 fork. Results are written to `benchmarks/build/results/jmh/`. A full run takes a few minutes.
 
+## Runtime allocation regression measurements
+
+See [the runtime mapper review measurements](../docs/perf-runtime-collections.md) for the cardinality benchmark, saved
+before/after JMH data, and an executable regression gate. These measure reused runtime mappers separately from mapper
+construction.
+
 ## What it measures
 
 The record benchmarks walk a `Company -> Department -> Address` tree and update the deeply-nested `Address::city`; the
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -16,9 +16,9 @@ fork. Results are written to `benchmarks/build/results/jmh/`. A full run takes a
 
 ## Runtime allocation regression measurements
 
-See [the runtime mapper review measurements](../docs/perf-runtime-collections.md) for the cardinality benchmark, saved
-before/after JMH data, and an executable regression gate. These measure reused runtime mappers separately from mapper
-construction.
+See [the runtime container mapping measurements](../docs/perf-runtime-collections.md) for the cardinality benchmark,
+saved before/after JMH data, and an executable regression gate. These measure reused runtime mappers separately from
+mapper construction.
 
 ## What it measures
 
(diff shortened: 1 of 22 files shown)


xchopin:
Hey @eschizoid :D as promised, since I got my AWS certification, I'm contributing to the project. I hope you'll enjoy this change

eschizoid:
Thanks @xchopin, great PR. That cycle bug was silent data loss waiting to happen. Glad you fixed the generics in the pairing spec instead of patching around it, and your benchmarks held up when we reran them on CI (115x on the copy-on-write case). We're keeping the benchmark gate.

PR #302, opened 2026-09-08
docs: update perf table

xchopin:
Related to my previous PR; update the perf comparison

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,11 +405,11 @@ In the included JMH workloads (MapStruct 1.6.3, CI hardware, JDK 25 —
 [methodology, environment, and both runs](docs/perf-mapstruct-comparison.md)), telescope codegen and MapStruct codegen
 land in the same performance class:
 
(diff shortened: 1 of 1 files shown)


eschizoid:
Pushed two commits to your branch. Spotless markdown formatting was what turned the build red, that's fixed and CI is green.

I also swapped in CI numbers. The line right above the table sources it from CI hardware, so local ns/op made the README contradict itself. Two runs of the Benchmarks workflow: flat 1.08x and 1.13x, nested 1.20x and 1.21x, deep 1.14x and 1.19x. Your flat matches. Nested and deep came out about 15% better on your machine than on the runner, which is why the absolute ns/op are gone again: they track whichever runner the job lands on, the ratio within a run doesn't.

You were right that the table was stale though. And the nested cell had been left without a number on purpose, because two earlier runs disagreed (1.04x, then 1.42x). Two runs agreeing to within 0.01 is enough to publish one, so nested has a real number now because of this PR.

Issue #314, opened 2026-09-09
Performance audit — September 2026 findings and sequencing


eschizoid:
Tracking issue for the September 2026 performance audit. Four independent review passes swept the codegen hot path, the container and collection lifts, the public DSL terminals and fusion engine, and the reflective substrate, hunting for the class of defect found in #298 — accidental superlinearity and per-operation work that belongs at build time.

Every finding below was verified against source and measured. Wall-clock numbers are machine-specific; allocation figures are deterministic.

## The honest negative, first

**None of this closes the MapStruct gap.** The generated bridge was diffed against MapStruct's generated implementation for the same shape: same allocation count, same pre-sized `ArrayList` plus enhanced-for loop, and telescope uses static leaf calls where MapStruct uses virtual ones. Measured allocation is 376.0 B/op for both, byte for byte. The deep-tier residual is code layout and inlining, not work. That line of attack is closed — `docs/perf-mapstruct-comparison.md` was right.

What the audit found instead is that telescope's own most common operation is pathologically slow, on a path no benchmark covers.

## Priority 1 — correctness-grade performance defects

| issue | defect | measured |
| --- | --- | --- |
| #303 | `read()` / `find()` evaluate every focus and return the first | 1000 leaf accessor calls to return one value; 25.7 us against 65 ns for `exists()` on the same path |
| #304 | `Records.indexOf` linear scan makes `patch` / `merge` / fixups quadratic | `patch` 18.8x slower than `forward` at arity 20 |

Both are the same species as #298: an operation that should be O(1) or O(n) is O(n) or O(n^2), hidden behind an API that looks cheap. #303 is the headline — reading one value currently costs more than materialising the entire list.

## Priority 2 — significant, well-scoped

| issue | defect | measured |
| --- | --- | --- |
| #305 | Set/Map rebuilds pass size as capacity, forcing a rehash | 1.19x at n=4096, 1.36x at n=32768 |
| #306 | `Mapper.into` re-resolves reader and writer per property per call | 23.2x slower than `forward` at arity 20 |
| #307 | Generated navigator rebuilds the lens per hop instead of using the constant it emits | 475 B and 133 ns against 0 B and 9.4 ns |
| #308 | `BridgeProcessor` emits size-as-capacity for Set/Map | codegen currently worse than the reflective path |
| #309 | `applyForward` / `applyBackward` duplicate the source read expression | correctness risk with non-pure getters, including a possible NPE |

[...]

Issue #311, opened 2026-09-09
perf: Lens.then omits a modify override, making composed writes quadratic in leaf reads


eschizoid:
`Lens.then(Lens)` (`Lens.java:173`), `Lens.then(Iso)` (`:189`), and `Iso.then(Lens)` (`Iso.java:293`) build a composed anonymous `Lens` that overrides only `get` and `set`. `modify` therefore falls through to the default at `Lens.java:103`, which is `set(source, f.apply(get(source)))`, and the composed `set` re-reads the source: `self.set(source, next.set(self.get(source), value))`. Reads recurse, so a single `modify` performs `d(d+1)/2` leaf reads at depth d instead of d.

Counted leaf `get()` invocations for one `modify`:

| depth | `Lens.then` chain | `Traversal.then` chain | optimal |
| ---: | ---: | ---: | ---: |
| 2 | 3 | 2 | 2 |
| 3 | 6 | 3 | 3 |
| 4 | 10 | 4 | 4 |
| 5 | 15 | 5 | 5 |

`Affine.then(Affine)` (`Affine.java:100`) and `Traversal.then` (`Traversal.java:136`) both carry the linear `self.modify(source, a -> next.modify(a, f))` override. `Lens` is the one rung of the lattice that does not.

## Why this is filed as latent

The DSL never constructs a composed `Lens` today, because `Telescope.optic` is declared `Traversal<S, A>` and every composition binds statically to `Traversal.then`. Nothing in production hits this path right now.

That makes it a trap rather than a bug: the correct fix for the `read`/`find` issue is to restore the lattice narrowing at the compose sites, and the moment that lands, every composed `Lens` starts paying the quadratic `modify`, regressing `update` at depth 3 and beyond. The two changes must ship together.

## Fix

Add the linear `modify` override to the composed `Lens` returned by all three `then` pairings, matching what `Affine` and `Traversal` already do.

## Verification

Add an `OpticLawsTest` sibling that asserts leaf-read counts are linear in depth for every pairing in the composition table. The table implies this invariant today but nothing enforces it, which is why the omission survived.

eschizoid:
Closing unmerged. The override is not semantics-preserving, and the reason it cannot easily be made so is worth more than the change.

[...]

Issue #304, opened 2026-09-09
perf: Records.indexOf linear scan makes patch/merge/fixups quadratic in field count


eschizoid:
`Records.indexOf` is a linear scan over the record's components, and it runs on every field read. The bean side does the identical job through a `HashMap`, so the record path — telescope's flagship shape — is strictly slower than the bean path.

```java
// Records.java:497
int indexOf(final String name) {
  for (var i = 0; i < components.length; i++) {
    if (components[i].getName().equals(name)) return i;
  }
  return -1;
}

// Beans.java:438 — the same lookup, O(1)
final var reader = GETTER_INVOKERS.get(beanClass).get(name);
```

Position in the component list, not the work done, dominates the cost. Measured on JDK 25, same record, same operation:

| read | 40-component record |
| --- | ---: |
| first component | 12 ns |
| last component | 111 ns |

Scaling on the last component: 18 ns at arity 5, 111 ns at arity 20.

## Why it is quadratic

Three production sites loop over every component while each read scans, giving O(n^2) in the field count:

| site | shape |
| --- | --- |
| `core/src/main/java/io/github/eschizoid/telescope/conversion/Mapper.java:359` | `sourceRefl.construct(sourceClass, name -> patched.containsKey(name) ? patched.get(name) : sourceRefl.read(base, name))` |
| `core/src/main/java/io/github/eschizoid/telescope/Merge.java:255` | one `src -> sourceRefl.read(src, name)` reader per auto-backfilled field |
| `core/src/main/java/io/github/eschizoid/telescope/TelescopeFixups.java:240` and `:306` | same construct-with-read-lambda shape |

End-to-end effect on the same record pair, measured against `forward` as the control:

| | arity 5 | arity 20 |
| --- | ---: | ---: |
| `Mapper.forward` | 44 ns | 144 ns |
| `Mapper.patch` | 402 ns (9.1x) | 2,713 ns (18.8x) |
| `Telescope.merge` with `auto(...)` | 306 ns (7.0x) | 1,668 ns (11.6x) |

`forward` scales roughly linearly with arity; `patch` scales 6.8x and `merge` 5.4x for the same 4x arity increase.

## Fix

The cheap fix is a `Map<String, Integer>` name index on `RecordInfo`, built once in `RecordInfo.of`, mirroring `Beans.GETTERS`. That alone flattens `Records.read` and removes the quadratic term at every call site.

[...]

PR #315, opened 2026-09-09
fix: read/find short-circuit, and record names resolve by index

### LINKED ISSUE #303: perf: read()/find() evaluate every focus instead of short-circuiting at the first
`read()` and `find()` evaluate every focus in the traversal and return the first one. On a path with n focuses they invoke the leaf accessor n times to produce a single value, so a read costs strictly more than the eager terminals that return everything.

Measured on `Telescope.of(Org.class).each(Org::teams).each(Team::users).field(User::email)`, JDK 25:

| focuses | `read()` | `exists()` | leaf accessor calls per `read()` |
| ------: | -------: | ---------: | -------------------------------: |
| 1 | 273 ns | 135 ns | 1 |
| 10 | 513 ns | 65 ns | 10 |
| 100 | 2,827 ns | 61 ns | 100 |
| 1000 | 25,740 ns | 65 ns | 1000 |

`exists()` is flat because it rides `Fold.visitWhile`. `read()` is dead-linear at ~23 ns per focus and unbounded. On a 3-hop navigator path it also allocates 848 B/op against 192 B/op for `toList()` on the same path — materialising the whole list is roughly 4x cheaper than reading one element.

## Mechanism

`read` (`core/src/main/java/io/github/eschizoid/telescope/Telescope.java:1657`) and `find` (`:1695`) fall through to `optic.getAll(source).iterator()`. The comment above that line claims it grabs the head "in O(1) without materialising the rest" — it does the opposite. Every DSL path roots at `Iso.identity()`, whose `getAll` is a one-element stream, and every hop wraps it in `Traversal.then`'s `self.getAll(source).flatMap(next::getAll)`. `flatMap`'s sink drains its inner stream per outer element, so the first `tryAdvance` on the wrapping spliterator materialises the entire subtree.

The `optic instanceof Lens` / `instanceof Affine` fast paths at `Telescope.java:1645` and `:1649` are meant to prevent exactly this, but they are unreachable for any composed path — see the root cause below.

## Root cause

`Telescope.optic` is declared `final Traversal<S, A>` (`Telescope.java:108`), so `optic.then(next.optic)` (`Telescope.java:1381`) binds statically to `Traversal.then(Traversal)`. The narrowing rules in the composition table (`Lens.then(Lens) = Lens`, and friends) are overloads rather than overrides, so they can never fire through the DSL facade. Every accessor-navigated path is a bare `Traversal` at runtime, including generated navigators, which compose via `path.then(Telescope.lens(...))` in `AbstractTelescopeProcessor`.

Two dead fast paths follow from the same cause: `visitFocuses`' Lens/Affine branches (`Telescope.java:1765`, harmless — it falls through to `visitWhile`), and `Affine.modifyF`'s single-focus override (`Affine.java:79`), whose absence makes `updateOptional` on a 6-hop single-focus path cost 942 ns against 280 ns for `update`.

## Fix

Route `read` and `find` through the loop primitive the lattice already ships: `Fold.findFirst` (`internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Fold.java:77`) does the `visitWhile` plus one-slot box dance that preserves the null-versus-absent distinction both terminals need. This is a behaviour-preserving change with no type changes, and it short-circuits at the first focus.

Restoring the lattice narrowing at the compose sites (a runtime `instanceof` dispatch in a private `compose(...)` helper) is the deeper fix and additionally revives the `Affine.modifyF` shortcut. It must not land alone: see the linked issue on `Lens.then(Lens)`'s missing `modify` override, which is latent today only because the DSL never constructs a composed `Lens`.

## Verification

No existing benchmark covers `read`/`find` on a navigation path — every `*ForwardRead` benchmark goes through a `Mapper` or `Bridge` whose optic is an `Iso` and therefore hits the fast path at `Telescope.java:1645`. That blind spot is why this survived. Add a `navReadFirst` row to the JMH suite over an `each().each().field()` path at n = 1 / 100 / 10000, plus a unit test asserting the leaf accessor is invoked exactly once per `read()`.

eschizoid:
Closes #303 and #304 — both P1 items from the audit tracked in #314. Two independent quadratic-or-worse defects in the read path, each with a benchmark that closes the blind spot that hid it.

## 1. `read()` / `find()` visited every focus (#303)

A `read()` on a 1000-focus path invoked the leaf accessor 1000 times to return one value and cost 25.7 us, while `exists()` on the identical path cost 65 ns.

| focuses | `read()` before | `read()` after | leaf accessor calls before | after |
| ------: | --------------: | -------------: | -------------------------: | ----: |
| 1 | 273 ns | 187 ns | 1 | 1 |
| 10 | 513 ns | 214 ns | 10 | 1 |
| 100 | 2,827 ns | 106 ns | 100 | 1 |
| 1000 | 25,740 ns | 204 ns | 1000 | 1 |

Both terminals fell through to `optic.getAll(source).iterator()`. Every DSL path roots at a one-element identity stream and each hop wraps it in `flatMap`, whose sink drains the inner stream per outer element, so the first `tryAdvance` materialised the whole subtree. The `instanceof Lens` / `instanceof Affine` fast paths that exist to prevent this are unreachable for composed paths: `Telescope.optic` is declared `Traversal<S, A>`, so `then` binds statically to `Traversal.then` and the composition-table narrowing rules — overloads, not overrides — never fire through the facade.

Both now walk `visitWhile` through a private `firstFocus` helper that stops on its first call. `FoldLaws` already pins that `visitWhile` enumerates exactly what `getAll` streams in the same order, so the swap is behaviour-preserving by an executable law rather than by inspection. The `Object[]` box keeps a found-null distinguishable from no focus, which `Fold.findFirst` collapses and both terminals need.

This is the shallow fix by choice. Restoring the lattice narrowing at the compose sites is the deeper one, and it would activate #311's latent quadratic `modify` in composed lenses — trading one regression for another. #311 stays open for whoever takes that on.

## 2. Record component names resolved by scanning (#304)

`RecordInfo.indexOf` walked the component array on every name-keyed read, while the bean side resolves the same thing through a map. Reads paid for a component's position rather than for the read:

| record arity | last-component read before | after |
| -----------: | -------------------------: | ----: |
| 5 | 18 ns | 6 ns |
| 20 | 111 ns | 9 ns |
| 40 | 111 ns | 11 ns |

Reading the last component of a 40-component record now costs the same as reading the first (11 ns against 11 ns; previously 111 against 12) — position-independent, which is the property that matters.

The construct-with-reader-lambda shape resolves a name per component of an entire rebuild, so the scan squared: `Mapper.patch` measured 18.8x `forward` on a 20-field record, and `merge` with `auto(...)` 11.6x. `Merge`'s auto backfill and both `TelescopeFixups` field overrides have the same shape.

Names now resolve through a `Map` built once per record class in `RecordInfo.of`, mirroring how `Beans` keys its invokers.

## Why both survived

Neither had a benchmark that could see it, and both gaps are now closed:

- No row measured `read`/`find` on a navigation path — every `*ForwardRead` goes through a `Mapper` or `Bridge` whose optic is an `Iso` and hits the fast path. `ReadFoldBenchmark` gains `readTelescope` / `findTelescope` next to a `readHand` loop floor.
- `recordComponentRead_lmf` reads index 1 of a three-component record, where walking and indexing are indistinguishable. `LmfBenchmark` gains a twenty-component record reading its last component.

`ReadShortCircuitTest` additionally pins the accessor-invocation count, so the head-grab property is enforced without depending on timing.

Full `./gradlew check` green across every module.


xchopin:
LGTM. Great performance improvement

Issue #308, opened 2026-09-09
perf: BridgeProcessor emits size-as-capacity for Set/Map, making codegen worse than the reflective path


eschizoid:
`BridgeProcessor` emits `new LinkedHashSet<E>(src.size())` (`:2986`) and `new HashMap<K, V>(src.size())` (`:3032`) for Set and Map fields. For hash-based containers the int argument is table capacity, not element count, so with the 0.75 load factor any n greater than `0.75 * nextPow2(n)` — n = 8, 13-16, 25-32, 100, 1000, and so on — forces a full table reallocation plus a rehash of about 0.75n entries mid-build.

Two things make this worse than the runtime-side sibling issue. MapStruct emits `Math.max((int)(x.size() / .75f) + 1, 16)` for exactly this reason, so the comparison is direct. And telescope's own reflective path already does it correctly (`ContainerLifts.java:264`, `:292`), which means codegen is currently worse than the reflective path it exists to beat.

The raw-container helper is a second instance: `rawAllocExpr` (`:2923`) always emits the no-arg constructor even when the chosen fallback impl is `java.util.HashMap` or `LinkedHashSet`, which do have capacity constructors, and `src.size()` is in scope. Filling from capacity 16 costs log2(n/16) successive rehashes.

Allocation measured on the emitted code (exact, from `getThreadAllocatedBytes`):

| n | Set emitted | Set fixed | Map emitted | Map fixed |
| ---: | ---: | ---: | ---: | ---: |
| 16 | 944 B | 864 B | 784 B | 704 B |
| 100 | 5,648 B | 5,120 B | 4,816 B | 4,288 B |
| 1000 | 52,400 B | 48,288 B | 44,368 B | 40,256 B |

Raw-container helper: 5,312 B against 4,288 B correctly sized at n=100 (+24%), 48,496 B against 40,256 B at n=1000 (+20%).

## Fix

In `hasPresizeCtor`, emit `<Impl>.new<Impl>(src.size())` for the four hash-based impls and keep `(src.size())` for `ArrayList`, whose int argument is exact. In `rawAllocExpr`, emit the sized form when the chosen impl is a known JDK default rather than a user subtype whose constructor set is unknown.

Note that three tests currently assert the wrong form and will need updating: `BridgeProcessorTest.java:853`, `:887`, `:935` (the `@DisplayName` says "pre-sized LinkedHashSet"), and `:1038`.

## Verification

[...]

PR #316, opened 2026-09-09
fix: size hash containers for their element count, not their table capacity

### LINKED ISSUE #305: perf: Set/Map rebuilds pass size as capacity, forcing a rehash on every container write
`new LinkedHashSet<>(n)` and `new LinkedHashMap<>(n)` set the initial table capacity, not the expected element count. With the default 0.75 load factor, inserting n elements into a table sized for n always crosses the threshold, so every rebuild pays one full table reallocation plus a rehash of every element and discards the original array.

PR #298 fixed exactly this in `ContainerLifts` and never touched the sibling sites. The asymmetry is visible in the source today:

```java
// ContainerLifts.java:265 — correct, sizes for n elements
LinkedHashSet.newLinkedHashSet(((Collection<?>) input).size());

// Traversals.java:87 — wrong, n is capacity
final var out = new LinkedHashSet<A>(source.size());
```

## Sites

| file | lines |
| --- | --- |
| `internal/src/main/java/io/github/eschizoid/telescope/internal/optics/collections/Traversals.java` | 87 (`eachSet`), 115 (`eachMapValue`), 182 (`eachIterable`, Set branch) |
| `internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Iso.java` | 224, 230 (`liftSet`), 246, 252 (`liftMapValues`) |
| `internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Getter.java` | 85 (`liftSet`), 109 (`liftMapValues`) |
| `core/src/main/java/io/github/eschizoid/telescope/ContainerLifts.java` | 298 (`WeakHashMap`, no size hint at all); 242, 243 (`Vector`, `Stack`, no size hint) |

The `ArrayList` siblings sitting next to each of these are correct — `ArrayList(int)` really is exact capacity. This is the Set/Map half of the same idea, missed.

## Impact

A/B measured against a patched copy, JDK 25, best of 4 rounds:

| path | n=4096 | n=32768 |
| --- | --- | --- |
| `.each(Set).update` | 313 us to 264 us (1.19x) | 3.20 ms to 2.35 ms (1.36x) |
| `.eachValue(Map).update` | 309 us to 260 us (1.19x) | 3.31 ms to 2.51 ms (1.32x) |

Allocation drops by exactly the discarded table array: 16,400 B/op at n=4096, 131,088 B/op at n=32768. Isolated microbenchmarks on the lift helpers show 1.10-1.25x across n=64 to 32768. `WeakHashMap` is worse in relative terms (1.35x to 1.50x) because it starts at capacity 16 and rehashes log2(n/16) times; note it has no `newWeakHashMap` factory, so the correct form is `new WeakHashMap<>(ceil(n / 0.75))`.

The blast radius is wider than the runtime DSL: `AbstractTelescopeProcessor.java:1122` emits `Telescope.asSet(path).each()`, which lands in `Traversals.eachSet()`. This is on the generated "if it compiles, it runs" path. The `Iso` and `Getter` helpers back the public `Mapper.liftSet()` / `liftMapValues()` and their `ForwardMapper` equivalents.

`Vector` and `Stack` measured only 1.02-1.06x because `Vector.add`'s per-element synchronisation dominates; include them for consistency, not for the win.

## Do not "fix" these

`new ConcurrentHashMap<>(n)` (`ContainerLifts.java:295`) and `new IdentityHashMap<>(n)` (`:297`) are already correct. CHM sizes internally for `n + n/2 + 1`, and `IdentityHashMap`'s int argument is documented as expected maximum size, not capacity. Changing either would make things worse.

## Verification

No current benchmark covers this — `ContainerAllocationBenchmark` drives `Telescope.mapper(...)`, which is the `ContainerLifts` path that was already fixed. Add an `.each()` / `.eachValue()` update benchmark at n = 16 / 256 / 4096 / 32768 with `-Pjmh.profilers=gc`, and a `WEAK_HASH_MAP` parameter to the existing benchmark.

eschizoid:
Closes #305 and #308 — the same defect on both sides of the library, tracked in #314.

The int constructor of a hash container takes *table capacity*, not an element count. At the default 0.75 load factor a table sized for n elements crosses its threshold on the nth insert, so every rebuild paid one full reallocation, a rehash of everything already inserted, and threw away the original array.

PR #298 fixed exactly this in `ContainerLifts` and never touched the siblings. The asymmetry was visible in the source:

```java
// ContainerLifts.java — fixed in #298
LinkedHashSet.newLinkedHashSet(((Collection<?>) input).size());

// Traversals.java — same idea, unfixed
final var out = new LinkedHashSet<A>(source.size());
```

## Runtime (#305)

Nine sites now route through the JDK's `newXxx` factories, which apply the load factor for the caller: three `Traversals` rebuilds (`eachSet`, `eachMapValue`, `eachIterable`'s Set branch), four `Iso` lift helpers, and two on `Getter`. `WeakHashMap` ships no such factory and its int argument is capacity, so its size is divided by the load factor explicitly. `Vector` takes the source size, which the neighbouring `ArrayList` branch already did.

Measured against a patched build, JDK 25:

| path | n=4096 | n=32768 |
| --- | --- | --- |
| `.each(Set).update` | 313 us to 264 us (1.19x) | 3.20 ms to 2.35 ms (1.36x) |
| `.eachValue(Map).update` | 309 us to 260 us (1.19x) | 3.31 ms to 2.51 ms (1.32x) |

Allocation falls by exactly the discarded table: 16,400 B/op at n=4096, 131,088 B/op at n=32768.

The reach is wider than the runtime DSL — `AbstractTelescopeProcessor` emits `Telescope.asSet(path).each()`, which lands in `Traversals.eachSet()`, so generated navigators were paying it too. The `Iso` and `Getter` helpers back the public `Mapper.liftSet()` / `liftMapValues()` and their `ForwardMapper` equivalents.

## Codegen (#308)

`BridgeProcessor` emitted the same defect, which put generated bridges *behind the reflective path they exist to beat* on allocation. A single `sizedAlloc` now owns the decision, so a container's sizing form is chosen in one place instead of at three emit sites.

| n | Set emitted | Set fixed | Map emitted | Map fixed |
| ---: | ---: | ---: | ---: | ---: |
| 16 | 944 B | 864 B | 784 B | 704 B |
| 100 | 5,648 B | 5,120 B | 4,816 B | 4,288 B |
| 1000 | 52,400 B | 48,288 B | 44,368 B | 40,256 B |

`ArrayList` keeps its constructor in both tiers — its int argument is an exact element capacity, and the existing assertion at `BridgeProcessorTest:280` still passes unchanged, which is the contrast worth keeping.

## Deliberately not changed

`new ConcurrentHashMap<>(n)` and `new IdentityHashMap<>(n)` are already correct: CHM sizes internally for `n + n/2 + 1`, and `IdentityHashMap`'s int argument is documented as expected maximum size rather than capacity. Both look like instances of this bug and are not.

## Verification

Codegen is covered by generated-source assertions, which can observe the emitted form directly; the Set and Map ones move to the factory form. The runtime side is invisible to unit tests — capacity affects allocation, not behaviour — so `ContainerWriteBenchmark` covers `.each` / `.eachValue` writes across cardinality (16 to 32768), each row against a hand-written loop floor. `ContainerAllocationBenchmark` could not catch it: it drives `Telescope.mapper`, exercising the container lifts rather than the traversal rebuilds.

Full `./gradlew check` green across every module.

### DIFF
--- a/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/ContainerWriteBenchmark.java
+++ b/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/ContainerWriteBenchmark.java
@@ -1,0 +1,85 @@
+package io.github.eschizoid.telescope.benchmarks;
+
+import io.github.eschizoid.telescope.Telescope;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+/**
+ * Traversal writes over hash containers, across cardinality. Sibling of {@link
+ * ContainerAllocationBenchmark}, which drives {@code Telescope.mapper(...)} and therefore covers
+ * the container lifts; these rows drive {@code .each(...)} / {@code .eachValue(...)} updates, which
+ * rebuild through the traversal instead.
+ *
+ * <p>Cardinality is the dimension that matters. A hash table sized for n entries resizes on the nth
+ * insert at the default load factor, so a rebuild that mis-sizes its table pays one full
+ * reallocation plus a rehash of everything it has already inserted — invisible at small n, and
+ * roughly a third of the cost by the tens of thousands. Each telescope row has a hand-written loop
+ * beside it as the floor; the ratio between them is the whole signal.
+ *
+ * <pre>{@code
+ * ./gradlew :benchmarks:jmh -Pjmh.includes=ContainerWriteBenchmark -Pjmh.profilers=gc
+ * }</pre>
+ */
+@State(Scope.Thread)
+public class ContainerWriteBenchmark {
+
+  public record TagSet(Set<String> tags) {}
+
+  public record TagMap(Map<String, String> tags) {}
+
+  @Param({ "16", "256", "4096", "32768" })
+  public int size;
+
+  private TagSet tagSet;
+  private TagMap tagMap;
+  private Telescope<TagSet, String> setElements;
+  private Telescope<TagMap, String> mapValues;
+
+  @Setup
+  public void setup() {
+    final var tags = LinkedHashSet.<String>newLinkedHashSet(size);
+    final var byName = LinkedHashMap.<String, String>newLinkedHashMap(size);
+    for (var i = 0; i < size; i++) {
+      tags.add("tag" + i);
+      byName.put("k" + i, "tag" + i);
+    }
+    tagSet = new TagSet(Collections.unmodifiableSet(tags));
+    tagMap = new TagMap(Collections.unmodifiableMap(byName));
+    setElements = Telescope.of(TagSet.class).each(TagSet::tags);
+    mapValues = Telescope.of(TagMap.class).eachValue(TagMap::tags);
+  }
+
+  @Benchmark
+  public TagSet eachSetUpdate() {
+    return setElements.update(tagSet, String::toUpperCase);
+  }
+
+  /** The loop floor for a Set rebuild: one pass, one correctly sized table. */
+  @Benchmark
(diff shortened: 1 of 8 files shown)


xchopin:
Nice improvements! LGTM

PR #326, opened 2026-09-10
perf(internal): keep a composed Lens write linear in depth

### LINKED ISSUE #311: perf: Lens.then omits a modify override, making composed writes quadratic in leaf reads

eschizoid:
Closes #311.

A composed `Lens` overrides `get` and `set` but not `modify`, so `modify` falls to the interface default — `set(source, f.apply(get(source)))` — and the composed `set` reads the outer focus *again* to rebuild it. Those reads recurse, so one `modify` at depth d costs `d(d+1)/2` leaf reads instead of d:

| depth | reads before | after |
| ---: | ---: | ---: |
| 2 | 3 | 2 |
| 3 | 6 | 3 |
| 5 | 15 | 5 |
| 8 | 45 | 8 |

`Affine.then` and `Traversal.then` already carry the linear `self.modify(source, a -> next.modify(a, f))` override. `Lens` was the one rung of the lattice that did not. The three pairings that produce a `Lens` — `Lens.then(Lens)`, `Lens.then(Iso)`, `Iso.then(Lens)` — now delegate the same way.

## Why it was latent, and why that is not a reason to leave it

Nothing in production builds a composed `Lens` today: `Telescope.optic` is declared `Traversal<S, A>`, so every DSL composition binds to `Traversal.then` and the narrowing rules in the composition table never fire. That is also why this could sit here unnoticed.

It stops being latent the moment anything restores that narrowing — which is exactly what #303's deeper fix would have done, and what a future attempt at it will do again. Landing the override first means that work does not have to discover this as a regression in its own numbers.

## Gate

`CompositionCostTest` counts leaf reads through a chain whose levels report when they are read, at depths 1 through 8, and asserts the count equals the depth. Values are identical either way — the count is the only observable — so a test that asserted results would pass on the quadratic version. Verified as a gate: reverting the `Lens.then(Lens)` override fails two of its three cases; restoring it passes all three.

It also pins that an `Iso` hop adds no read, and that the composed write still produces the right value.

### DIFF
--- a/internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Iso.java
+++ b/internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Iso.java
@@ -302,6 +302,13 @@ public interface Iso<A, B> extends Lens<A, B>, Prism<A, B> {
       public A set(final A source, final C value) {
         return self.from(next.set(self.to(source), value));
       }
+
+      // The set above converts and reads; Lens's default modify would then read again through get.
+      // Converting once and delegating to the inner modify keeps a composed write linear in depth.
+      @Override
+      public A modify(final A source, final Function<? super C, ? extends C> f) {
+        return self.from(next.modify(self.to(source), f));
+      }
     };
   }
 
--- a/internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Lens.java
+++ b/internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Lens.java
@@ -182,6 +182,15 @@ public interface Lens<S, A> extends Affine<S, A>, Getter<S, A> {
       public S set(final S source, final B value) {
         return self.set(source, next.set(self.get(source), value));
       }
+
+      // Without this, modify falls to Lens's default — set(source, f.apply(get(source))) — and the
+      // set above reads the outer focus a second time. Composed that way, one modify at depth d
+      // costs d(d+1)/2 leaf reads instead of d. Delegating to the outer modify keeps it linear, the
+      // way Affine and Traversal already do for their compositions.
+      @Override
+      public S modify(final S source, final Function<? super B, ? extends B> f) {
+        return self.modify(source, a -> next.modify(a, f));
+      }
     };
   }
 
@@ -197,6 +206,13 @@ public interface Lens<S, A> extends Affine<S, A>, Getter<S, A> {
       @Override
       public S set(final S source, final B value) {
         return self.set(source, next.from(value));
+      }
+
+      // Same reason as the Lens pairing above: route through the outer modify so the outer focus
+      // is read once. An Iso needs no read of its own — it converts in both directions.
+      @Override
+      public S modify(final S source, final Function<? super B, ? extends B> f) {
+        return self.modify(source, a -> next.from(f.apply(next.to(a))));
       }
     };
   }
--- a/internal/src/test/java/io/github/eschizoid/telescope/internal/optics/CompositionCostTest.java
+++ b/internal/src/test/java/io/github/eschizoid/telescope/internal/optics/CompositionCostTest.java
@@ -79,6 +79,22 @@ class CompositionCostTest {
   }
 
   @Test
+  @DisplayName("a chain rooted at an Iso reads each level once too")
+  void isoRootedChainIsLinear() {
+    // Iso.then(Lens) produces a Lens as well and carries the same hazard — but only when the lens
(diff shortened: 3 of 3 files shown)


eschizoid:
Closing unmerged. The override is not semantics-preserving, and the reason it cannot easily be made so is worth more than the change.

`Lens.modify`'s default is `set(source, f.apply(get(source)))`, and the null asymmetry that `Lens.java:96-100` states as policy — *"reads of a missing intermediate should surface, writes through a missing intermediate should construct the path"* — is produced by exactly that shape. The `get` half surfaces; the `set` half constructs. Delegating to `self.modify` short-circuits null before either happens, so both halves change:

| case | main | this branch |
| --- | --- | --- |
| `OUTER.then(viaIso).modify(null, f)` | `Outer[inner=Inner[v=null!]]` | **NullPointerException** |
| `OUTER.then(INNER).modify(new Outer(null), f)` | NullPointerException — the missing level surfaces | **`Outer[inner=Inner[v=null!]]`** — a level that did not exist is invented |

The second is the worse direction: a loud failure becomes a silent fabrication. It also leaves the composed optic internally inconsistent — `composed.modify(s, f)` and `composed.set(s, f.apply(composed.get(s)))` now describe different optics, which no law in `OpticLawsTest` encodes, which is why the suite stayed green.

And the cost the override removes is the same `set` re-read that produces the semantics. Linear *and* faithful is not reachable by overriding `modify` alone — it would need the composed `set` to accept an already-read intermediate, which the `Lens` interface has no shape for.

**What holds up:** #311's diagnosis is correct. A composed `Lens` really does pay `d(d+1)/2` leaf reads — 45 at depth 8 — and `Affine`/`Traversal` really do carry linear overrides that `Lens` lacks. It stays open.

**What the next attempt needs to know**, recorded there so it is not rediscovered:

- The null contract must be preserved at the composed level, not inherited from `self.modify`.
- `CompositionCostTest` counts reads and cannot see any of this; a null-source and null-intermediate case per pairing is required alongside.
- The win is currently unrealised. Nothing in main sources, generated code, examples, or benchmarks composes a `Lens` — `Telescope.optic` is declared `Traversal`, and `Lens`/`Iso` `then` are overloads rather than overrides, so a statically-`Traversal` receiver always binds `Traversal.then`. This is lattice correctness, not a live hot path, and should be framed that way.
- One incidental improvement worth keeping: the fix invokes a getter once instead of twice per composed modify, which matters for a lazy-init or proxy-hydrating getter.

`CompositionCostTest` itself was verified load-bearing — removing each of the three overrides independently fails exactly the cases that gate it — so it is worth resurrecting with the null cases added whenever this is attempted again.

PR #361, opened 2026-09-11
fix: rebuild an interface-typed Map field insertion-ordered, as Set is

### LINKED ISSUE #348: codegen: interface-typed Map fields rebuild as HashMap, dropping the iteration order MapStruct preserves
Interface-typed `Map` fields rebuild as `HashMap` (`concreteImplFqn`, `BridgeProcessor.java:~2286`, deliberately matching the runtime's `mapAllocatorFor`), where the generated MapStruct mapper for the same pair emits `LinkedHashMap.newLinkedHashMap(map.size())`. Set fields DO preserve order (`LinkedHashSet`) — the Map side is the odd one out on both telescope paths.

Two consequences. Semantic: a `LinkedHashMap` source behind a `Map`-typed field silently loses iteration order through a bridge, and Set-vs-Map behave inconsistently within the same library. Benchmark fairness: the `container_map_*` rows compare telescope building the cheaper container against MapStruct building the dearer one, so the published ~1.04-1.06x container figures are slightly flattered.

If changed to `LinkedHashMap` for parity (both runtime `ContainerLifts` and the emitter, so the two paths stay in lockstep), the existing `container_map_telescope_*` rows are the gate and will show a small honest regression — which is fine, and the allocation comparison becomes apples-to-apples.

Found by the emitted-code audit.

eschizoid:
Closes #348.

**This changes a default, so it is not mine to merge — flagging it for a decision rather than landing it.**

A conversion that was not asked to reorder should not reorder. A field declared as the bare `Map` interface picked `HashMap` as its family default on both paths, so a `LinkedHashMap` source behind that field came out in hash order. Three facts, each verified rather than taken from the issue:

- **Codegen**: `concreteImplFqn` maps `SET -> java.util.LinkedHashSet` but `MAP_VALUES -> java.util.HashMap`.
- **Runtime**: `ContainerLifts.setAllocatorFor` treats `Set.class` as `LinkedHashSet`; `mapAllocatorFor` treats `Map.class` as `HashMap`.
- **MapStruct**, for the same pair, emits `LinkedHashMap.newLinkedHashMap( map.size() )` — and `LinkedHashSet.newLinkedHashSet( set.size() )` beside it.

So Map was the odd one out twice over: against telescope's own Set handling, and against the mapper we are trying to be a drop-in for. Both telescope paths move together here, so codegen and the reflective mapper still produce the same runtime class for the same field — the existing cross-strategy parity test is what holds that, and its path-to-path assertion never failed during this change.

A field declared as `HashMap` asked for that class specifically and still gets it. Only the bare-interface default moved.

## Blast radius

`LinkedHashMap extends HashMap`, so casts and `instanceof HashMap` are unaffected — I checked rather than assumed. What changes is the iteration order (from unspecified to the source's) and a `getClass() == HashMap.class` identity check. Not source- or binary-breaking, hence `fix:` rather than `fix!`.

## Verification

Four new tests, and the interesting one is the precondition: `theKeysUsedHereActuallyReorderUnderAHashMap` asserts that the ten keys the other tests use really do iterate differently under a `HashMap`. With keys that happened to hash in insertion order, every other assertion in the class would hold no matter which container the rebuild picked — the test would be untriggerable and look fine.

Each half of the fix is revert-proven separately, and each probe fails exactly the path it reverted:

| hunk reverted | fails | passes |
| --- | --- | --- |
| codegen `MAP_VALUES` only | codegen order, backward | runtime order, precondition |
| runtime `Map.class` only | runtime order, backward | codegen order, precondition |

The `items` Set field rides along on every conversion as a second control: it has always rebuilt insertion-ordered, so it stays green either way and separates "the Map default changed" from "the container rebuild stopped preserving anything".

## Four existing tests updated

These pinned the old default and are deliberate expectation changes, not incidental churn:

- `BridgeProcessorTest` ×3 — generated-text assertions for `HashMap` imports and `newHashMap(src.size())`.
- `BridgeCodegenTest` — the cross-strategy parity test. Note **which** of its two assertions failed: the hardcoded `assertEquals(HashMap.class, ...)`. The one that actually guards the seam, codegen-class == runtime-class, kept passing, which is the evidence the two paths stayed in lockstep.

Suites green: `:codegen` 208, `:core` 764 (+5), `:lombok` 30, `:internal` 344, `:quarkus` 6, `:spring-boot-starter` 7.

## Cost

Measured, both runs gc-profiled, with MapStruct as the control since this change cannot touch it — **[full numbers in the comment below](https://github.com/eschizoid/telescope/pull/361#issuecomment-5636234111)**.

The short version: the control moved 1.20x between the two runs, so the timings from this pair are a runner difference and I am not publishing a time delta from them. Allocation is deterministic and decisive — telescope codegen goes **6712 -> 7528 B/op, landing exactly on MapStruct's 7528**, where it previously allocated 816 B/op less by building the cheaper container. The +816 is precisely what `LinkedHashMap` costs here: 100 entries x 8 bytes for the two extra references per entry, plus 16 for the map object.

If this merges, the container row in `docs/perf-mapstruct-comparison.md` — which currently reports that allocation gap as telescope's real win on that tier — has to change with it. I have not touched the doc here, since it should only move if the decision is to take this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,8 +410,8 @@ performance class:
 | --------------------------- | ------------------------------------------------------------------- |
 | flat (5 scalars)            | ~1.07× — a fifth of a nanosecond                                    |
 | nested (one nested type)    | 1.04×–1.46× across runs — a microbenchmark, not a service shape     |
-| deep (3 levels + list hops) | 1.07×–1.19× across runs — 4 ns on a 62 ns conversion at the low end |
-| Set or Map field, 100 items | a tie on time, and ~11% less allocated on the Map shape             |
+| deep (3 levels + list hops) | 1.06×–1.19× across runs — 4 ns on a 62 ns conversion at the low end |
+| Set or Map field, 100 items | a tie on time, and the same allocation                              |
 
 No codegen? `Telescope.mapper(...)` composes each record/bean pair into a single MethodHandle: zero annotations, no
 build step, within ~1.04–3.3× of MapStruct in the same workloads (flat ~3.3×, nested ~2.7×, deep ~1.3×, container fields
--- a/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
+++ b/codegen/src/main/java/io/github/eschizoid/telescope/codegen/BridgeProcessor.java
@@ -2428,11 +2428,12 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
   // FQN of the concrete, instantiable class to allocate for a container field of the given declared
   // type — the declared subtype itself when it is an instantiable class (ArrayList, TreeSet,
   // TreeMap, …), else the default impl for the interface family. The interface-family defaults
-  // match
-  // the runtime DeepMap allocators for the bare interface raws: List → ArrayList
-  // (listAllocatorFor),
-  // Set → LinkedHashSet (setAllocatorFor), Map → HashMap (mapAllocatorFor), so codegen and the
-  // reflective path produce the same runtime class for an interface-typed field.
+  // match the runtime DeepMap allocators for the bare interface raws: List → ArrayList
+  // (listAllocatorFor), Set → LinkedHashSet (setAllocatorFor), Map → LinkedHashMap
+  // (mapAllocatorFor), so codegen and the reflective path produce the same runtime class for an
+  // interface-typed field. The two hash families default to their insertion-ordered form because a
+  // conversion that is not asked to reorder should not: an ordered source behind an interface-typed
+  // field keeps its order across the rebuild.
   private static String concreteImplFqn(final TypeMirror container, final FieldPlan.Kind kind) {
     final var el = (TypeElement) ((DeclaredType) container).asElement();
     if (el.getKind() == ElementKind.CLASS && !el.getModifiers().contains(Modifier.ABSTRACT)) {
@@ -2441,7 +2442,7 @@ public final class BridgeProcessor extends AbstractTelescopeProcessor {
     return switch (kind) {
       case LIST -> "java.util.ArrayList";
       case SET -> "java.util.LinkedHashSet";
-      case MAP_VALUES -> "java.util.HashMap";
+      case MAP_VALUES -> "java.util.LinkedHashMap";
       default -> throw new IllegalStateException("not a collection/map kind: " + kind);
     };
   }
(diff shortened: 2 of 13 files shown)


eschizoid:
## Measured cost

Two runs of `MapStructComparisonBenchmark.container_map`, `-Pjmh.profilers=gc`, 5 warmup + 10 measured: [34611542315](https://github.com/eschizoid/telescope/actions/runs/34611542315) on `main` and [34611507811](https://github.com/eschizoid/telescope/actions/runs/34611507811) on this branch.

**Read the control first, and it disqualifies the timings.** MapStruct's rows cannot be touched by this change, and they moved 1.200x (forward) and 1.220x (backward) between the two runs — the branch run landed on a slower runner. Any raw before/after nanosecond figure from this pair is that runner difference, not this change.

Normalising to within-run ratios against MapStruct:

| row | main | branch |
| --- | --- | --- |
| codegen forward | 1.130x | 1.014x |
| codegen backward | 0.966x | 0.971x |
| runtime forward | 1.028x | 1.060x |

Backward is steady. Forward reads as *improving*, which is not a credible result of making telescope allocate more — and the published headline run has this row at 1244 ns against MapStruct's 1262 (telescope ahead), so `main`'s 1423.987 today is the outlier. **The time cost is not resolvable from this pair.** I would rather say that than publish a number the control has already invalidated.

**Allocation is decisive, and it is the durable half.** It is deterministic rather than hardware-dependent — visible here in MapStruct's rows reporting an identical 7528.000 B/op across both runs despite the 20% timing swing.

| row | main | branch | delta |
| --- | --- | --- | --- |
| MapStruct (control) | 7528 | 7528 | +0 |
| telescope codegen | 6712 | **7528** | +816 |
| telescope runtime | 6744 | 7560 | +816 |

So the codegen path now allocates **exactly** what MapStruct allocates, where it previously allocated 816 B/op less by building the cheaper container. The runtime path keeps the constant +32 B/op it already carried over MapStruct.

The +816 is mechanically what `LinkedHashMap` costs, not a surprise: 100 entries x 8 bytes for the two extra references each `LinkedHashMap.Entry` carries, plus 16 for the map object's own `head`/`tail`. That it lands on the control's figure to the byte is the evidence this is the container swap and nothing else.

**What this does to the published container figures.** The `container, Map-valued` row of `docs/perf-mapstruct-comparison.md` currently reports **7,528 vs 6,712 B/op** as telescope allocating about 11% less. That row was comparing different containers. If this merges, the honest figure is parity, and the doc's "what is real on that row is the allocation" sentence has to go with it — I have not touched the doc in this PR, since it should only change if the decision is to merge.

PR #362, opened 2026-09-11
perf: size the default impl a raw-container helper allocates

### LINKED ISSUE #346: perf: sizedAlloc misses raw container subtypes in emitRawContainerHelper
`sizedAlloc` (the #308 fix) did not reach `emitRawContainerHelper` (~line 2907): raw container SUBTYPES — `class ImageUrls extends ArrayList<ImageUrl>`, a `HashMap` subtype — rebuild through `rawAllocExpr`, explicitly the no-arg constructor, then fill. That is the resize chain #308 removed everywhere else: log2(n) grows plus a rehash per rebuild for hash-family subtypes. The `sizedAlloc` default branch (~line 2326) has the same edge — any concrete class outside the five JDK names gets `new X<>()`, so a field declared `LinkedHashMap` is sized while a trivial user subtype of it is not.

Presizing a subtype is only safe when the subtype declares or inherits the `(int)` constructor, which is checkable in the processor; otherwise a sized default-impl intermediate plus `addAll` still beats the resize chain for the hash family.

**No benchmark fixture has a raw-subtype container**, so the defect and any fix are both invisible to the suite — the same green-while-wrong shape #308 itself had before the container tier existed. The fixture row (a raw ArrayList subtype and a raw HashMap subtype in `ContainerAllocationBenchmark`/`ContainerWriteBenchmark`) must land with or before the fix; allocation via `-Pjmh.profilers=gc` is the deterministic gate.

Found by the emitted-code audit, traced in `BridgeProcessor`.

eschizoid:
Closes #346.

`sizedAlloc` reached the generic container helpers but not the raw one, so a field pairing an interface with a raw container subtype allocated its default impl from the default capacity and grew it while filling — where the generic path for the same interface sized it from the source. That half is fixed.

## The other half cannot be fixed, and the issue's premise is wrong about why

The issue proposes presizing a subtype "when the subtype declares **or inherits** the `(int)` constructor". **Java does not inherit constructors.** A trivial `class ImageUrls extends ArrayList<ImageUrl> {}` has its implicit no-arg constructor and nothing else — which I verified rather than argued:

```
Trivial    ctors: [[]]
TrivialMap ctors: [[]]
TrivialSet ctors: [[]]
```

So the `(int)` path does not exist for exactly the shape the issue names. And where a subtype *does* declare one, the processor cannot know what the argument means — `Page(int pageNumber)` is indistinguishable from a capacity from inside `javax.lang.model`, and calling it with `src.size()` would be a semantic bug, not an optimisation. So raw subtypes keep the no-arg allocation. That is not a deferral; it is the only correct emission.

The issue's fallback — a sized intermediate plus `addAll` — is also worth skipping, and measuring the JDK says why:

| fill | resulting table/array |
| --- | --- |
| `ArrayList` subtype, `addAll(1000)` | 1000 — exact, one grow |
| `HashMap` subtype, `putAll(1000)` | 2048 |
| `HashMap` subtype, 1000x `put()` | 2048 |

`addAll` already sizes the list exactly, and `putAll` presizes the map, so the identity copy paths were never the problem. The element-bridging loop reaches the same final table as `putAll` does — the cost is the abandoned intermediate tables, which is real but which an extra intermediate container would pay for twice rather than remove.

## What changed

`sizedAlloc` gained a rendered-name parameter so it can emit fully-qualified names for the raw helper, which imports nothing, and `rawAllocExpr` now routes through it. Only the five JDK default impls are sized; every other class falls through to the allocation it had before, so a generic user subtype is untouched.

The generated bridge for the new fixture shows both halves side by side:

```java
// interface side — sized
final var out = new java.util.ArrayList<...RawUrlA>(src.size());
// raw subtype — cannot be
final var out = new ...RawListB();
```

## Verification

Two existing tests failed the moment the fix landed, and they are precisely the two mixed generic-to-raw cases it targets — which is the evidence the defect was reachable rather than theoretical. Both now assert the sized form **and** assert the unsized form is absent, so a regression that emitted both would not pass.

## The benchmark

#346 requires the fixture to land with the fix, because no fixture had a raw-subtype container and the defect was invisible to the suite — the same green-while-wrong shape the container tier itself was added to catch.

`RawContainerBenchmark` varies cardinality, which is the dimension that exposes an unsized rebuild: a container filled element by element from its default capacity grows a logarithmic number of times, and every abandoned table counts toward allocation. At one element nothing grows, so the low rows are the control for the high ones — a change that moved every row equally would be measuring something other than sizing. The `plain` field is deliberately mixed, so one direction allocates a sizable default impl and the other a subtype that cannot be sized, and a single pair of rows carries both.

Runs dispatched on this branch and on a baseline branch carrying the fixture against an unfixed emitter, since a benchmark that exists on only one side has nothing to measure. Numbers to follow.

### DIFF
--- a/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/RawContainerBenchmark.java
+++ b/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/RawContainerBenchmark.java
@@ -1,0 +1,62 @@
+package io.github.eschizoid.telescope.benchmarks;
+
+import java.util.ArrayList;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+/**
+ * Generated conversion of raw container subtypes, across cardinality.
+ *
+ * <p>Cardinality is the dimension that makes an unsized rebuild visible: a container filled element
+ * by element from its default capacity grows a logarithmic number of times, and every abandoned
+ * table counts toward the allocation the {@code gc} profiler reports. At one element nothing grows,
+ * so the low rows are the control for the high ones — a change that shifted every row equally would
+ * be measuring something other than sizing.
+ *
+ * <p>Run with {@code -Pjmh.profilers=gc}: allocation is deterministic and survives a runner
+ * difference, where the timings do not.
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Mode.AverageTime)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+public class RawContainerBenchmark {
+
+  @Param({ "1", "16", "256", "4096" })
+  public int size;
+
+  private RawHolderA source;
+  private RawHolderB target;
+
+  @Setup
+  public void setup() {
+    final var urls = new RawListA();
+    final var index = new RawMapA();
+    final var plain = new ArrayList<RawUrlA>(size);
+    for (int i = 0; i < size; i++) {
+      urls.add(new RawUrlA("u" + i));
+      index.put("k" + i, new RawUrlA("u" + i));
+      plain.add(new RawUrlA("u" + i));
+    }
+    source = new RawHolderA(urls, index, plain);
+    target = RawHolderABridge.forward(source);
+  }
+
+  /** Both raw-subtype sides plus the interface-to-subtype direction of the mixed field. */
+  @Benchmark
+  public RawHolderB forward() {
+    return RawHolderABridge.forward(source);
+  }
+
+  /** The reverse, where the mixed field's output is a JDK default impl that can be sized. */
+  @Benchmark
+  public RawHolderA backward() {
+    return RawHolderABridge.backward(target);
+  }
+}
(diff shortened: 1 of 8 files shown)


eschizoid:
## Measured

`RawContainerBenchmark`, `-Pjmh.profilers=gc`, 5 warmup + 10 measured: [34618634342](https://github.com/eschizoid/telescope/actions/runs/34618634342) on the unfixed baseline branch, [34618637493](https://github.com/eschizoid/telescope/actions/runs/34618637493) on this one.

**The control is `forward`, and it is byte-identical across both runs.** That row is all raw subtypes, which this change cannot size, so it should not move — and it does not, on any cardinality:

| size | baseline | fixed |
| ---: | ---: | ---: |
| 1 | 392 | 392 |
| 16 | 2104 | 2104 |
| 256 | 33768 | 33768 |
| 4096 | 493864 | 493864 |

That is what scopes the result: the same benchmark, same fixture, same cardinalities, and the untouched path did not shift by a single byte.

**`backward`, which allocates a sizable default impl for the mixed field, drops at every cardinality:**

| size | baseline | fixed | delta |
| ---: | ---: | ---: | ---: |
| 1 | 392 | 360 | -32 (-8.2%) |
| 16 | 2104 | 1944 | -160 (-7.6%) |
| 256 | 33768 | 30288 | -3480 (-10.3%) |
| 4096 | 493864 | 460048 | -33816 (-6.8%) |

It wins at size 1 too, where nothing resizes: `new ArrayList<>()` grows to the default capacity of ten on first add, so a single-element list over-allocates whether or not a chain follows.

**On timing, read the control first and then discount it.** `forward` moved -9.9% at 4096 between the two runs while its allocation stayed byte-identical — so that is a faster runner, not this change, and the raw before/after nanoseconds are not a measurement. Normalising to the within-run `backward/forward` ratio: 1.018x -> 0.836x (size 1), 1.000x -> 0.933x (256), 0.993x -> 0.937x (4096). Size 16 goes the other way at +7.4%, but its bands (+/-16.0 against +/-11.7 on a ~380 ns row) overlap heavily and it resolves nothing. Three rows consistent with the allocation result and one unresolved is what this pair supports — the allocation table above is the durable half.

Issue #345, opened 2026-09-11
perf: generated navigators rebuild invariant optics on every call, and no benchmark can see it


eschizoid:
Every navigation call on a generated navigator rebuilds invariant optics: `<X>Telescope.id()` emits `path.then(Telescope.lens(X::id, (s, v) -> new X(...))).hop(new OpticNode.Focus("id"))` per invocation — a fresh Lens, a fresh composed Telescope, a fresh OpticNode, and hop's trail-copy — although the byte-identical lens already exists as `<X>FieldOptics.id`. Container steps likewise route through `Telescope.asList(path).each()` per call. Composition allocation is inherent to the fluent API; the lens construction and the OpticNode are invariants.

Constraint from the prior rejected fix: do NOT hoist per-field work into class-level statics that widen the navigator's `<clinit>` — one missing optional class then kills every field (the blast-radius finding that closed the earlier navigator issue). The acceptable shapes are nested per-field lazy holder classes (one `<clinit>` per field, first-use init) or lazy per-field paths. Secondary: the setter lambdas are duplicated between navigator and FieldOptics — two LMF call sites and two lambda bodies per field.

**No benchmark can see this either way**: every existing benchmark pre-builds paths at `@Setup`, so per-call navigator fluency is invisible to the whole suite. The measurement must land first — a row driving `<X>Telescope.of().field().read(x)` inside the hot loop with a prebuilt path as the control — then the A/B on CI with the gc profiler, per the house rule. Without the new row, neither the win nor a regression is observable, which is how the last three perf defects shipped.

PR #364, opened 2026-09-11
bench: measure navigating a generated navigator per call

eschizoid:
Groundwork for #345, which does not close here.

#345 asks for the measurement to land before the emitter change, because every existing benchmark builds its path in `@Setup` — so what a fluent navigation call costs is invisible to the entire suite. That is the same green-while-wrong shape that let the last three perf defects ship, and it means neither a win nor a regression from changing the navigator emission would be observable today.

## What it measures

The generated navigator rebuilds its optics on every call. From the emitted `BenchHolderSrcTelescope`:

```java
public Telescope<R, String> name() {
  return path.then(Telescope.lens(BenchHolderSrc::name, (s, v) -> new BenchHolderSrc(v, ...)))
             .hop(new OpticNode.Focus("name"));
}
```

A fresh `Lens`, a fresh composed `Telescope`, a fresh `OpticNode`, and `hop`'s trail copy — per invocation — although a byte-identical lens already exists as a constant on the sibling `BenchHolderSrcFieldOptics`.

## Design

Each navigator row has a **prebuilt-path row beside it** reading the same field of the same value. The read is common to both, so the difference is what the fluent call adds; a change that moved both rows equally would be measuring the read rather than the construction.

**Depth is the dimension**, and it is the one that discriminates. One hop and three hops do the same amount of reading but build a different number of intermediate optics, so a per-call construction cost appears as a gap that widens with depth, where a read cost would not. The chain is `department().address().city()` against `name()`.

No production code. Numbers to follow once the run lands; the fix is a separate PR so it has a before side to be measured against.

### DIFF
--- a/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/NavigatorFluencyBenchmark.java
+++ b/benchmarks/src/jmh/java/io/github/eschizoid/telescope/benchmarks/NavigatorFluencyBenchmark.java
@@ -1,0 +1,70 @@
+package io.github.eschizoid.telescope.benchmarks;
+
+import io.github.eschizoid.telescope.Telescope;
+import java.util.concurrent.TimeUnit;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+
+/**
+ * Navigating a generated navigator inside the measured loop, which no other benchmark does: every
+ * existing row builds its path once in {@code @Setup}, so the cost of the fluent call itself is
+ * invisible to the whole suite.
+ *
+ * <p>Each navigator row has a prebuilt-path row beside it reading the same field of the same value.
+ * The pair is what separates the two costs: the read is common to both, so the difference is what
+ * the fluent call adds, and a change that moved both equally would be measuring the read.
+ *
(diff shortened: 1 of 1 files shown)


eschizoid:
## Baseline

[Actions run 34624149553](https://github.com/eschizoid/telescope/actions/runs/34624149553), `-Pjmh.profilers=gc`, 5 warmup + 10 measured. This is the before side; nothing is fixed yet.

| row | ns/op | B/op |
| --- | ---: | ---: |
| `prebuiltShallow` (1 hop, path built once) | 2.633 ± 0.008 | ~0 |
| `navigateShallow` (1 hop, navigated per call) | 38.567 ± 0.177 | 232 |
| `prebuiltDeep` (3 hops, path built once) | 25.066 ± 0.085 | 112 |
| `navigateDeep` (3 hops, navigated per call) | 186.703 ± 2.134 | 904 |

**Navigating fluently costs 14.6x a prebuilt read at one hop**, and 7.4x at three. In absolute terms the call adds 35.9 ns and 232 B/op at one hop, and 161.6 ns and 792 B/op at three.

The one-hop control allocates **essentially nothing** — the profiler reports ~10⁻⁶ B/op — against 232 B/op for the identical read reached fluently. Every byte of that is the path being rebuilt.

**The gap widens with depth, which is the discriminator the design was built around.** A read cost would scale with the work done per element and show up in both rows; a per-call *construction* cost scales with the number of hops built, and only in the navigator rows. Time added goes 35.9 -> 161.6 ns and allocation added goes 232 -> 792 B as the path goes from one hop to three. That is construction, not reading.

So the defect #345 describes is real and considerably larger than "an extra allocation": at one hop the fluent call is fourteen times the cost of the read it performs. The suite could not see any of this before, because every other row builds its path in `@Setup`.

Bands are tight (±0.008 to ±2.134 on their respective magnitudes) and allocation is deterministic, so this baseline is usable as a before side without a re-run. The fix lands separately and gets measured against exactly these rows.

Issue #371, opened 2026-09-11
codegen: FieldOptics builds every field in one clinit, so one unresolvable type disables the runtime path for all of them


eschizoid:
`<X>FieldOptics` builds every field's constant in one `<clinit>`, so a single unresolvable type takes down the whole holder — and with it the runtime reflective path for every field of that type, not just the one that cannot be resolved.

Measured on a generated `@Focus record User(String name, Gadget gadget, Widget widget, int age)` with `Gadget.class` deleted from the runtime classpath and the classes loaded through a filtered classloader:

```
UserTelescope.of().name()            -> OK
Telescope.of(User.class).field(User::name)
  -> NoClassDefFoundError: demo/Gadget
     at MetadataHolderProbe.readConstants(MetadataHolderProbe.java:112)
     at Telescope.field(Telescope.java:921)
```

Reading `name` fails on a type only `gadget` mentions. The codegen navigator contains the same failure; the runtime path does not, because it reads the holder.

## Why it matters

This is the exact blast radius #345 rejected an earlier navigator fix for, still present one class over. It is also the asymmetry that makes the guarantee awkward to describe: after #370 the navigator contains a missing optional type, so "one bad field cannot kill the others" is true of `<X>Telescope` and false of `<X>FieldOptics`.

The shape that a missing type is plausible in — an optional dependency present at compile time and absent at runtime — is the one ADR-0006's probe exists to serve, since the probe's whole point is to short-circuit when the holder is present and fall back when it is not. Here the holder is present and its initializer is what fails, so the fallback never engages.

## Fix shape

The same treatment #370 applied to the navigator: one nested private holder per field rather than one initializer for all of them, so a field that cannot resolve fails alone and `MetadataHolderProbe` falls back to the LMF path for it while the others keep their constants.

Worth checking whether the probe should also catch `NoClassDefFoundError` / `ExceptionInInitializerError` around its read and degrade to the reflective path, rather than propagating — that would make the failure recoverable even where per-field splitting does not reach.

## Verification

A test that compiles a `@Focus` type, removes one component's type from the runtime classpath, and asserts the *other* components still resolve through `Telescope.of(...).field(...)`. It has to run through a classloader that genuinely lacks the class; an in-memory compilation alone will not reproduce it, since the failure is at link time rather than compile time.

Found while empirically verifying #370's containment claim, which is what established the navigator half is now safe.

eschizoid:
[...]

PR #372, opened 2026-09-11
fix: treat a holder that cannot initialize as one that is absent

eschizoid:
Small, and deliberately **not** closing #371 — the measurement that came out of investigating it overturned that issue's diagnosis, and this is the part that survived.

## What changed

`MetadataHolderProbe` already answers `Optional.empty()` when no `<X>FieldOptics` is on the classpath, and the caller takes the reflective path. A holder that *is* present but whose class initializer cannot complete propagated the `LinkageError` instead and took the call down with it.

The holder exists to skip the reflective build. It is an optimization, not a dependency, so an unusable one should cost the fast path and nothing else. Both cases now answer the same way.

## What this does not do

I reached this from #371, which says `FieldOptics` building every field in one `<clinit>` is what disables the runtime path when one component type is missing. **That diagnosis is wrong.** With `Gadget` removed from the runtime classpath:

```
codegen  navigator name() -> OK
codegen  navigator age()  -> OK
runtime  field(User::name) -> FAIL NoClassDefFoundError
    at java.base/java.lang.Class.getRecordComponents0(Native Method)
    at io.github.eschizoid.telescope.internal.Records$RecordInfo.of(Records.java:331)
```

`Class.getRecordComponents()` resolves every component's type, so reflecting on the record requires the missing class regardless of what the holder does. The original report's trace stopped at the probe because the probe was the *first* thing to need the type — not the only one. With this change the probe steps aside and the reflective fallback fails two frames later, for a reason telescope does not own.

So splitting `FieldOptics` per field would change nothing observable, and I have said so on #371 rather than leave the issue pointing at work that would not help.

## Verification

A hand-written holder fixture whose initializer throws, alongside the existing malformed-holder fixtures — the established pattern in that suite, and stable across processor regenerations. It asserts the probe answers empty **twice**, because the second read is where the JVM reports the erroneous class as `NoClassDefFoundError` rather than `ExceptionInInitializerError`, and `LinkageError` has to cover both.

Revert-proven: narrowing the catch to `RuntimeException` fails that test alone.

`:internal` 345 PASSED, full suite green, run with `--rerun-tasks` after `--stop`.

One honest note on scope: for a *missing component type* specifically, this branch is unreachable in practice — the JDK fails first. What it does cover is a holder that cannot initialize for any other reason, which is the case the malformed-holder suite around it already exists to police.

### DIFF
--- a/internal/src/main/java/io/github/eschizoid/telescope/internal/MetadataHolderProbe.java
+++ b/internal/src/main/java/io/github/eschizoid/telescope/internal/MetadataHolderProbe.java
@@ -94,6 +94,13 @@ public final class MetadataHolderProbe {
       return Optional.of(new HolderRef(holder, constants, constructor));
     } catch (final ClassNotFoundException e) {
       return Optional.empty();
+    } catch (final LinkageError e) {
+      // The holder exists to skip the reflective build, so it is an optimization and not a
+      // dependency: one that cannot initialize is a reason to take the slower path, the same answer
+      // already given when no holder is on the classpath at all. A type present when the holder was
+      // compiled and absent at run time arrives here, and it carries every field of its owner with
+      // it, because the holder builds them all in one initializer.
+      return Optional.empty();
     } catch (final ReflectiveOperationException e) {
       throw new IllegalStateException("Failed to probe metadata holder " + holderName, e);
     }
@@ -124,8 +131,8 @@ public final class MetadataHolderProbe {
       throw new IllegalStateException(
         "Metadata holder " +
           holder.getName() +
-          " has a `constants()` method but its shape is wrong (must be `public static Map<...>`). " +
-          "Re-run the @Focus / @BeanFocus processor."
+          " has a `constants()` method but its shape is wrong (must be `public static" +
+          " Map<...>`). Re-run the @Focus / @BeanFocus processor."
       );
     }
     final var result = (Map<String, Object>) method.invoke(null);
@@ -157,7 +164,8 @@ public final class MetadataHolderProbe {
           holder.getName() +
           " is missing the required `public static " +
           target.getSimpleName() +
-          " construct(Function<String, Object>)` method. Re-run the @Focus / @BeanFocus processor.",
+          " construct(Function<String, Object>)` method. Re-run the @Focus / @BeanFocus" +
+          " processor.",
         e
       );
     }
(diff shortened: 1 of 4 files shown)


eschizoid:
Closing unmerged. The review falsified the premise this rests on, and the right answer is that the throw it replaced was correct.

**The argument was "the holder is an optimization, so failing to use it should cost the fast path and nothing else."** The second half is false for beans. Measured on a `@BeanFocus` bean with both a normalising builder and plain setters, same class, same call:

```
holder INTACT     probe -> PRESENT   set(code,'xy') -> Bean2[name=alice, code=XY]
holder CORRUPTED  probe -> EMPTY     set(code,'xy') -> Bean2[name=alice, code=xy]
```

The generated holder rebuilds through `Bean2.builder()…build()`, which normalises; the reflective `BeanWriter` picks the setters and bypasses it. So the degraded path returns a **different value**, silently. Before this change that scenario threw and named the holder. Records are value-identical under degradation — a record has one rebuild strategy on both paths — which is why the fixture here could not see it.

**Second, it swallows the drift the neighbouring code exists to shout about.** Measured `main` vs this branch: `NoSuchMethodError` (a setter or builder method renamed since the holder compiled) and `UnsupportedClassVersionError` (holder newer than the runtime) both went from a precise, actionable error to `Optional.empty()`. Six lines away in the same file, a *milder* version of the same drift — a missing `constants()` rather than a broken one — raises `IllegalStateException` with "Re-run the @Focus / @BeanFocus processor." I asked for this to be attacked specifically, and it is a real regression in diagnostics.

**Third, it reverses the sibling probe's stated policy.** `BridgeHolderProbe` documents the opposite choice for the identical exposure: malformed-holder exceptions "are NOT memoised — they re-throw on every call so codegen drift surfaces consistently rather than hiding behind a cached failure."

Narrowing the catch to `ExceptionInInitializerError | NoClassDefFoundError` would fix the second and third points but not the first, so it does not rescue the change.

Two further problems with my own test, for the record: its second assertion proves nothing (a counting classloader shows `ClassValue` serves the second read without touching the class, so the "arrives in its other form" comment states a property that is not true), and its name claims the holder "degrades to the reflective path" while the body only asserts `isEmpty()` and never reads through that path — which is exactly the assertion that would have gone red for a bean.

The genuinely useful result is the one that falsified this, and it is bigger than this PR: **the holder path and the reflective path are not semantically equivalent for beans with builders.** Filed separately.

The original investigation's finding also stands and is recorded on #371: a missing component type fails at `Class.getRecordComponents`, not at the holder, so neither this nor per-field splitting addresses it.

PR #403, opened 2026-09-16
fix: do not choose a write surface that cannot carry the whole bean

### LINKED ISSUE #398: fix: the reflective writer drops properties it cannot set, where generated code preserves them
A bean with a no-argument constructor, a builder, and setters for only *some* of its properties is rebuilt differently by the two paths, and the reflective one loses data.

Measured on a bean `{name=alice, code=AB}` with a setter for `name` only, writing `name`:

```
generated (builder)   -> name=bob  code=AB
reflective (setters)  -> name=bob  code=null
```

The reflective writer prefers setters as soon as *any* exist, and installs a no-op writer for each property it cannot write. The property is not merely written differently — it is dropped, and the rebuilt bean carries a null where the source had a value. Nothing is raised.

## Why this is the more serious half

A sibling issue covered a bean with a *complete* setter surface, where the two paths disagreed about normalisation: one applied the builder's `build()` and the other did not, so both produced a populated bean and the values differed. That was fixed by aligning the generated path with the reflective one.

This case cannot be fixed the same way, because here the reflective path is the one that is wrong. Aligning generated code with it would mean adopting the data loss. So the generated path deliberately keeps the builder for this shape, and the two continue to disagree — with the generated side preserving the value.

## What would close it

The reflective writer needs to stop preferring a setter surface it cannot fully use. Its own rule asks whether *any* setter exists; asking whether *every* property has one would put it on the builder for this bean, which is what the generated path already does. That is a behavioural change to the reflective path and needs its own measurement of which beans it moves.

Worth deciding at the same time whether a partially-writable bean should be rejected outright rather than silently rebuilt at all, since neither surface can reproduce it faithfully.

eschizoid:
Closes #398.

The reflective writer preferred setters as soon as **one** existed, then installed a no-op for every property it could not write. Rebuilding is how a write works, so writing any property through such a bean dropped the others:

```
before:  set(name, "bob")  ->  PartialBean[name=bob, code=null]     <- code was "AB"
after:   set(name, "bob")  ->  PartialBean[name=bob, code=AB]
```

Nothing was raised either way.

## What does and does not change

The skip is right where nothing better exists. A getter-only property — a computed field, an immutable accessor — has no setter and never will, and leaving it at its default is what other mappers do and what the code already documents. That case is untouched.

It stops being right the moment a builder is also on offer that can carry the value. That is the only case this changes. Setters are still preferred wherever they cover the bean, and still preferred over field injection where they do not but no builder exists.

## Behavioural note

A bean with a partial setter surface **and** a builder now rebuilds through the builder. A value that was silently nulled is now preserved, so anyone whose code depended on the null will see a change — though depending on it seems unlikely, since the null was data loss rather than a documented default.

I have not marked this breaking. Worth a second opinion on that.

## This class of bug has a scar here already

`hasAnySetter` carries a note about an unrelated `set`-named method flipping the strategy, "whose per-property no-ops then silently lost every field of a getter-only bean". That was fixed by tightening what counts as a setter. This tightens how many of them it takes.

## Tests

`PartialBean`'s `code` has a builder method and no setter, which is what makes the tests able to fail — where setters cover a bean, both surfaces reproduce the same value and every assertion holds whichever is chosen.

Revert-proof: restoring "any setter wins" fails both write tests with `expected: <AB> but was: <null>` and leaves the read control green. A read never rebuilds, so it reports the property present whichever surface a write would have used — which is exactly what kept this quiet.

Full suite: 1432 passing. No in-tree fixture changes strategy.


eschizoid:
Closing unmerged. A fixture battery found four regressions, one of them strictly worse than the loss this fixes: for a bean whose setters and builder each reach a different property, the requested write is now the one silently discarded.

The root error is that the predicate is one-sided. It asks whether the setters cover every property, when the decision is a choice between two surfaces and the question has to be whether the builder covers at least as much. The builder surface skips silently too, so the change relocated which property gets dropped rather than making the loss loud.

Everything measured is recorded on #398, including two shapes that went from working to throwing, the loss of no-arg-constructor side effects, and the fact that a scan of every compiled class in the repository found exactly one type whose strategy changes — the fixture this branch added. That last point is why relanding should start from the battery rather than end with it.

Not abandoning the fix. The data loss is real and worth closing; this particular predicate is not the way.

Issue #379, opened 2026-09-12
codegen/runtime: container allocation policy is implemented twice, where ADR-0012's shared spec exists for exactly this


eschizoid:
ADR-0012 established the pattern for a decision both the runtime and the processor must make: put it in `internal/pairing` once, and let each side supply a `PropertySystem<T>` so the same rules run over reflection handles or over `javax.lang.model` mirrors. `PairingRules` + `PropertyNames` + `PairingMessages` are consumed today by `DeepMap`, `Merge`, `MirrorProps`, `AbstractTelescopeProcessor` and `MapperVerifierProcessor` — one implementation, five consumers, no drift.

**Container allocation policy has never been given that treatment, and it is the same shape of decision.** It lives in two independent places:

- `BridgeProcessor.concreteImplFqn` — interface family → implementation, emitted as text
- `ContainerLifts.setAllocatorFor` / `mapAllocatorFor` — the same decision as a `Function<Object, Object>`

Plus the policy layered on top of it: which allocations can be sized, which carry a comparator, and which pairs are refused.

## The evidence that this costs something

Three separate defects in one day, each the same root cause — two implementations of one decision, drifting:

1. **#373** — the holder path rebuilds a bean through its builder, the reflective path through setters. Different values, silently.
2. **#376 / PR #378** — five interface families were added to the processor's table; the runtime's lacked them, so the same DTO compiled under `@Bridge` and threw `IllegalStateException` under `mapper(...)`.
3. **The comparator, in the same PR** — codegen emitted `new TreeMap<>()` and dropped the ordering the runtime preserved, crashing for non-comparable keys. Then a compile-time refusal I added rejected five shapes the runtime accepts, and the fix was to emit a hand-written copy of `ContainerLifts.checkComparator` as a string literal.

That third one is the tell. The current state of the branch has the runtime's policy written **twice**: once as Java in `ContainerLifts`, once as a string the processor emits. They are kept equal by a test asserting the two messages match — which works, and which is exactly the kind of coupling a shared spec exists to remove.

## The infrastructure is already there

`ContainerView`'s javadoc says the raw class handle is *"carried so lifted conversions can allocate the declared concrete class"*, and `PairingRules.containerViewOf` exists *"so the kind selection rules live once"*. The seam was built anticipating this and neither side allocates through it.

## Shape

A `ContainerAllocation` (or an extension of `PairingRules`) in `internal/pairing` answering, for a `ContainerView<T>`:

- which class to allocate
[...]

PR #407, opened 2026-09-17
fix: allocate for every container the generated path allocates for

### LINKED ISSUE #385: fix: four declared container types are accepted by @Bridge and refused by the reflective mapper
Four declared container types are accepted by `@Bridge` and rejected at runtime by `Telescope.mapper(...)`. The two are meant to be interchangeable for the same type pair, so either can be swapped for the other — except on these.

| declared field type | `@Bridge` | `Telescope.mapper(...)` |
| --- | --- | --- |
| `Hashtable<K, V>` | emits `new java.util.Hashtable<>()`, works | throws `no allocator for Map subtype java.util.Hashtable` |
| `AbstractList<E>` | abstract, so the family default `ArrayList` applies and is assignable | throws |
| `AbstractSet<E>` | family default `LinkedHashSet`, assignable | throws |
| `AbstractMap<K, V>` | family default `LinkedHashMap`, assignable | throws |

## Why the runtime refuses

`ContainerLifts` names about thirty raw classes across its three allocator methods, and falls back to `Beans.intermediateAllocator` for anything it does not name. That fallback cannot bind a `java.base` class, because `MethodHandles.privateLookupIn` refuses one, so every JDK container the table omits reaches the throw. `Hashtable` is simply absent from the table. The three abstract types have no accessible constructor either, so even a table row would not help them — they need the family default, which is what codegen already does for them.

Codegen has no such constraint: it writes the allocation as source text and lets javac bind it, so one rule covers every instantiable class at once.

## What would close it

Two independent pieces, either useful alone:

The runtime deciding "is the declared type instantiable?" rather than "is it in my table?" fixes the three abstract types, because it routes them to the family default exactly as codegen does.

A `MethodHandles.publicLookup().findConstructor` fallback before the throw fixes `Hashtable` and every other public JDK container with a public no-arg constructor. `publicLookup` does not need `privateLookupIn`, so the wall the table works around is not there. Worth noting in whichever change lands: a constructor found this way needs reachability metadata under native image, where the hard-coded allocators are safe by construction — so the family defaults should stay direct and only the tail should use the fallback.

## Why no test caught it

Every existing parity test names its container types by hand, so the set it covers is the set someone thought of. A test that enumerates the JDK's own container classes from the `jrt:` filesystem and asserts the runtime allocator agrees with codegen's rule would be exhaustive over `java.base`, deterministic, and fail on exactly these four today. That test is worth writing before the fix, since it is the thing that proves the fix did something.

## Related

The selection rule these two sides disagree about is the one #379 proposes to share.

eschizoid:
Closes #385, and finds three more than it named.

The generated path and the reflective path decide what to allocate for a container field separately. Seven `java.base` types came out differently — the generated path allocates them, the reflective path threw:

```
AbstractList, AbstractSet, AbstractMap    abstract; the generated path uses the family default
Hashtable, Properties                     concrete; the generated path allocates them directly
SequencedMap, SequencedSet                interfaces added in Java 21
```

A bean using one of these compiles under `@Bridge` and throws under `mapper(...)`, which is the swap the two paths exist to make interchangeable.

## The gate came first, and it is the part worth keeping

The issue named four, because four is what somebody thought of. This enumerates container types from `java.base` itself, so a type nobody remembered is still checked — the `Sequenced` interfaces in particular would not have been found by inspection.

It restates the generated path's rule as the thing the reflective path has to agree with: allocate the declared class where one can be made, otherwise the family default, and only where that default actually fits the declaration. That last clause is what keeps the gate honest — `AbstractSequentialList`, `EnumSet` and `ConcurrentNavigableMap` all surfaced on the first run and are correctly excluded, because no family default is assignable to them and the generated path cannot handle them either. Demanding parity there would be demanding more than parity.

Restating the rule rather than importing it is this shape's cost: the two could drift. Two lines guarding thirty-odd types is a trade worth making until the rule has one home, which is what #379 is about.

## The fix

Two questions were missing, both of which the generated path already asks.

A declared type that cannot be instantiated at all is asking for whatever implements its contract, so the family default stands in.

A concrete one gets a public-lookup constructor handle. The hard-coded table exists because `privateLookupIn` refuses `java.base`; `publicLookup` does not need it and binds a public no-argument constructor on any exported class.

The table keeps every common shape direct, so only the tail reaches the handle. That matters under native image, where a constructor found this way needs reachability metadata and a hard-coded allocator needs none.

## Verification

All five representative types now map end to end, not merely allocate — `Hashtable`, `AbstractList`, `AbstractSet`, `AbstractMap` and `SequencedSet` each produce a correctly converted target.

Revert-proof: removing the fallback from all three families turns the corpus gate red.

Full suite: 1458 passing.


eschizoid:
Round 1 of review found this creating five divergences of its own, in the opposite direction to the ones it closes, and a gate that could not see them.

A declared type no family default implements — `AbstractSequentialList`, `ConcurrentNavigableMap`, an adopter's own interface over `List`/`Set`/`Map` — was refused by both paths before. After the first commit the generated path still refused at compile time while the reflective path built a plan and threw a bare `ClassCastException` on the first conversion. That is the worse direction: the refusal named the type and said what to do instead, and it arrived while the mapper was being built, which is where both starters catch it — each constructs every mapper at container start so a bad pairing fails there rather than per request.

The fallback now stands in only where the default is actually one of the declared type.

The gate had the same defect one level up. It used the rule as a **filter**, skipping the types the generated path refuses, so it asserted nothing about exactly the types the fix got wrong — it passed identically with and without the guard. It now asserts in both directions, checks the refusal arrives at plan time rather than as a cast error later, and carries adopter-shaped containers, which no scan of the JDK reaches.

Rewriting it turned up one more, which nothing else had: the restated rule omitted codegen's family-specific defaults, so it called `SortedSet`, `NavigableSet`, `SortedMap`, `NavigableMap` and `ConcurrentMap` refused when the generated path allocates `TreeSet`/`TreeMap`/`ConcurrentHashMap` for them.

Also corrected from the PR body above: **`Properties` is not fixed by this change.** It never reaches `mapAllocatorFor` — a type with no type arguments yields no container view, so its only route is `mapCopyIso`, which this does not touch. The original claim came from the gate calling the lift directly and bypassing the pairing decision. Six types are fixed, not seven.

Family defaults are now allocated presized, matching what the generated path emits for the same declaration.

Issue #404, opened 2026-09-16
codegen: three @Transform shapes the fit check cannot see


eschizoid:
Three shapes reach the `@Transform` fit check and pass it, then fail inside the generated file. They are not cases the check gets wrong; they are cases it cannot see. All three behave identically before and after the check learned to validate both directions, so none is a regression.

## A using class implementing the raw `BridgeFn`

The block short-circuits when the interface instantiation carries no type arguments, so nothing reaches the part that resolves the class's own methods. Those methods resolve perfectly well — a class implementing the raw interface still declares concrete `forward` and `backward` — so the information to decide is available and simply never consulted.

Relaxing the arity guard closes it, but the fallbacks that read the type arguments then need a raw-case path, and the diagnostic has to say something sensible when there are no arguments to name. That is its own behaviour change with its own test surface.

## Two applicable overloads whose parameters are unordered

`forward(CharSequence)` alongside `forward(Comparable<?>)`, called with a `String`. Neither parameter is more specific than the other, so the resolver keeps whichever it saw first, while javac calls the emitted call site ambiguous. The resolver needs the same answer javac gives — that there is no answer — rather than a first-past-the-post one.

## A widening-primitive overload

`forward(long)` alongside `forward(Integer)`, called with an `int`. Javac resolves this in its strict-invocation phase, where widening beats boxing. Assignability models loose invocation and cannot order `long` against `Integer` at all.

## Why they group

Each is the resolver reaching a point where it needs javac's own overload-resolution semantics rather than an approximation of them, and the approximation is right for every shape outside these three. Worth deciding together whether to model those semantics further or to refuse the shapes outright with a diagnostic saying the row is too subtle to verify.

A narrower relative also exists and is not worth its own issue: a package-private overload on a transform in a different package from the source is invisible to the generated bridge, exactly as a private one is, and only the private case is currently skipped.

## What would make any of this cheaper to work on

The verdict matrix that found all of these — fn shape, source field type, target field type, expected accept or reject — exists only as a throwaway probe. Landing it as a parameterised test compiled through the full pipeline would give the next person to touch this check every answer in one run. Three separate rounds of review were needed to establish what it already knows.

eschizoid:
Narrowing this to two shapes. The raw-`BridgeFn` one is fixed in the linked pull request; the other two are not, and one of them I briefly and wrongly reported as closed.

[...]

PR #409, opened 2026-09-17
fix(codegen): check a raw BridgeFn through the methods it declares

eschizoid:
Closes the first of the three shapes in #404, and retires the other two — both turn out to be handled already.

## The live one

The fit check required two type arguments before it would look at anything, so a class implementing the raw `BridgeFn` passed straight through unchecked:

```java
@SuppressWarnings("rawtypes")
public final class Fn implements BridgeFn {
  @Override public Object forward(final Object o) { return o; }
  @Override public Object backward(final Object o) { return String.valueOf(o); }
}
```

Over a `String` to `String` pair that produced `incompatible types: Object cannot be converted to String` inside the generated file — the failure this check exists to replace. The one shape with nothing to fall back on was also the one shape not checked at all.

Rawness was never what made it undecidable. The class still declares the two methods the emitted call sites bind to, and those are what the check compares. It only needed to stop demanding arguments it does not use when a member resolves. Where neither a resolved member nor a type argument describes a direction there is genuinely nothing to decide, and that is skipped rather than guessed.

Both halves are tested. Refusing every raw implementation would satisfy the mismatch test while breaking the fitting one — an unchecked conversion is the defect, rawness is not.

## The other two, re-measured

Both were reported against an earlier commit, before the check resolved members rather than reading the interface's type arguments. Re-running them on `main`:

**Ambiguous parameters** — `forward(CharSequence)` alongside `forward(Comparable<?>)` with a `String` argument — now produces a proper diagnostic rather than a raw `reference to forward is ambiguous`.

**The widening-primitive overload** — `forward(long)` alongside `forward(Integer)` over an `int` field — is **accepted, and correctly so**. `forward(long)` binds, returns `String`, and the `String` slot takes it; backward converts through `Integer`. The emission compiles and the conversion works. Accepting it is the right answer, not a miss.

So #404 described three shapes and one of them was real by the time it was written. The member-resolution work closed the other two as a side effect.

Revert-proof: restoring the arity guard fails the raw-mismatch test and leaves the raw-fitting control green.

Full suite: 1459 passing.


eschizoid:
**Correcting this pull request's own body.** It claimed the other two shapes in #404 were handled. A derive-first review refuted both, and the way it refuted the first is worth recording.

**The ambiguous-overload shape is not closed.** I re-measured it and read a diagnostic, and concluded the shape was handled. The diagnostic said `does not fit backward` — a *different* defect in my fixture firing first, with the ambiguity untouched. Narrowed so only the ambiguity remains, the shape still reaches emission with `reference to forward is ambiguous` on `main` and on this branch. That is the mantra-7 trap in its purest form: an assertion that held on both sides of the thing I thought I was measuring.

First-past-the-post selection is still live and still order-dependent. The same class with two overloads returning different types gives opposite verdicts depending on which is declared first, because the tie-break cannot order `CharSequence` against `Comparable`. The existing order-independence test uses `CharSequence` against `Integer`, where only one candidate is applicable — so it never reaches the unordered case it appears to cover.

**The widening-primitive shape: my reading of the fixture was right, my conclusion was not.** `forward(long)` alongside `forward(Integer)` over an `int` field is accepted, and accepting it is correct — the emission compiles and converts. But it is right by luck rather than by modelling: both overloads return a fitting type, so any choice yields "fits". Change one return and the same shape produces an accept-then-fail, and flipping declaration order alone turns a working pair into a false reject. The original diagnosis stands.

So #404 goes from three shapes to two, not to zero. This closes the raw-`BridgeFn` one only, and #404 stays open for the other two, which are one defect: the check approximates javac's overload resolution, and the approximation is right everywhere except where two candidates are genuinely unordered.

Two further regressions from round 1 are fixed in the commit above — a setter wider than its getter, and a package-private member in another package. Both were already wrong for typed transforms; checking raw ones is what made them reachable by a second route.

eschizoid:
Round 2 of review found the write-slot fix from round 1 was a symptom fix, and it has been reverted.

Comparing against a setter's parameter is right only when setters are the rebuild strategy. A target with a builder or a name-matched constructor that also declares a wider setter was judged against a member the emission never writes through — so the row was accepted and the generated file then failed to compile, which is the failure this check exists to replace. It also went order-dependent, because the setter lookup takes the first of its name while javac binds the most specific.

Two false rejects traded for two accept-then-fails is the wrong direction, so both directions compare against the declared type again — the behaviour before this branch. The conservative half of the trade is the better half: a refusal naming the field beats an incompatible-types error inside a file the author never wrote.

The underlying defect is real and now filed separately, with everything measured. Fixing it needs the rebuild strategy the emitter picks, which is decided after this check runs, so it is an ordering change in a large file rather than a line.

**The accessibility fix stays, with its argument corrected.** A carrier-form pair emits its bridge into the carrier's package rather than the source's, so asking about the source's package was wrong twice over — excluding a member the bridge can call, and including one it cannot.

What remains in this pull request is the raw-`BridgeFn` gap it opened with, which a 625-shape sweep put at 525 emission failures turned into diagnostics with no regressions, plus that accessibility correction.

PR #427, opened 2026-09-20
feat: add observe method

xchopin:
# Summary

Add `observe(Consumer)` to a Telescope path and to generated navigators. Place it after any navigation hop to inspect the value reached there without changing the path's focus or result type. Reads observe current values; updates observe rebuilt values after deeper edits.

# Why

The current DSL has no clear way to attach logging, monitoring, or other effects to intermediate values. Callers have to traverse the graph again or hide those effects inside a transformation function.

A company import can log each department and team as it collects users:

```java
final List<User> importedUsers = Telescope.of(Company.class)
  .each(Company::departments)
  .observe((dept) -> log.info("Importing department {}", dept.name()))
  .each(Department::teams)
  .observe((team) -> log.info("Importing team {}", team.name()))
  .each(Team::users)
  .toList(company);
```

A service can count every user visited by a read:

```java
final var usersRead = meterRegistry.counter("company.users.read");

final long userCount = Telescope.of(Company.class)
    .each(Company::departments)
    .each(Department::teams)
    .each(Team::users)
    .observe(user -> {
      try {
        usersRead.increment();
      } catch (RuntimeException error) {
        log.warn("Could not count user {}", user.id(), error);
      }
    })
    .count(company);
```

For slower publishing, callers can choose their own executor and failure policy inside the callback. For example, a
statement run can submit region summaries and customer statement requests to separate queues while reading the original values:

```java
final long customers = Telescope.of(StatementBatch.class)
  .each(StatementBatch::regions)
  .observe((region) -> {
    final var summary = new RegionSummary(region.code(), region.customers().size());
    callbackExecutor.execute(() -> queuePublisher.publish("statement-operations", summary));
  })
  .each(Region::customers)
  .observe((customer) -> {
    final var request = new StatementRequest(customer.id(), customer.email(), customer.balance());
    callbackExecutor.execute(() -> queuePublisher.publish("statement-generation", request));
  })
  .count(batch);
```

The caller owns executor saturation, callback failures, and delivery guarantees. 

This path submits work while visiting the input; it does not wait for broker delivery.

# Benefits

- One reusable, typed path can observe values at several levels without changing the transformation result.
- `observe` runs only when a terminal operation visits a focus. Short-circuit reads stop invoking observers once they
  stop visiting values.
- Updates report rebuilt values from the inside out. Failed effectful updates do not report rebuilt values.
- Observed paths decline multi-edit fusion so the callback is not skipped by a fused structural pass.
- Focused tests cover read order, short-circuiting, rebuilt values, effectful failures, filters, empty descendants,
  generated navigator forwarding, and the internal traversal behavior.

# How to review

1. Read `Telescope.observe` and `Traversal.observe` for read and write semantics. Check the no-hop-record comment beside `Telescope.observe`: fusion rebuilds from hop records, so observed paths use the sequential fold.
2. Review `TelescopeObserveTest`, `TraversalObserveTest`, and `FocusCodegenTest` for the edge cases and generated forwarder.
3. Run `./gradlew check`. The `internal` and `core` Javadoc tasks are also relevant to the new API docs.


core/src/main/java/io/github/eschizoid/telescope/Telescope.java
```diff
@@ -1317,6 +1319,37 @@ public Telescope<S, A> filter(final Predicate<? super A> predicate) {
...
+    return new Telescope<>(optic.observe(observer), fieldOptics, chain, firstHopName, trail);
+  }
+
+  /**
+   * Submit observations to a caller-owned executor without waiting for callback completion. Reads
+   * submit current values; writes submit rebuilt values after deeper edits. Runtime callback
+   * failures and rejected submissions go to {@code onError}. The caller owns executor shutdown and
+   * callback thread safety. A direct executor may execute callbacks inline.
+   */
+  public Telescope<S, A> observeAsync(
```
eschizoid: Can we drop `observeAsync` and `AsyncObserver`? Your `observe` already covers it:

```java
path.observe(a -> executor.execute(() -> log(a)))
```

`AsyncObserver` bakes in decisions the caller should make: what happens on a rejected task, where callback errors go, the `System.Logger` fallback. I'd rather keep executors out of `:core` entirely. If async observation needs first-class support later we can add an `Applicative` witness on `modifyF` like the `update*` methods, which we can do without breaking anything.

`TelescopeObserveAsyncTest` and the async parts of the README and `docs/effects.md` would go with it. Everything else stays.

core/src/main/java/io/github/eschizoid/telescope/Telescope.java
```diff
@@ -1317,6 +1319,37 @@ public Telescope<S, A> filter(final Predicate<? super A> predicate) {
...
+   * updates observe only successful rebuilt values; {@link #trace(Object)} does not run observers.
+   *
+   * <pre>{@code
+   * Telescope.of(Company.class)
+   *     .each(Company::departments).observe(dept -> log(dept))
+   *     .each(Department::teams).observe(team -> log(team));
+   * }</pre>
+   */
+  public Telescope<S, A> observe(final Consumer<? super A> observer) {
+    // The callback makes multi-edit fusion's changed evaluation order observable.
```
eschizoid: The reason this works is that `trail` without `hops` leaves `hops` null, which makes `Fusion` decline, so `Telescope.all` runs the sequential fold over the composed optic where the observer lives. If someone later threads `hops` through, fused edits skip the observer. Your multi-edit test catches that, but the comment should say the mechanism so whoever breaks it understands the red test:

```suggestion
    // No hop record on purpose: Fusion rebuilds edits from hop records, not the composed optic,
    // so a fused pass would never call the observer. Null hops makes Fusion decline and the
    // sequential fold run the composed optic instead.
```

internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Traversal.java
```diff
@@ -37,15 +38,75 @@
...
+    final var self = this;
+    return new Traversal<>() {
+      @Override
+      public boolean hasObservers() {
+        return true;
+      }
+
+      @Override
+      public Stream<A> getAll(final S source) {
+        return self.getAll(source).peek(observer);
```
eschizoid, line 66: `Stream.peek` can be elided when a terminal doesn't need to traverse (e.g. `count()` on a sized stream). Safe today because `count`/`exists` go through `visitWhile`, which you override. Someone writing a terminal as `getAll(...).count()` would silently lose observations, so:

```suggestion
        // Stream.peek may be elided by terminals that don't traverse, so counting and one-focus
        // terminals must ride visitWhile (overridden below), never a stream terminal over getAll.
        return self.getAll(source).peek(observer);
```

eschizoid (changes requested):
Nice work. The `getAllForUpdate` read/write split, `hasObservers` through `filter`/`then`, and keeping observed paths out of fusion are all right, and the multi-edit test actually fails if the fusion behavior regresses.

One removal (`observeAsync`, inline) and two comment suggestions. The null guards in `read`/`find` change behavior for the Affine/Traversal arms.

xchopin:
Hey @eschizoid thank you for the review 🙏🏻 I'll work on the changes later and update you

internal/src/main/java/io/github/eschizoid/telescope/internal/optics/Traversal.java
xchopin: So, I placed a warning before the return. The `count` and `exists` methods (`core/src/main/java/io/github/eschizoid/telescope/Telescope.java:1880`) use visitation, so they invoke observers (even when a stream’s peek could be elided). 

A future terminal implemented with a `getAll(...).count()` would still be vulnerable, hence the comment records that constraint

eschizoid (approved):

